Issue #3, opened 2018-11-08
Feature Request: Support optimization flags in the build actions


sanathkr:
**Description:**
We can support several optimizations to improve the build process. This is a feature request to support optimization flags that can be passed to each build action. If an action cannot support a particular optimization, it should error out:

### Optimization 1: Use prebuilt binaries 
**Benefit**: Makes build faster

When building dependencies with native modules, use pre-built binaries that are available locally or download from internet. This saves time in recompiling the same binaries over and over again. It can also make the build more stable because sometimes compilation fails or times out.

**Requested by build actions**: 
1. node-npm
1. python-pip

### Optimization 2: Remove unnecessary dependencies 
**Benefit**: Produces slim zips

Remove unnecessary dependency libraries, or files or folders from the output artifact. For example, remove tests, test data files, or dependencies that are already available in transitive closure.  

**Requested by build actions**: 
1. node-npm
1. python-pip

gojko:
A combination of PRs (https://github.com/awslabs/aws-lambda-builders/pull/63 and https://github.com/awslabs/aws-lambda-builders/pull/53) removes the need for this in Node. People can configure packaging sufficiently with this so that further flags aren't really needed. If you decide to implement something globally later for other builders, we could use the same flags to automate these steps further.

PR #5, opened 2018-11-08
Initial draft of Python PIP builder design doc

stealthycoin:
Initial draft of Python PIP builder design doc. Let me know what you think.

### DIFF
--- a/lambda_builders/actions/python_pip/DESIGN.md
+++ b/lambda_builders/actions/python_pip/DESIGN.md
@@ -1,0 +1,149 @@
+## Python - PIP Lambda Builder
+
+### Scope
+
+This package is an effort to port the Chalice packager to a library that can
+be used to handle the dependency resilution portion of packaging Python code
+for use in AWS Lambda. The scope for this builder is to take an existing
+directory containing customer code, and a top-levle `requirements.txt` file
+specifying third party depedencies. The builder will examine the dependencies
+and use pip to build and include the dependencies in the customer code bundle
+in a way that makes them importable.
+
+### Challenges
+
+Python is particularly difficult to package for other platforms because it is
+heavily coupled to your local environment. This is because `setup.py` the
+"config" to describe an install is just a python file, which means authors can
+run code to check things about your current system, and then use that to make
+decisions about how to install the package. These decisions are obviously not
+valid once we move the built package to a different platform.
+
+Python packaging also has the concept of a wheel, which is more of a self
+contained package that can be trivially installed into a python environment
+and does not execute any code. Our goal is to build up a set of these that
+are known to be compatible with AWS Lambda.
+
+### Interface
+
+The top level interface will consist of a single function with the signature:
+
+```python
+def build_depenencies(source_path, config=None):
+    """Builds a source directory's dependencies into the directory.
+
+	:type source_path: str
+	:param source_path: The source directory that needs to have dependencies
+		built for it. It is assumed that this customer code is correct and
+		would function if it had all of its dependencies installed alongside
+		it. It is assumed that the dependencies are specified in a top level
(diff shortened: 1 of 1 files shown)


lambda_builders/actions/python_pip/DESIGN.md
```diff
@@ -0,0 +1,149 @@
...
+```python
+def build_depenencies(source_path, config=None):
+    """Builds a source directory's dependencies into the directory.
+
+	:type source_path: str
+	:param source_path: The source directory that needs to have dependencies
+		built for it. It is assumed that this customer code is correct and
+		would function if it had all of its dependencies installed alongside
+		it. It is assumed that the dependencies are specified in a top level
+		``requirements.txt`` file. This file can be removed before deploying
```
sanathkr: I would assume that this method is passed a path to `requirements.txt`. SAM CLI will ensure the file is at top-level within the source folder but this should not be an assumption that low level build actions must make.

lambda_builders/actions/python_pip/DESIGN.md
```diff
@@ -0,0 +1,149 @@
...
+		to lambda, but the this function will not modify any existing files
+		in this directory.
+
+	:type config: :class:`lambda_builders.actions.python_pip.config`
+	:param config: A configuration object which contains additional config
```
sanathkr: I would call this out as optional, and not implemented, for now. This requirement exists for several build actions. So I want to standardize this as Optimizations that can turned on/off by customers - https://github.com/awslabs/aws-lambda-builders/issues/3

lambda_builders/actions/python_pip/DESIGN.md
```diff
@@ -0,0 +1,149 @@
...
+There are a couple ways error reporting can be done.
+
+1) We can raise a python error and have the caller catch and deal with
+   it at the end. The error may not be significant, for example if the user
+   is already aware of the problem and has a post-build hook that fixes the
+   bundle with a custom dependency injection. So this exception would be thrown
+   at the end and not interrupt the rest of the dependency building.
+2) Return None or a list of aggregated error objects that can be inspected by
+   the caller.
+3) Use the UI object. The builder as it is written right now has a notion of a
```
sanathkr: I like the idea of a UI object. This allows us to abstract functionality like "write with colors" etc within the object. We could also redirect the data to a file transparently. Feel free to make assumptions about this object for now. We can implement a real UI object separately.

lambda_builders/actions/python_pip/DESIGN.md
stealthycoin: Ah yes I meant to write that, my bad.

lambda_builders/actions/python_pip/DESIGN.md
stealthycoin: Cool. For now Whatever wrapper around this low level builder can create and inject it.

lambda_builders/actions/python_pip/DESIGN.md
stealthycoin: Ok I can make this more in line with the params outlined in your other PR. I will still assume I have an adapter between them to make this low level builder more extractable and specific to python.

sanathkr: Makes sense

lambda_builders/actions/python_pip/DESIGN.md
sanathkr: Ah I missed it ;)

sanathkr:
Let me know once you have pushed a new draft. I will approve and merge

sanathkr:
Accidentally deleted this file from the repo after creating the new directory structure. I will add this to my other open PR

PR #6, opened 2018-11-08
Builder interface with ability to dynamically load actions & run

sanathkr:
`lambda_builders/builder.py` contains the `Builder` class that callers will
instantiate and use. This class dynamically loads the entrypoint for the given
build action and runs it.

Lots of things are remaining, but sending out a WIP PR to give you an idea of how this is going to work:

- [x] Complete implementation
- [x] Unit tests & linter fixes
- [x] Functional tests for loading and running a custom workflow 
- [x] Functional test for `utils.copytree`

### DIFF
--- a/aws_lambda_builders/registry.py
+++ b/aws_lambda_builders/registry.py
@@ -1,36 +1,99 @@
 """
-Registry of auto-discovered workflows
+Registry of auto-discovered workflows. Classes and methods in this module are thread-safe.
 """
 
-# TODO: Is this datastructure threadsafe?
-# Maintains a registry of builders
-_REGISTRY = {}
+import threading
+
+from aws_lambda_builders.exceptions import WorkflowNotFoundError
 
 
-def get_builder(language, language_framework, application_framework):
+class Registry(object):
     """
-    Find and return a builder class capable of building for the given combination of capabilities
+    Stores a registry of workflows. Only one workflow matching a given capability can be stored. If attempting to
+    register two workflows for same capability, it will raise an exception.
 
-    :param language:
-    :param language_framework:
-    :param application_framework:
+    This class is thread-safe for multiple writers (ie. registers happening simultaneously from different threads).
+    """
 
+    def __init__(self, write_lock=None):
+        self._write_lock = write_lock or threading.Lock()
+        self._data = {}
+
+    def __getitem__(self, capability):
+        key = self._make_key(capability)
+        return self._data[key]
+
+    def __setitem__(self, capability, value):
+
+        key = self._make_key(capability)
+
+        try:
+            self._write_lock.acquire()
+
+            if key in self._data:
+                raise KeyError("A workflow with given capabilities '{}' is already registered".format(key))
+
+            self._data[key] = value
+
+        finally:
+            self._write_lock.release()
+
+    def __contains__(self, capability):
+        key = self._make_key(capability)
+        return key in self._data
+
+    def __len__(self):
+        return len(self._data)
+
+    def clear(self):
+        try:
+            self._write_lock.acquire()
+            self._data.clear()
+        finally:
+            self._write_lock.release()
+
+    @staticmethod
+    def _make_key(capability):
+        """
+        Given the capabilities, generate a string that can be used as the key for the registry object
+        """
+
+        # Key is created by concatenating the capabilites data with underscore.
+        # This delimiter is positional ie. if a value is not provided, the delimiter still needs to exist in the key.
+        # This helps us be forwards compatible with new capabilities
+        return "_".join([
+            capability.language or "",
+            capability.dependency_manager or "",
+            capability.application_framework or ""
+        ]).lower()
+
+
+# Built-in registry of workflows.
+DEFAULT_REGISTRY = Registry()
+
+
+def get_workflow(capability, registry=DEFAULT_REGISTRY):
+    """
+    Find and return a workflow class capable of acting on the given combination of capabilities
+
+    :type capability: aws_lambda_builders.workflow.capabilities
+    :param capability:
+        The capabilities you want from this workflow
+
+    :type registry: aws_lambda_builders.registry.Registry
+    :param registry:
+        Registry to fetch the workflow from
+
+    :rtype: aws_lambda_builders.workflow.BaseWorkflow
     :return:
-    :raises BuilderNotFoundError: If a builder was not registered
-    """
-    pass
+        A workflow class that has the given capabilities. Note: This returns a class and not an instance of the class.
 
-
-class WorkflowMetaClass(type):
-    """
-    A metaclass that maintains the registry of loaded builders
+    :raises WorkflowNotFoundError: If a workflow with given capabilities was not found
     """
 
-    def __new__(mcs, name, bases, class_dict):
-        """
-        Add the builder to registry when loading the class
-        """
+    if capability not in registry:
+        raise WorkflowNotFoundError(language=capability.language,
+                                    dependency_manager=capability.dependency_manager,
+                                    application_framework=capability.application_framework)
 
-        cls = type.__new__(mcs, name, bases, class_dict)
-
-        return cls
+    return registry[capability]
--- a/aws_lambda_builders/utils.py
+++ b/aws_lambda_builders/utils.py
@@ -1,0 +1,58 @@
+"""
+Common utilities for the library
+"""
+
+import shutil
+import os
+import logging
+
+
+LOG = logging.getLogger(__name__)
+
+
+def copytree(source, destination, ignore=None):
+    """
+    Similar to shutil.copytree except that it removes the limitation that the destination directory should
+    be present.
+
+    :type source: str
+    :param source:
+        Path to the source folder to copy
+
+    :type destination: str
+    :param destination:
+        Path to destination folder
+
+    :type ignore: function
+    :param ignore:
+        A function that returns a set of file names to ignore, given a list of available file names. Similar to the
+        ``ignore`` property of ``shutils.copytree`` method
+    """
+
+    if not os.path.exists(destination):
+        os.makedirs(destination)
+
+        try:
+            # Let's try to copy the directory metadata from source to destination
+            shutil.copystat(source, destination)
+        except WindowsError as ex:  # pylint: disable=undefined-variable
+            # Can't copy file access times in Windows
+            LOG.debug("Unable to copy file access times from %s to %s", source, destination, exc_info=ex)
+
+    names = os.listdir(source)
+    if ignore is not None:
+        ignored_names = ignore(source, names)
+    else:
+        ignored_names = set()
+
+    for name in names:
+        # Skip ignored names
+        if name in ignored_names:
+            continue
+
+        new_source = os.path.join(source, name)
+        new_destination = os.path.join(destination, name)
+        if os.path.isdir(new_source):
+            copytree(new_source, new_destination, ignore=ignore)
+        else:
+            shutil.copy2(new_source, new_destination)
--- a/aws_lambda_builders/workflow.py
+++ b/aws_lambda_builders/workflow.py
@@ -4,28 +4,78 @@ Implementation of a base workflow
 
 import os
 import logging
+
+from collections import namedtuple
 import six
 
-from aws_lambda_builders.registry import WorkflowMetaClass
+from aws_lambda_builders.registry import DEFAULT_REGISTRY
+from aws_lambda_builders.exceptions import WorkflowFailedError, WorkflowUnknownError
+from aws_lambda_builders.actions import ActionFailedError
 
 LOG = logging.getLogger(__name__)
 
 
-class BaseWorkflow(six.with_metaclass(WorkflowMetaClass, object)):
+# Named tuple to express the capabilities supported by the builder.
+# ``Language`` is the programming language. Ex: Python
+# ``LangageFramework`` is the framework of particular language. Ex: PIP
+# ``ApplicationFramework`` is the specific application framework used to write the code. Ex: Chalice
+Capability = namedtuple('Capability', ["language", "dependency_manager", "application_framework"])
+
+
+class _WorkflowMetaClass(type):
+    """
+    A metaclass that maintains the registry of loaded builders
+    """
+
+    def __new__(mcs, name, bases, class_dict):
+        """
+        Add the builder to registry when loading the class
+        """
+
+        cls = type.__new__(mcs, name, bases, class_dict)
+
+        # We don't want to register the base classes, so we simply return here.
+        # Also, skip further steps if the class is marked for testing
+        if cls.__name__ == 'BaseWorkflow' or cls.__TESTING__:
+            return cls
+
+        # Validate class variables
+
+        # All classes must provide a name
+        if not isinstance(cls.NAME, six.string_types):
+            raise ValueError("Workflow must provide a valid name")
+
+        # All workflows must express their capabilities
+        if not isinstance(cls.CAPABILITY, Capability):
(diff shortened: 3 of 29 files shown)


lambda_builders/builder.py
```diff
@@ -0,0 +1,129 @@
...
+            Optional dictionary of optimization flags to pass to the build action. **Not supported**.
+
+        :type options: dict
+        :param options:
+            Optional dictionary of options ot pass to build action. **Not supported**.
+
+        :raises lambda_builders.exceptions.UnsupportedBuiltActionError
+        """
+
+        action_klass = self._load_action(action_name)
```
stealthycoin: How about `action_cls` to be more in line with python convention?

lambda_builders/builder.py
```diff
@@ -0,0 +1,129 @@
...
+                              runtime=runtime,
+                              optimizations=optimizations,
+                              options=options)
+
+        # TODO: Verify source & artifacts directory & manifests exists
+
+        if not action.is_supported():
+            raise UnsupportedManifestError(action_name, manifest_path)
+
+        action.resolve()
```
stealthycoin: For python at least, part of resolving the dependencies is compiling them. Since I don't know if a particular dependency is acceptable until after I try 'compiling' it. I think this section is a noop for python.

lambda_builders/builder.py
```diff
@@ -0,0 +1,129 @@
...
+        action.compile()
+        action.copy_source()
```
stealthycoin: Is this maybe delegating too many tasks to one class? The way I am visualizing this, I think each action here in the resolve/compile/copy_source list should be a separate class that is delegated too by the builder. 

The class hierarchy would look like:

* BaseBuilder
** PythonBuilder

* BaseAction
** PythonDependencyBuildAction
** CopySourceAction

More subclasses are added per language, and each specific builder is essentially this:
```python
class BaseBuilder(object):
    def __init__(self, *args):
        self.actions = []

    def build(self):
        for action in self.actions:
            action.execute()


class PythonBuilder(BaseBuilder):
    def __init__(self):
        self.actions = [
            PythonDependencyBuildAction(<subset of *args that makes sense here>),
            CopySourceAction(<subset of *args that make sense here>),
        ]


class PythonDependencyBuildAction(BaseAction): 
    def __init__(self, artifact_dir_path, requirements_path, runtime, dependency_builder=None):
        self._artifact_dir_path = artifact_dir_path
        self._requirements_path = requirements_path
        self._runtime = runtime
        if dependency_builder is None:
            dependency_builder = PythonPipDependencyBuilder()
        self._dependency_builder = dependency_builder

    def execute(self):
        self._dependency_builder.build_dependencies(  # <-- this would be whatever the interface is i'm proposing in my PR.
            self._artifact_dir_path, 
            self._requirements_path, 
            self._runtime,
        ) 

```    
I like breaking this all out and making it clearly separated by language/tool, all separately testable and composable. New languages would be really easy to add by just adding the the `Actions` they need, and making a subclass of `Builder` that arranges them in the correct order. Makes python's "Special" case of no compile readily apparent too.

I think without this we are coupling the idea of how the build works to our builder, which if we are supporting x languages and y package managers I don't think is going to be very sustainable/flexible/testable. Downside is we are subprocessing for each action that can't be performed in pure python. 

What do you think? I could sketch this out in more concrete code if it's not clear.

lambda_builders/builder.py
```diff
@@ -0,0 +1,129 @@
...
+        """
+
+        entrypoint_path = os.path.join(self.build_actions_dir, action_name, self.entrypoint_file_name)
+
+        if not os.path.exists(entrypoint_path):
+            raise UnsupportedBuildActionError(action_name, search_path=self.build_actions_dir)
+
+        entrypoint_module = importlib.import_module(entrypoint_path)
+
+        for _, klass in inspect.getmembers(entrypoint_module,
```
stealthycoin: What versions of python are we supporting? If we go with the class based approach I sketched out ahead we could make the specific `Builder` subclasses know which (language, packager) combo then can build for and register themselves against a master list which will make them discoverable and easy to use from the top level API.

```python
builder_cls = get_builder_for('python', 'pip')
builder = builder_cls(*args)
builder.build()
```

`get_builders` would just look it up in the master list that the classes register themselves against so there wouldn't be any need for all the custom logic to find them with inspect etc. This would kind of remove the need for the `entrypoint.py` files, since you could just get access to the builders directly at the top level and they all have the same interface so you can just call into their `__init__` and `build()`.

lambda_builders/builder.py
sanathkr: I love this! Makes it way more composable and testable. For out of process builder, we can spawn process once and pass JSONRPC commands for each action. 

I really like how actions can be skipped or reordered. 

I will modify the implementation to use this pattern.

lambda_builders/builder.py
sanathkr: We have to support Py2.7 and >= Py3.4. 

I like this approach. It is more plug-n-play. The reason I chose import & inspect is to lazily load the build action modules. I think we can still achieve lazy loading with your idea. I'll prototype that.

Also, I do want each builder to implement the entrypoint.py or something that explicitly makes an arbitrary folder as a build action. In the future, this can provide a mechanism to support pluggable/custom actions that customers provide. 

Not sure if I am over thinking this. I will prototype and check how this all works..

sanathkr:
@stealthycoin Just pushed a WIP implementation of automatically registering builders using metaclasses. I have not run this code yet, but I want to understand if this approach generally looked good.

lambda_builders/lib/base.py
```diff
@@ -0,0 +1,145 @@
+"""
```
stealthycoin: I'm not sold on the directory structure we have so far. I like my directory structure to be conducive to new contributors finding things easily and making it clear which component is where, and making it easy to extend current functionality. We have to take special care here as well since we have a fold in our code where we need to know about language-specific things, and on the other side we do not. At the high level we have two main abstractions.`Builder` and `Action`. Builder's do not need to have any specific knowledge about the language they are building for, they just encode what they are building in their classname and their capabilities object, as well as containing a list of actions. 

The actions themselves do not need to know anything specific to their language as some will be general purpose like `CopySourceAction`. Others like `PythonPipResolveAction` obviously are specific to python, but don't acutaly have to have any logic in them specific to python, the actions execute methods are essentially just going to proxy out out to the lower level classes that implement language-specific procedures. I think the ladder classes are the only ones that really need to be "below the fold" of language-specific. So what do you think about the project layouts below:

```
aws_lambda_builders
├── __init__.py
├── __main__.py <- CLI hooks
├── action.py   <- `BaseAction`, as well as all of  its subclasses are in here.
├── build.py    <- The main builder is started from here. This is what selects a language-specific builder and runs it.
├── builders.py <- Language specific builders live in here. They can all be registered by doing from builders import *
├── exceptions.py
├── language_support
│   ├── __init__.py
│   ├── dotnet
│   │   └── __init__.py
│   ├── javascript
│   │   └── __init__.py
│   └── python
│       └── __init__.py <- Here is where all the python specific things happen. Even though the `PythonPipResolveAction` lives in the `actions.py` above, it imports pieces from here and runs them. 
└── registry.py 
4 directories, 10 files
```

Downside of this approach is that all the related actions/builders/low level implementation details are not sorted by language, they are instead sorted by their place in the class hierarchy. So adding a new builder would mean doing three things:
* Add a `CoolNewThingBuilder` to the `builders.py` file.
* Add any new `Action` subclasses needed by the new builder to the `actions.py` file.
* Any low level logic needed by the new `Action`s to perform the build of a `CoolNewThing` would go in `language_support/cool_new_thing/__init__.py` (or sibling files) and would be imported by the new actions.


Or what about this
```
aws_lambda_builders
├── __init__.py
├── __main__.py
├── action.py   <- This now just has the BaseAction class and any common language-agnostic Actions like CopySourceAction
├── build.py    <- This is more or less the same as above.
├── builders  <- Now instead of having all the builders/actions in one place, they are sorted by their language. You now need to use your technique to iterate these submodules and import the builders so that they get included in your registry.
│   ├── __init__.py
│   ├── dotnet
│   │   ├── __init__.py
│   │   ├── actions.py
│   │   └── builders.py
│   ├── javascript
│   │   ├── __init__.py
│   │   ├── npm_packager.js <- custom stuff would live alongside the python in these modules or in more subfolders.
│   │   ├── actions.py
│   │   └── builders.py
│   └── python
│       ├── __init__.py
│       ├── packager.py <- low level packager code that is called by the PythonPipResolveAction to do its job.
│       ├── actions.py  <- This now contains python specific actions like PythonPipResolveAction
│       └── builders.py <- The python specific builders would be defined here
├── exceptions.py
└── registry.py

4 directories, 16 files
```

Benefit here is that our high level build/action system is pulled out, and each language-specific piece acts almost like a plugin, its its own self contained directory. Someone could develop their own "package" with the structure
```
ruby
├── __init__.py
├── packager.rb
├── utils.rb
├── actions.py
└── builders.py
```
And essentially drop into the builders package (or maybe we can have a notion of a `BUILDERS_PATH` that is searched for these things and the default entry is this vended builders dir.) to get it to work. This seems the friendliest to me.

Thoughts?

lambda_builders/actions/python_pip/builder.py
```diff
@@ -0,0 +1,28 @@
+
+from lambda_builders.lib.base import BaseBuilder, BuilderCapability
+from lambda_builders.lib.action_copy_source import CopySourceAction
+from .actions import PythonPipResolveAction
+
+
+class PythonPipBuilder(BaseBuilder):
+
+    CAPABILITY = BuilderCapability(Language="python", LanguageFramework="PIP", ApplicationFramework="")
```
stealthycoin: Can we snake case the argument names?

lambda_builders/lib/base.py
```diff
@@ -0,0 +1,145 @@
...
+    """
+    # TODO: Do we need a `error_code` of some sorts here? If so, we should create a named list of error codes.
+    pass
+
+
+# Named tuple to express the capabilities supported by the builder.
+# ``Language`` is the programming language. Ex: Python
+# ``LangageFramework`` is the framework of particular language. Ex: PIP
+# ``ApplicationFramework`` is the specific application framework used to write the code. Ex: Chalice
+BuilderCapability = namedtuple('BuilderCapability', ["Language", "LanguageFramework", "ApplicationFramework"])
```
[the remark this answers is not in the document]
sanathkr: An application could be written in Python + PIP + an app framework like Chalice/Zappa/Serverless Framework. So this becomes a whole new builder that knows how to handle Chalice+Python+PIP projects. It will simply shell out to `chalice package` and optionally do a few more things. My goal was to support building not just vanilla Python/JS/Java projects, but also projects that use specialized app frameworks that need a custom action

lambda_builders/lib/base.py
```diff
@@ -0,0 +1,145 @@
...
+
+    def execute(self):
+        """
+        The builder will run this method on each action. This method should complete the action, and if it fails
+        raise appropriate exceptions.
+
+        :raises lambda_builders.exceptions.ActionError: Instance of this class if something went wrong with the
+            action
+        """
+        raise NotImplementedError()
```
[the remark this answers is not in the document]
sanathkr: Will do

lambda_builders/actions/python_pip/builder.py
```diff
@@ -0,0 +1,28 @@
...
+    SUPPORTED_MANIFESTS = ["requirements.txt"]
+
+    _IGNORE_FILES = ('*.pyc', '.git', '__pycache__', ".DS_Store", "*.egg-info", "*.egg", "pip-log.txt")
```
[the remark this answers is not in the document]
sanathkr: I agree :)

lambda_builders/actions/python_pip/builder.py
```diff
@@ -0,0 +1,28 @@
...
+                 artifacts_dir,
+                 manifest_path,
+                 runtime=None,
+                 optimizations=None,
+                 options=None):
+
+        super(PythonPipBuilder, self).__init__(
+            source_dir, artifacts_dir, manifest_path, runtime, optimizations, options)
+
+        self.actions = [
```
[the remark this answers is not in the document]
sanathkr: Yes! when I wrote this builder it was super easy :)

lambda_builders/actions/python_pip/builder.py
sanathkr: Sure thing.

sanathkr:
I force pushed to the branch. SO closing & re-opening the PR to trigger Travis again

lambda_builders/exceptions.py
```diff
@@ -0,0 +1,41 @@
...
+    """
+    Raised when a builder for the request capabilities does not exist
+    """
+
+    def __init__(self, capabilities):
+        msg = "A builder for the given capabilities '{}' was not found".format(capabilities)
+        Exception.__init__(self, msg)
+
+
+class UnsupportedManifestError(Exception):
```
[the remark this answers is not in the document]
sanathkr: I implemented this

lambda_builders/lib/base.py
sanathkr: This was implemented in #8

aws_lambda_builders/registry.py
```diff
@@ -1,36 +1,99 @@
...
 
-    :return:
-    :raises BuilderNotFoundError: If a builder was not registered
+    This class is thread-safe for multiple writers (ie. registers happening simultaneously from different threads).
     """
-    pass
 
+    def __init__(self, write_lock=None):
+        self._write_lock = write_lock or threading.Lock()
+        self.data = {}
```
garnaat: A nit.  I would make this _data since I think the assumption is this is private to the class.

aws_lambda_builders/utils.py
```diff
@@ -0,0 +1,54 @@
...
+
+    if not os.path.exists(destination):
+        os.makedirs(destination)
+
+        try:
+            # Let's try to copy the directory metadata from source to destination
+            shutil.copystat(source, destination)
+        except WindowsError:  # pylint: disable=undefined-variable
+            # Can't copy file access times in Windows
+            pass
```
garnaat: Is this safe?  It's not clear to me what we are smothering with this pass statement.

aws_lambda_builders/workflow.py
```diff
@@ -4,28 +4,78 @@
...
+from aws_lambda_builders.exceptions import WorkflowFailedError, WorkflowUnknownError
+from aws_lambda_builders.actions import ActionFailedError
 
 LOG = logging.getLogger(__name__)
 
 
-class BaseWorkflow(six.with_metaclass(WorkflowMetaClass, object)):
+# Named tuple to express the capabilities supported by the builder.
+# ``Language`` is the programming language. Ex: Python
+# ``LangageFramework`` is the framework of particular language. Ex: PIP
```
garnaat: Would PackageFramework be better?  LanguageFramework confused me initially into thinking it meant ApplicationFramework.  Maybe just me.

garnaat: Or maybe just Packager?

sanathkr: Good point. I'll call it `DependencyFramework` then..

aws_lambda_builders/utils.py
sanathkr: Not sure. I'll just add a debug log here instead

aws_lambda_builders/registry.py
sanathkr: Sounds good.

aws_lambda_builders/workflow.py
sanathkr: I went thru a websites for whole bunch of dependency frameworks like PIP, Yarn, NPM, Bundle etc. `DependencyManager` is the closest consistent definition

Issue #11, opened 2018-11-13
Check for supported version of runtime in shell before executing workflow


sanathkr:
**Description:**
Every workflow has to make sure it has a compatible version of Python/Node/NPM/PIP etc in the shell before processing. This is necessary to make sure customers are building in a version of programming language that is compatible with AWS Lambda runtime (ex: py3.6 vs py2.7 are incompatible). 

This is grunt work that could be abstracted into the `BaseWorkflow` similar to `BaseWorkflow.SUPPORTED_MANIFESTS`. This will also help us generate a standard error message on failure to satisfy prerequisites across all workflows.

TheSriram:
This is addressed with Resolver and Validator features.

PR #22, opened 2018-11-15
fix: validate runtime

TheSriram:
- Based on the language supplied and lambda runtimes, validate the
  host runtime to make sure they are compatible.

Addresses: https://github.com/awslabs/aws-lambda-builders/issues/11


aws_lambda_builders/runtime.py
```diff
@@ -0,0 +1,49 @@
+"""
+Supported Runtimes and their validations.
+"""
+
+import os
+import subprocess
+from enum import Enum
+
+from aws_lambda_builders.exceptions import MisMatchRuntimeError
+from aws_lambda_builders.workflows.python_pip.runtime_validator import validate_python_cmd
```
sanathkr: no, you can't do this :( Think of anything inside the `workflow` folder as a whole separate library that you don't have access to. You can only use the interfaces but not individual implementation. 

This allows us to evolve the workflows indendepent of the wrapper library

TheSriram: Sounds good, I could move it. My motivation behind this was to have language specific validation logic, live inside that specific language's module.

tests/functional/test_cli.py
```diff
@@ -53,11 +53,13 @@ def test_run_hello_workflow(self):
                 "artifacts_dir": self.artifacts_dir,
                 "scratch_dir": "/ignored",
                 "manifest_path": "/ignored",
+                "runtime": None,
```
TheSriram: I had to explicitly set this to None, matching it to my python runtime didnt work, and I'm not sure why. Integ tests still work.

sanathkr: We should set `runtime` to something like `ignored` so the validator fails open.

TheSriram: 👍

aws_lambda_builders/runtime.py
```diff
@@ -0,0 +1,66 @@
...
+    @classmethod
+    def validate_runtime(cls, required_language, required_runtime):
+        """
+        Checks if the language supplied matches the required lambda runtime
+        :param string required_language: language to check eg: python
+        :param string required_runtime: runtime to check eg: python3.6
+        :raises ValueError: Unsupported Lambda Runtime
+        :raises MisMatchRuntimeError: Version mismatch of the lanugage vs the required runtime
+        """
+        try:
```
sanathkr: you could just do a `if key in dict` instead of try/catch. That's more pythonic

aws_lambda_builders/runtime.py
```diff
@@ -0,0 +1,66 @@
...
+            cmd = _RUNTIME_VERSION_RESOLVER[required_language](required_language, required_runtime)
+        except KeyError:
+            # The language is not currently supported, so there is no runtime validation.
+            return
+        if not Runtime.has_value(required_runtime):
```
sanathkr: I would just fail open here. I know this is useful to have, but our runtime checks are never going to be perfect given the long list of languages & dependency managers we have. So I'd rather just fail and log a warning instead.

tests/integration/workflows/python_pip/test_python_pip.py
```diff
@@ -28,31 +29,40 @@ def setUp(self):
         self.builder = LambdaBuilder(language="python",
                                      dependency_manager="pip",
                                      application_framework=None)
+        self.runtime = "{language}{major}.{minor}".format(
```
sanathkr: nice!! this makes the test work across multiple runtimes :)

aws_lambda_builders/runtime.py
```diff
@@ -0,0 +1,66 @@
...
+            minor=minor)]
+    return cmd
+
+
+_RUNTIME_VERSION_RESOLVER = {
+    "python": validate_python_cmd
+}
+
+
+class Runtime(Enum):
```
sanathkr: wait, this needs to work in Py2.7. Enum isn't available there and I am not interested in taking a new dependency on Enum because it doesn't offer a whole lot of functionality in this case ;) 

Can you refactor this to remove Enum?

aws_lambda_builders/runtime.py
```diff
@@ -0,0 +1,66 @@
+"""
```
sanathkr: Can we also call this `validate.py` instead of runtime? It is more representative of what this file does.

tests/integration/workflows/python_pip/test_python_pip.py
TheSriram: Yes! :)

aws_lambda_builders/runtime.py
TheSriram: Sure removed the enum, its a simple class now.

aws_lambda_builders/runtime.py
TheSriram: Done.

aws_lambda_builders/runtime.py
TheSriram: the reason I didnt want that, was the all if elsing inside that function. maybe a decorator would help? I removed the try catch now.

aws_lambda_builders/runtime.py
TheSriram: Done. warning logged.

PR #32, opened 2018-11-22
Python 37 support

sanathkr:
*Description of changes:*
Lambda supports Python 3.7. Adding support for building Py37 functions in the builder.

Tested this change locally by running integration tests in Python 3.7

### DIFF
--- a/aws_lambda_builders/workflows/python_pip/packager.py
+++ b/aws_lambda_builders/workflows/python_pip/packager.py
@@ -55,17 +55,35 @@ class PackageDownloadError(PackagerError):
     pass
 
 
+class UnsupportedPythonVersion(PackagerError):
+    """Generic networking error during a package download."""
+    def __init__(self, version):
+        super(UnsupportedPythonVersion, self).__init__(
+            "'%s' version of python is not supported" % version
+        )
+
+
 def get_lambda_abi(runtime):
-    return {
+    supported = {
         "python2.7": "cp27mu",
         "python3.6": "cp36m",
         "python3.7": "cp37m"
-    }.get(runtime)
+    }
+
+    if runtime not in supported:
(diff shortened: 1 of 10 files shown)


aws_lambda_builders/workflows/python_pip/packager.py
```diff
@@ -138,7 +145,7 @@ class DependencyBuilder(object):
         'sqlalchemy'
     }
 
-    def __init__(self, osutils, pip_runner=None):
+    def __init__(self, osutils, pip_runner=None, runtime="python2.7"):
```
stealthycoin: Defaulting to 2.7 😲 ?

aws_lambda_builders/workflows/python_pip/packager.py
```diff
@@ -56,6 +55,14 @@ class PackageDownloadError(PackagerError):
     pass
 
 
+def get_lambda_abi(runtime):
+    return {
+        "python2.7": "cp27mu",
+        "python3.6": "cp36m",
+        "python3.7": "cp37m"
+    }.get(runtime)
```
stealthycoin: Is there any way we can ever get here with an unsupported `runtime`? If so do we want to raise something here like `UnsupportedRuntimeError` with a property for the `runtime` value? Or is this impossible, because of the validator? If its impossible I think I'd rather `[runtime]` so we at least get a very obvious `KeyError` instead of a `None` silently sneaking in.

stealthycoin: Ok yea. I think I like having an explicit error for it. Thoughts?

aws_lambda_builders/workflows/python_pip/packager.py
sanathkr: haha removed the default.

PR #44, opened 2018-12-01
[WIP] support for building nodejs_npm functions

gojko:
*Description of changes:* Initial support for building nodejs_npm, for early code review. This does not yet support building projects with local dependencies.


tests/functional/workflows/python_pip/test_utils.py
```diff
@@ -20,3 +24,41 @@ def test_can_read_unicode(self, tmpdir, osutils):
         content = osutils.get_file_contents(filename, binary=False,
                                             encoding='utf-16')
         assert content == checkmark
+
+    def test_find_executable_returns_path_for_known_exe(self, tmpdir, osutils):
+        found = osutils.find_executable('python')
+
+        assert os.path.realpath(found) == os.path.realpath(sys.executable)
+
+    def test_find_executable_returns_none_for_non_existing_exe(self, tmpdir, osutils):
```
sanathkr: awesome1 thanks for wrigint functional tests

tests/integration/workflows/nodejs_npm/test_nodejs_npm.py
```diff
@@ -0,0 +1,79 @@
...
+import os
+import shutil
+import tempfile
+from unittest import TestCase
+
+from aws_lambda_builders.builder import LambdaBuilder
+from aws_lambda_builders.exceptions import WorkflowFailedError
+
+
+class TestNodejsNpmWorkflow(TestCase):
```
sanathkr: Love the integ tests and handling of a variety of cases!!

tests/integration/workflows/nodejs_npm/test_nodejs_npm.py
```diff
@@ -0,0 +1,79 @@
...
+
+        self.builder.build(source_dir, self.artifacts_dir, self.scratch_dir,
+                           os.path.join(source_dir, "package.json"),
+                           runtime=self.runtime)
+
+        expected_files = {"package.json", "included.js", "node_modules"}
+        output_files = set(os.listdir(self.artifacts_dir))
+        self.assertEquals(expected_files, output_files)
+
+        expected_modules = {"minimal-request-promise"}
```
sanathkr: sweet!

tests/integration/workflows/nodejs_npm/test_nodejs_npm.py
```diff
@@ -0,0 +1,79 @@
...
+    def test_fails_if_package_json_is_broken(self):
+
+        source_dir = os.path.join(self.TEST_DATA_FOLDER, "broken-package")
+
+        with self.assertRaises(WorkflowFailedError) as ctx:
+            self.builder.build(source_dir, self.artifacts_dir, self.scratch_dir,
+                               os.path.join(source_dir, "package.json"),
+                               runtime=self.runtime)
+
+        self.assertIn("Unexpected end of JSON input", str(ctx.exception))
```
sanathkr: what does this really mean? Is this a common case? Should we catch and create a custom exception with more descriptive message?

tests/integration/workflows/nodejs_npm/testdata/npm-deps/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "npmdeps",
+  "version": "1.0.0",
+  "description": "",
+  "files": ["included.js"],
```
sanathkr: Is this `files` section an advanced usecase? I felt like ppl usually just zipped up whatever is in their function's folder. 

If this is advanced, I want to make sure we cater to the basic users as well.

sanathkr:
Copying conversation from slack on difference between `claudia pack` and this implementation for sake of tracking:

#### `claudia pack` vs `node_npm` workflow from @gojko 
There are features in claudia pack that are not in the code yet, but 90% of the common cases will be covered by what I submitted in the PR

- There is a lot of complexity to handle local file dependencies, which I find very useful for my work as I like to keep multiple projects in the same repository and just refer locally to them as dependencies. Once I know that the overall code design is OK, I’ll port that over.

- Another thing very useful in `claudia pack` that’s not here is the ability to exclude optional dependencies (we discussed that earlier, I will add an flag that depends on an environment variable to activate/deactivate that).

- The third thing I also think is worth porting over from `claudia pack` is an option to run a post-package script for any kind of project specific build. NPM has the ability to put build phase scripts (similar to Maven) into the manifest, not just dependencies... so this would be a way for developers to customise the stuff sent up to lambda in an automated way

aws_lambda_builders/workflows/nodejs_npm/actions.py
```diff
@@ -0,0 +1,92 @@
+"""
+Action to resolve NodeJS dependencies using NPM
+"""
+
+from aws_lambda_builders.actions import BaseAction, Purpose, ActionFailedError
+from aws_lambda_builders.workflows.python_pip.utils import OSUtils
```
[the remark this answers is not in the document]
gojko: should I just copy the whole class into nodejs_npm and remove unnecessary functions, or do you want to promote osutils to something more generic that can be inherited/used by both flows? I assume that even when you move individual workflows into separate repositories, some generic stuff (the base actions/workflows) will be in some generic utility package? In that case, osutils could go there.

aws_lambda_builders/workflows/nodejs_npm/actions.py
```diff
@@ -0,0 +1,92 @@
...
+    def execute(self):
+        try:
+            if not self.osutils.file_exists(self.manifest_path):
+                raise ActionFailedError('package.json not found in: %s' % self.manifest_path)
+
+            package_path = "file:%s" % self.osutils.abspath(self.osutils.dirname(self.manifest_path))
+
+            LOG.debug("NODEJS packaging %s to %s", package_path, self.scratch_dir)
+
+            tarfile_name = self.subprocess_npm.main(['pack', '-q', package_path], cwd=self.scratch_dir)
```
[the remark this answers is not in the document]
gojko: yes, as far as I know. `npm pack` produces a tarball (https://docs.npmjs.com/cli/pack)

aws_lambda_builders/workflows/nodejs_npm/actions.py
```diff
@@ -0,0 +1,92 @@
...
+
+            LOG.debug("NODEJS packed to %s", tarfile_name)
+
+            tarfile_path = self.osutils.joinpath(self.scratch_dir, tarfile_name)
+
+            self.osutils.extract_tarfile(tarfile_path, tarfile_path + '-unpacked')
+
+            self.osutils.copytree(self.osutils.joinpath(tarfile_path + '-unpacked', 'package'), self.artifacts_dir)
```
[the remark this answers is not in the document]
gojko: I'll leave this decision to you but copysource also has additional features such as file filtering which we might not want here (I assume we trust NPM to know what files to include/exclude). This again goes back to the idea to share osutils which I would prefer. We need to have some way of untaring anyway, which requires some form of osutils here.

aws_lambda_builders/workflows/nodejs_npm/actions.py
```diff
@@ -0,0 +1,92 @@
...
+            self.osutils = OSUtils()
+
+        self.subprocess_npm = subprocess_npm
+
+        if self.subprocess_npm is None:
+            self.subprocess_npm = SubprocessNpm(self.osutils)
+
+    def execute(self):
+        try:
+            if not self.osutils.file_exists(self.manifest_path):
```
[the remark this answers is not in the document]
gojko: should I remove the guard then?

aws_lambda_builders/workflows/nodejs_npm/actions.py
```diff
@@ -0,0 +1,92 @@
...
+                raise ActionFailedError('package.json not found in: %s' % self.manifest_path)
+
+            package_path = "file:%s" % self.osutils.abspath(self.osutils.dirname(self.manifest_path))
```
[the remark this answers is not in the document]
gojko: fixed this

aws_lambda_builders/workflows/nodejs_npm/actions.py
```diff
@@ -0,0 +1,92 @@
...
+        self.subprocess_npm = subprocess_npm
+
+        if self.subprocess_npm is None:
+            self.subprocess_npm = SubprocessNpm(self.osutils)
+
+    def execute(self):
+        try:
+            LOG.debug("NODEJS installing in: %s from: %s", self.artifacts_dir, self.manifest_path)
+
+            if not self.osutils.file_exists(self.osutils.joinpath(self.artifacts_dir, 'package.json')):
```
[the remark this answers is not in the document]
gojko: this is is a guard against someone flipping the order of actions in the future. It's not a massive performance hit (just checking for file existence), so I'd prefer to leave it unless you strongly object.

aws_lambda_builders/workflows/nodejs_npm/actions.py
```diff
@@ -0,0 +1,92 @@
...
+                raise ActionFailedError('package.json not found in: %s' % self.artifacts_dir)
+
+            self.subprocess_npm.main(
+                    ['install', '-q', '--no-audit', '--no-save', '--production'],
```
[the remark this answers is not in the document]
gojko: `--production` is right 99% of the time. I doubt anyone would ever want to send up unit testing tools and linting tools to Lambda. We will deal with the remaining 1% by letting users override these parameters in an env var, so that they could exclude optional dependencies. this would let them also change and install dev dependencies if they really want.

aws_lambda_builders/workflows/nodejs_npm/npm.py
```diff
@@ -0,0 +1,62 @@
...
+            env_vars = self._osutils.environ()
+
+        invoke_npm = [self.npm_exe] + args
+
+        LOG.debug("executing NPM: %s", invoke_npm)
+
+        p = self._osutils.popen(invoke_npm,
+                                stdout=self._osutils.pipe,
+                                stderr=self._osutils.pipe,
+                                env=env_vars,
```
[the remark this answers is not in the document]
gojko: fixed this in the PR

aws_lambda_builders/workflows/python_pip/utils.py
```diff
@@ -86,3 +87,12 @@ def mtime(self, path):
     @property
     def pipe(self):
         return subprocess.PIPE
+
+    def find_executable(self, execname):
+        return distutils.spawn.find_executable(execname)
```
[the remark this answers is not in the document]
gojko: ok, I removed it from the PR.

aws_lambda_builders/workflows/nodejs_npm/npm.py
```diff
@@ -0,0 +1,62 @@
...
+
+class SubprocessNpm(object):
+
+    def __init__(self, osutils=None, npm_exe=None):
+        if osutils is None:
+            osutils = OSUtils()
+        self._osutils = osutils
+
+        if npm_exe is None:
+            npm_exe = osutils.find_executable('npm')
```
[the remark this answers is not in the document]
gojko: removed from the PR

aws_lambda_builders/workflows/nodejs_npm/npm.py
```diff
@@ -0,0 +1,62 @@
...
+        if not osutils.file_exists(npm_exe):
+            raise NpmNotFoundError(npm_exe)
+
+        self.npm_exe = npm_exe
+
+    def main(self, args, cwd=None, env_vars=None):
+        if env_vars is None:
+            env_vars = self._osutils.environ()
+
+        invoke_npm = [self.npm_exe] + args
```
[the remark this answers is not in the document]
gojko: done!

aws_lambda_builders/workflows/nodejs_npm/actions.py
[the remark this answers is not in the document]
sanathkr: Yeah, I'd say for now copy the osutils to nodejs_npm workflow and retain only functions you care about. We'll clean this debt up later.

aws_lambda_builders/workflows/nodejs_npm/actions.py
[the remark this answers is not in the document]
sanathkr: For now, retain the guard in one action, and not in both. If you want, you can even create a separate action that checks for package.json only and does nothing else.

aws_lambda_builders/workflows/nodejs_npm/actions.py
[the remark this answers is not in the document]
sanathkr: okay, cool. Let's stick with this then. 

One downside of this is, when we implement hot-reloading we have to modify this implementation. But that's for later :)

aws_lambda_builders/workflows/nodejs_npm/actions.py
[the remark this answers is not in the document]
sanathkr: I would stick with CopySource action if you want. You can always set the exclude list to empty which will not exclude anything.

I like to use CopySource action because in the future we might implement `rsync` like capabilities to it, which might copy only the files that changed and not everything. This will come in super handy for hot-reloading.

aws_lambda_builders/workflows/nodejs_npm/actions.py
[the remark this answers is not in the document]
sanathkr: I would skip this from here because it duplicate logic which something might take dependency on later.

aws_lambda_builders/workflows/nodejs_npm/actions.py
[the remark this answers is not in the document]
sanathkr: Perfect! Let's stick with this then.

tests/integration/workflows/nodejs_npm/test_nodejs_npm.py
sanathkr: Perfect. So we will any print the output from NPM here. That should give users enough information to fix their manifest.

tests/integration/workflows/nodejs_npm/testdata/npm-deps/package.json
sanathkr: perfect!

aws_lambda_builders/workflows/nodejs_npm/npm.py
```diff
@@ -0,0 +1,53 @@
...
+        if osutils is None:
+            osutils = OSUtils()
+        self.osutils = osutils
+
+        if npm_exe is None:
+            npm_exe = 'npm'
+
+        self.npm_exe = npm_exe
+
+    def main(self, args, cwd=None):
```
sanathkr: Actually, its a bit weird to call this `main()`. How about `run()` instead?

aws_lambda_builders/workflows/nodejs_npm/actions.py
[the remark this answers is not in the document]
gojko: resolved in https://github.com/gojko/aws-lambda-builders/commit/cec44aea77e1b8d557ac501927c8d1506a4b8647

aws_lambda_builders/workflows/nodejs_npm/npm.py
```diff
@@ -0,0 +1,53 @@
...
+
+        if not isinstance(args, list):
+            raise NpmExecutionError('args must be a list')
```
sanathkr: So I would raise `ValueError` instead here. Users dont' call this method directly, so they don't need to know about the error when the caller did something wrong. `ValueError` is not caught any where and it would crash the process. 

I see `NpmExecutionError` as equivalent of HTTP 4xx (user error) and `ValueError` as equivalent of HTTP 5xx (internal error)

aws_lambda_builders/workflows/nodejs_npm/actions.py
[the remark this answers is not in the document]
gojko: ok, should I leave some kind of todo note there?

aws_lambda_builders/workflows/nodejs_npm/actions.py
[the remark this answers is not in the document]
gojko: resolved with https://github.com/gojko/aws-lambda-builders/commit/4a00c1e05cc58b57ffdbf1de2d1250ed9a534881

aws_lambda_builders/workflows/python_pip/utils.py
```diff
@@ -10,6 +10,7 @@
 import shutil
```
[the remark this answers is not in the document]
gojko: resolved with https://github.com/gojko/aws-lambda-builders/commit/cec44aea77e1b8d557ac501927c8d1506a4b8647

aws_lambda_builders/workflows/nodejs_npm/actions.py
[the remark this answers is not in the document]
gojko: also, in theory, we can do hot reloading to check if package.json got modified. if yes, then rerun the whole tar thing. if only a file within the project got modified, and not package.json, then it's pretty safe to assume that packaging didn't change and dependencies did not change as well, so it's OK to just safe-reload that file.

aws_lambda_builders/workflows/nodejs_npm/npm.py
```diff
@@ -0,0 +1,53 @@
...
+        if osutils is None:
+            osutils = OSUtils()
+        self.osutils = osutils
+
+        if npm_exe is None:
+            npm_exe = 'npm'
+
+        self.npm_exe = npm_exe
+
+    def main(self, args, cwd=None):
```
sanathkr: you can simplify this further by using this neat Python trick:

`run(*args, cwd=None)` - then your caller can call this method using `run("install", "--production", cwd='blah')`, but you will end up getting a list value within the `args` variable that you can call Popen with.

Not that you should make this change, but felt like a neat Py trick you might like :)

aws_lambda_builders/workflows/nodejs_npm/npm.py
```diff
@@ -0,0 +1,53 @@
...
+
+class SubprocessNpm(object):
+
+    def __init__(self, osutils=None, npm_exe=None):
+        if osutils is None:
+            osutils = OSUtils()
+        self.osutils = osutils
+
+        if npm_exe is None:
+            npm_exe = 'npm'
```
TheSriram: In the future if there are multiple paths of npm on your system, how would you imagine passing them in? or maybe thats not a big concern.

I started a PR here: https://github.com/awslabs/aws-lambda-builders/pull/35 that would help with that. But I'm going to take a step back and maybe write a mini design doc to get started on it. I was curious to know if you had any comments on that?

aws_lambda_builders/workflows/nodejs_npm/npm.py
```diff
@@ -0,0 +1,62 @@
...
+LOG = logging.getLogger(__name__)
+
+
+class NpmError(Exception):
+    pass
+
+
+class NpmNotFoundError(NpmError):
+    def __init__(self, npm_path):
+        super(NpmNotFoundError, self).__init__(
```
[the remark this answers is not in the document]
gojko: resolved in https://github.com/gojko/aws-lambda-builders/commit/3e945b281e63e82d83e0fe3fe8f71b9c50f978aa

aws_lambda_builders/workflows/nodejs_npm/npm.py
gojko: resolved in https://github.com/gojko/aws-lambda-builders/commit/3e945b281e63e82d83e0fe3fe8f71b9c50f978aa

aws_lambda_builders/workflows/nodejs_npm/npm.py
gojko: resolved in https://github.com/gojko/aws-lambda-builders/commit/3e945b281e63e82d83e0fe3fe8f71b9c50f978aa

aws_lambda_builders/workflows/nodejs_npm/npm.py
gojko: thanks for the tip. I'd prefer to leave it like this because it's a single list to pass on, it will make it easier to switch to user-provided values when we introduce the env var to override default arguments. (as in that case the first element would have to stay the same, but the users will be able to override other values)

gojko:
It looks like the overall design of the code is OK? I'll then add unit tests and we can start closing off the first release of this. 

Regarding NPM, we could somehow split out integration tests that require NPM from the rest, so you can run them on two different integration test environments. Let me know if that's an option.

aws_lambda_builders/workflows/nodejs_npm/actions.py
```diff
@@ -0,0 +1,85 @@
...
+from .npm import SubprocessNpm, NpmExecutionError
+
+LOG = logging.getLogger(__name__)
+
+
+class NodejsNpmPackAction(BaseAction):
+
+    NAME = 'NpmPack'
+    DESCRIPTION = "Packaging source using NPM"
+    PURPOSE = Purpose.COMPILE_SOURCE
```
sanathkr: is this really Compile Source? I think this is `COPY_SOURCE`. Compile is for things like golang which actually does the compilation

aws_lambda_builders/workflows/nodejs_npm/workflow.py
```diff
@@ -0,0 +1,56 @@
...
+import hashlib
+
+from aws_lambda_builders.workflow import BaseWorkflow, Capability
+from aws_lambda_builders.actions import CopySourceAction
+
+from .actions import NodejsNpmPackAction, NodejsNpmInstallAction
+from .utils import OSUtils
+
+
+def unique_dir(manifest_path):
```
sanathkr: why do you need this? for hot-reloading?

If you need a unique directory, you can use `tempfile.mkdtemp()` within the scratch directory.

gojko: I need this so that if you run multiple workflows for various lambda functions and give me the same scratch dir, I don't end up compiling overlapping packages :) you mentioned wanting to support hot reloading later, so I thought that it's better to have a stable name that depends on the manifest, especially as the copysource command will be wired to that. 

so if a workflow runs for the same manifest, it will try to reuse the same directory to prepare the sources. the dependencies won't be installed here, but in the artifact dir directly

aws_lambda_builders/workflows/nodejs_npm/actions.py
gojko: ok, fixed in https://github.com/gojko/aws-lambda-builders/commit/2542cdbbcbaf4afe7ce17b1749a75804b96b1848

aws_lambda_builders/workflows/nodejs_npm/workflow.py
sanathkr: First requirement of running multiple lambdas on same scratch_dir won't happen. Caller will setup diff scratch dir for each function and each execution. 

I am okay with the stable name, but the directories are cleared & re-created on each run. So it won't really do much help here. 

I would like to skip this for now as we don't have a use-case now and revisit it later. Thoughts?

aws_lambda_builders/workflows/nodejs_npm/workflow.py
```diff
@@ -0,0 +1,56 @@
...
+
+    EXCLUDED_FILES = (".aws-sam")
+
+    def __init__(self,
+                 source_dir,
+                 artifacts_dir,
+                 scratch_dir,
+                 manifest_path,
+                 runtime=None,
+                 osutils=None,
```
sanathkr: good one

sanathkr:
Everything looks good, except for the unique_dir comment. Do start writing unit tests. I will create a separate Travis/Appveyor PR to run integ tests

sanathkr:
Also, is there a min version of Nodejs & NPM that this implementation assumes?

aws_lambda_builders/workflows/nodejs_npm/workflow.py
gojko: OK - I misunderstood the purpose of `scratch_dir` then. If it's a new directory every time, it makes things very easy. I'll just hard-code the name of the dir. I fixed this in https://github.com/gojko/aws-lambda-builders/commit/d183932559d05f19899a6176d2d65f5b06f2abf5 - let me know if this makes sense now.

gojko:
> Also, is there a min version of Nodejs & NPM that this implementation assumes?

this just runs the NPM executable, so as long as there is one on the path, it will work. NPM tends to ignore parameters that it does not understand, so `--no-audit` and `--no-save` will have no effect on earlier versions where they are not supported.

aws_lambda_builders/workflows/nodejs_npm/workflow.py
sanathkr: Cool, that makes sense

aws_lambda_builders/workflows/nodejs_npm/npm.py
```diff
@@ -0,0 +1,49 @@
...
+import logging
+
+from aws_lambda_builders.exceptions import LambdaBuilderError
+
+from .utils import OSUtils
+
+LOG = logging.getLogger(__name__)
+
+
+class NpmExecutionError(LambdaBuilderError):
```
sanathkr: actually, you misunderstood my previous comment. Don't inherit directly from LambdaBuilderError. Instead just follow the pattern of having your custom base error class that does the interpolation, and subclasses just set the MESSAGE attribute.

In this case, since you basically have only one exception, you can consider this to be the base error. Something like this:

```
class NpmExecutionError(Exception):
  MESSAGE = "NPM Failed: {message}"
  def __init__(self, **kwargs):
             Exception.__init__(self, self.MESSAGE.format(**kwargs))
```

aws_lambda_builders/workflows/nodejs_npm/actions.py
```diff
@@ -0,0 +1,85 @@
...
+
+import logging
+from aws_lambda_builders.actions import BaseAction, Purpose, ActionFailedError
+from .utils import OSUtils
+from .npm import SubprocessNpm, NpmExecutionError
+
+LOG = logging.getLogger(__name__)
+
+
+class NodejsNpmPackAction(BaseAction):
```
sanathkr: docstrings

aws_lambda_builders/workflows/nodejs_npm/actions.py
```diff
@@ -0,0 +1,85 @@
...
+
+    NAME = 'NpmPack'
+    DESCRIPTION = "Packaging source using NPM"
+    PURPOSE = Purpose.COPY_SOURCE
+
+    def __init__(self, artifacts_dir, scratch_dir, manifest_path, runtime, osutils=None, subprocess_npm=None):
```
sanathkr: docstrings

aws_lambda_builders/workflows/nodejs_npm/npm.py
gojko: ok, done this in https://github.com/gojko/aws-lambda-builders/commit/d141caf9b61c39d2aeab763367921b15a65c2568

gojko:
@sanathkr I've added the documentation and some more unit tests. Is there anything else missing for the first version, or are we ready to go with this?

aws_lambda_builders/workflows/nodejs_npm/workflow.py
```diff
@@ -38,11 +43,21 @@ def __init__(self,
...
 
+        npm_pack = NodejsNpmPackAction(tar_dest_dir,
+                                       scratch_dir,
+                                       manifest_path,
+                                       osutils=osutils,
+                                       subprocess_npm=subprocess_npm)
+
+        npm_install = NodejsNpmInstallAction(artifacts_dir,
+                                             subprocess_npm=subprocess_npm)
         self.actions = [
```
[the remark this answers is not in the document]
gojko: thanks

tests/unit/workflows/nodejs_npm/test_npm.py
```diff
@@ -0,0 +1,79 @@
+
+from unittest import TestCase
+from mock import patch
+
+from aws_lambda_builders.workflows.nodejs_npm.npm import SubprocessNpm, NpmExecutionError
+
+
+class FakePopen:
```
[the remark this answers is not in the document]
gojko: seemed like more trouble than what it was worth for such a simple case. I'll leave it as it is if you don't mind

sanathkr:
@gojko  I just pushed a commit to your branch fixing appveyor & travis - https://github.com/awslabs/aws-lambda-builders/pull/44/commits/7eaad47750fa586b2509078816f8f7dc1dad9570

Also, can you update the PR to send against `develop` branch instead of `master`?

The code looks good to me! I think we are good to merge once the integ tests pass

gojko:
> Also, can you update the PR to send against `develop` branch instead of `master`?

I've updated the base to merge into develop using the github UI, not sure if that will do the trick or not. if this looks wrong to you, i'll create a completely clean PR

sanathkr:
Yup, that's good enough :)

Looks like the integ tests are failing on Windows because its not able to find node/npm :(

gojko:
I wonder if npm is a separate package you need to install there? it looks like node is installed, but it's not finding NPM.

sanathkr:
that's surprising. I tested the appveyor config here - https://github.com/sanathkr/aws-lambda-builders/commit/713d5b2434443682ee05145b1c0a0ac3383f326e.  It did print npm version - https://ci.appveyor.com/project/sanathkr/aws-lambda-builders/builds/20767787/job/5k44u8p3pujh7kg3

aws_lambda_builders/workflows/nodejs_npm/actions.py
```diff
@@ -0,0 +1,110 @@
...
+
+    """
+    A Lambda Builder Action that packages a Node.js package using NPM to extract the source and remove test resources
+    """
+
+    NAME = 'NpmPack'
+    DESCRIPTION = "Packaging source using NPM"
+    PURPOSE = Purpose.COPY_SOURCE
+
+    def __init__(self, artifacts_dir, scratch_dir, manifest_path, osutils, subprocess_npm):
```
jfuss: We are inheriting but not calling super. Would be a good practice to start especially with this being a library.

aws_lambda_builders/workflows/nodejs_npm/actions.py
```diff
@@ -0,0 +1,110 @@
...
+        self.osutils = osutils
+        self.subprocess_npm = subprocess_npm
+
+    def execute(self):
+        """
+        Runs the action.
+
+        :raises lambda_builders.actions.ActionFailedError: when NPM packaging fails
+        """
+        try:
```
jfuss: Does this try/expect need to be this wide? Best practice is to scope as small as is reasonable.

aws_lambda_builders/workflows/nodejs_npm/actions.py
```diff
@@ -0,0 +1,110 @@
...
+            package_path = "file:{}".format(self.osutils.abspath(self.osutils.dirname(self.manifest_path)))
```
jfuss: @sanathkr Will this `osutils.abspath` and `osutils.dirname` be replaced with the PathResolve changes? We are doing path things here and using `os.path` style operations which makes this harder to read. If instead we had actually objects (like pathlib), these are easier to maintain.

Or even if `osutils` returned `Path`s, you will be able to operate on them normally.

aws_lambda_builders/workflows/nodejs_npm/npm.py
```diff
@@ -0,0 +1,90 @@
...
+
+        :type npm_exe: str
+        :param npm_exe: Path to the NPM binary. If not set,
+            the default executable path npm will be used
+        """
+        self.osutils = osutils
+
+        if npm_exe is None:
+            if osutils.is_windows():
+                npm_exe = 'npm.cmd'
```
jfuss: Does npm have a `npm.exe` at all?

aws_lambda_builders/workflows/nodejs_npm/utils.py
```diff
@@ -0,0 +1,40 @@
...
+Commonly used utilities
+"""
+
+import os
+import tarfile
+import subprocess
+import sys
+
+
+class OSUtils(object):
```
jfuss: Do we really need all the wrapping around os.path? We should really move on up into pathlib and operate on objects

aws_lambda_builders/workflows/nodejs_npm/utils.py
```diff
@@ -0,0 +1,40 @@
...
+        return subprocess.PIPE
+
+    def dirname(self, path):
+        return os.path.dirname(path)
+
+    def abspath(self, path):
+        return os.path.abspath(path)
+
+    def is_windows(self):
+        return sys.platform.lower().startswith('win')
```
jfuss: What about cygwin?
https://docs.python.org/3/library/sys.html#sys.platform

Can we follow was SAM CLI does for this already: https://github.com/awslabs/aws-sam-cli/blob/develop/samcli/lib/samlib/cloudformation_command.py#L33

tests/functional/workflows/nodejs_npm/test_utils.py
```diff
@@ -0,0 +1,81 @@
...
+
+import sys
+
+import os
+
+import shutil
+
+import tempfile
+
+from aws_lambda_builders.workflows.nodejs_npm import utils
```
jfuss: These imports are not following PEP8

tests/functional/workflows/nodejs_npm/test_utils.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+@pytest.fixture
+def osutils():
+    return utils.OSUtils()
+
+
+class TestOSUtils(object):
```
jfuss: This test class is using pytest but others are using unittest. Can we keep everything on one framework (preferable unittest)

tests/unit/workflows/nodejs_npm/test_workflow.py
```diff
@@ -0,0 +1,28 @@
...
+    """
+    the workflow requires an external utility (npm) to run, so it is extensively tested in integration tests.
+    this is just a quick wiring test to provide fast feedback if things are badly broken
+    """
+
+    def test_workflow_sets_up_npm_actions(self):
+
+        workflow = NodejsNpmWorkflow("source", "artifacts", "scratch_dir", "manifest")
+
+        assert len(workflow.actions) == 3
```
jfuss: use self.assert* instead

tests/unit/workflows/nodejs_npm/test_npm.py
```diff
@@ -0,0 +1,92 @@
...
+        self.osutils.popen.assert_called_with(['/a/b/c/npm.exe', 'pack', '-q'],
+                                              cwd='/a/cwd', stderr='PIPE', stdout='PIPE')
+
+    def test_returns_popen_out_decoded_if_retcode_is_0(self):
+
+        self.popen.out = b'some encoded text\n\n'
+
+        result = self.under_test.run(['pack'])
+
+        assert result == 'some encoded text'
```
jfuss: self.assert*

There are some others that have plain asserts, please update those as well

tests/unit/workflows/nodejs_npm/test_workflow.py
```diff
@@ -0,0 +1,28 @@
+
+from unittest import TestCase
+
+from aws_lambda_builders.workflows.nodejs_npm.workflow import NodejsNpmWorkflow
+
+from aws_lambda_builders.workflows.nodejs_npm.actions import NodejsNpmPackAction, NodejsNpmInstallAction
+
+from aws_lambda_builders.actions import CopySourceAction
```
jfuss: PEP 8 imports

tests/functional/workflows/nodejs_npm/test_utils.py
sanathkr: actually, this test class was copied from `python_pip/utils.py`, which was copied from Chalice's code. So this has to be cleaned up anyway. I would not spend more time diverging the two if possible.

aws_lambda_builders/workflows/nodejs_npm/npm.py
sanathkr: no it doesn't unfortunately. I verified on a windows box

aws_lambda_builders/workflows/nodejs_npm/actions.py
sanathkr: yes exactly. they will all go away when pathresolver can give you a pathlib object back

sanathkr:
Just realized, we never wrote a design document for this. This is just a simple markdown file that explains what the workflow does, like this - https://github.com/awslabs/aws-lambda-builders/blob/develop/aws_lambda_builders/workflows/python_pip/DESIGN.md.

Goal is to let other implementors understand why certain decisions were made. It helps us set context for future design changes.

If you got time, can you add a short design document? If you feel the PR is already big, we can do it in a separate PR later

aws_lambda_builders/workflows/nodejs_npm/actions.py
gojko: ok, resolved with https://github.com/gojko/aws-lambda-builders/commit/53f9d349da988abf445ce06b4d291240c69bbb7c

aws_lambda_builders/workflows/nodejs_npm/utils.py
gojko: I just copied this design from the python_pip builder implementation, so I guess this is for a more general discussion. Happy to change it if you decide internally that you want to take a different direction, just let me know

aws_lambda_builders/workflows/nodejs_npm/utils.py
gojko: I've changed the OS check to be the same as in the SAM CLI here https://github.com/gojko/aws-lambda-builders/commit/2c5d7807a95f38729ce0dc1dc4f4bf776b3b7a9c. Checking for .exe doesn't matter here as we know it's a .cmd, but I guess this can be handled in a more generic way as part of the path resolver PR (https://github.com/awslabs/aws-lambda-builders/pull/35)

tests/functional/workflows/nodejs_npm/test_utils.py
gojko: can you give me a bit more guidance on what to change? I read through the PEP8 import section and can't find anything wrong with this, so I'm missing some context here (I'm not actively working with python)

tests/functional/workflows/nodejs_npm/test_utils.py
gojko: resolved with https://github.com/gojko/aws-lambda-builders/commit/ed23979bc52ed87b6b2e1e3700d7051dcbd4b0f1

tests/unit/workflows/nodejs_npm/test_workflow.py
gojko: resolved with https://github.com/gojko/aws-lambda-builders/commit/fcd47a9753ed88e0995a26493882aa7b8fbe3402

tests/unit/workflows/nodejs_npm/test_npm.py
gojko: resolved with https://github.com/gojko/aws-lambda-builders/commit/fcd47a9753ed88e0995a26493882aa7b8fbe3402

gojko:
> If you got time, can you add a short design document? If you feel the PR is already big, we can do it in a separate PR later

I added this here https://github.com/awslabs/aws-lambda-builders/pull/44/commits/09d41909bb0113f4824f562ed1d727b6cb637704

aws_lambda_builders/workflows/nodejs_npm/utils.py
jfuss: I didn't realize we already do this. We can refactor these things later.

aws_lambda_builders/workflows/nodejs_npm/actions.py
jfuss: Awesome. Just wanted to make sure

aws_lambda_builders/workflows/nodejs_npm/actions.py
jfuss: I see. That makes sense. Just wanted to make sure we have this reasonably scoped.

tests/functional/workflows/nodejs_npm/test_utils.py
jfuss: Appologizes on not providing additional context.

So the Style for Python is typically referred to PEP8. In this imports, should be grouped into 3 categories.

```
# imports for modules in the Python language (i.e. os or tempfile)
import tempfile

# Imports for modules of your dependencies (anything specified in the requirements.txt)
import pytest

# imports for modules that are within you module (this is anything in aws_lambda_builders)
```

So this file's imports would look like 
```
import os
import shutil
import sys
import tempfile

import pytest

from aws_lambda_builders.workflows.nodejs_npm import utils

```

gojko: thanks. This should do the trick: https://github.com/awslabs/aws-lambda-builders/pull/44/commits/6863067023cbf28d5b4dfc68061a8f6ca41654b3

tests/unit/workflows/nodejs_npm/test_workflow.py
gojko: resolved in https://github.com/awslabs/aws-lambda-builders/pull/44/commits/6863067023cbf28d5b4dfc68061a8f6ca41654b3

aws_lambda_builders/workflows/nodejs_npm/npm.py
```diff
@@ -0,0 +1,90 @@
...
+    easy to consume execution results.
+    """
+
+    def __init__(self, osutils, npm_exe=None):
+        """
+        :type osutils: aws_lambda_builders.workflows.nodejs_npm.utils.OSUtils
+        :param osutils: An instance of OS Utilities for file manipulation
+
+        :type npm_exe: str
+        :param npm_exe: Path to the NPM binary. If not set,
```
TheSriram: I'm ok with keeping this here for the PR, but eventually we need the workflow to start with absolute paths to dependency managers, language etc to ensure a correct build?

I have started a PR here with design notes: https://github.com/awslabs/aws-lambda-builders 

Let me know on your thoughts!

TheSriram:
The PR looks good to me. My only concern was to be able plugin path resolvers and path validators appropriately without much effort after this gets merged, which is what I have commented.

aws_lambda_builders/workflows/nodejs_npm/DESIGN.md
```diff
@@ -0,0 +1,126 @@
...
+
+To speed up Lambda startup time and optimise usage costs, the correct thing to
+do in most cases is just to package up production dependencies. During development 
+work we can expect that the local `node_modules` directory contains all the 
+various dependency types, and NPM does not provide a way to directly identify
+just the ones relevant for production. To identify production dependencies, 
+this packager needs to copy the source to a clean temporary directory and re-run
+dependency installation there.
+  
+It's reasonable to expect that some developers will not carefully separate
```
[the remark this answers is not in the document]
gojko: I guess the option Sanath and I discussed is more to do with optional dependencies than dev dependencies, so I'll clarify that in the design doc. 

NPM has the concept of optional dependencies, very frequently used for excluding aws-sdk for from the deployment package, but leaving it as a non-dev dependency for small lambdas. AWS SDK is already available on the Lambda VM, so this makes deployment significantly faster for single-file lambdas, for example. I dislike this because it's not 100% clear what version of the API gets used, but people do this frequently enough that we should handle it in some way. the proposal was to use an env var to override NPM arguments.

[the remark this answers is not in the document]
gojko: I've updated this in https://github.com/awslabs/aws-lambda-builders/pull/44/commits/fdc2707266b97add790aa7b1a653bfef832685bf

aws_lambda_builders/workflows/nodejs_npm/DESIGN.md
```diff
@@ -0,0 +1,126 @@
...
+is as follows.
+
+#### Step 1: Prepare a clean copy of the project source files
+
+Execute `npm pack` to perform project-specific packaging using the supplied
+`package.json` manifest, which will automatically exclude temporary files, 
+test resources and other source files unnecessary for running in a production 
+environment.
+
+This will produce a `tar` archive that needs to be unpacked into the artefacts directory.
```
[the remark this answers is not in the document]
gojko: thanks, fixed in https://github.com/awslabs/aws-lambda-builders/pull/44/commits/fdc2707266b97add790aa7b1a653bfef832685bf

aws_lambda_builders/workflows/nodejs_npm/npm.py
gojko: @TheSriram I guess as long as you pass an executable into the workflow as a standard parameter, we can easily pass it to the NPM wrapper there, so it should be a single line of code change once it is ready.

aws_lambda_builders/workflows/nodejs_npm/DESIGN.md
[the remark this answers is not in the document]
jfuss: That makes sense. Thanks for clarifing it.

sanathkr:
@gojko  Please confirm this contribution is under the terms of the Apache 2.0 license. Thanks.

gojko:
@sanathkr  I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #47, opened 2018-12-05
WIP: Golang workflow using "dep" dependency manager 

sanathkr:
*Description of changes:*
Starting with a design document to communicate what I am building. Will add implementation soon.

### DIFF
--- a/aws_lambda_builders/workflows/go_dep/DESIGN.md
+++ b/aws_lambda_builders/workflows/go_dep/DESIGN.md
@@ -1,0 +1,19 @@
+### Go-Dep Lambda function builder
+Building Lambda functions written in Golang is very straightforward. We use the `dep` dependency manager and 
+`go` tool to install dependencies and cross-compile the function for architecture that Lambda supports.
+
+
+The workflow does the following:
+1. Check for manifest **Gopkg.toml**
+2. Install dependencies using `dep ensure -v`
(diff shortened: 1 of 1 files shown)


aws_lambda_builders/workflows/go_dep/DESIGN.md
```diff
@@ -0,0 +1,19 @@
+### Go-Dep Lambda function builder
+Building Lambda functions written in Golang is very straightforward. We use the `dep` dependency manager and 
+`go` tool to install dependencies and cross-compile the function for architecture that Lambda supports.
+
+
+The workflow does the following:
+1. Check for manifest **Gopkg.toml**
+2. Install dependencies using `dep ensure -v`
+3. Compile using `GOOS=linux GOARCH=amd64 go build -ldflags="-s -w" -o $ARTIFACT_DIR/$HANDLER $SOURCE_DIR`
```
roryj: One question I have is what are the goals for the `sam build` command. I would love to use a command like this in my workflow, but what if I already have well defined build steps as part of a Makefile in my repo? Make and Makefiles are a pretty common construct within Golang. Could we do something simpler where we just call `make build`? That way if developers want slightly different arguments, or multiple steps they can still use the sam build api. Maybe this make file could be generated if it is not present the first time.

PR #48, opened 2018-12-05
Design: Path Resolver

TheSriram:
First pass of design to get comments from the community.

Introduction of two pluggable components.
 * Path Resolver
 * Runtime Validator

### DIFF
--- a/designs/Path_Resolver_DESIGN.md
+++ b/designs/Path_Resolver_DESIGN.md
@@ -66,16 +66,40 @@ class BaseWorkflow(object):
 		...
 		...
 	
-	def get_executable():
+	def get_executable(self):
 		return PathResolver(language=x, runtime=x.y).path
 	
-	def get_validator():
+	def get_validator(self):
 		return RuntimeValidator(runtime_path=/usr/bin/x.y, language=x)
+		
+	@sanitize(executable_path=self.get_executable(),validator=self.get_validator())
+	def run(self):
 ```
 
 This way we have de-coupled validtion of the path and the actual finding of the path to the executable. Both of these methods can be over-riden in any workflow that subclasses BaseWorkflow.
 
+There will be default implementations of ```PathResolver``` and ```RuntimeValidator``` that can still be used by workflow authors if they dont want to specialize them. 
+
+A decorator on top of run, would actually execute validation on the resolved path to ensure that its safe to actually start the workflow.
+
 ### Implementation
+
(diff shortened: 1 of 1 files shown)


designs/Path_Resolver_DESIGN.md
```diff
@@ -0,0 +1,118 @@
...
+
+The general tenets for this would look like as follows.
+
+1. Every workflow needs to have a PathResolver and a Validator. The resolver can be a custom version or use the bundled opioniated resolver.
+2. The workflow needs to fail if the resolved executable path does not match the supplied runtime version. This fails quickly and stops incorrect artifacts from being built.
+3. The resolved executable path should be available to every action.
+
+### Open questions
+
+* Should we constrain the interface for the PathResolver with an abstract base class?
```
jfuss: What is the pros and cons about this? I think we need an defined interface, at the very least.

designs/Path_Resolver_DESIGN.md
```diff
@@ -0,0 +1,118 @@
...
+We will have a path resolver class, whose functionality is inherently simple to begin with. It has a candidate list of locations to look for the executable depending upon the runtime and language, and return the path.
+
+```python
+class PathResolver(object):
+     def __init__(self, language, runtime):
+        self.language = language
+        self.runtime = runtime
+        self.executables = [self.runtime, self.language]
+        
+     def path(self):
```
jfuss: Is the path allowed to be symlinks or should it be a fully resolved path?

TheSriram: symlinks should be allowed in my opinion. (helps users have a quick fix to their path, with just changing symlinks)

TheSriram: Path need not be fully resolved here, the chosen path will be exec'd by the validator to make a decision on the validity of the path, and if it matches the runtime.

designs/Path_Resolver_DESIGN.md
TheSriram: Pros of this, is that any implementations will need to have appropriate methods defined, so that we dont run into funkiness at runtime. But yeah, I'm just proposing abc as a way to solidly lock down the interface.

awood45:
I'm going to look to add Ruby to this after we merge the Ruby build logic. Basically, the following shell one-liner will check compatibility with the `ruby2.5` runtime:

```shell
ruby -e "unless RUBY_VERSION.match(/2\.5\.\d/); exit(1); end"
```

PR #50, opened 2018-12-10
fix: Create scratch_dir in builder.py and bump version to 0.0.4.dev1

jfuss:
*Issue #, if available:*

*Description of changes:*
Create the scratch directory in the builder so each workflow can safely assume the directory is available to use.


requirements/base.txt
```diff
@@ -1 +1 @@
-six~=1.11.0
+six~=1.11
```
sanathkr: thanks!

tests/functional/test_cli.py
```diff
@@ -19,6 +19,7 @@ def setUp(self):
 
         self.source_dir = tempfile.mkdtemp()
         self.artifacts_dir = tempfile.mkdtemp()
+        self.scratch_dir = "/tmp/scratch"
```
sanathkr: windows

jfuss: I would expect that to fail in appveyor through.

tests/functional/test_builder.py
```diff
@@ -19,6 +19,7 @@ def setUp(self):
 
         self.source_dir = tempfile.mkdtemp()
         self.artifacts_dir = tempfile.mkdtemp()
+        self.scratch_dir = "/tmp/scratch"
```
sanathkr: windows?

You should really use a `tempfile` here instead of /tmp/scratch because the test will be creating this directory. 

You can use something like `os.path.join(tempfile.mkdtemp(), "scratch")` to save yourself the headache of finding a temp filename yourself

jfuss: I didn't think of this. I first thought of doing `tempfile.mkdtemp()` but that creates the directory, which is why I did what I did. Will update.

PR #53, opened 2018-12-11
Extend Node.js NPM support for .npmrc configuration

simalexan:
*Description of changes:*
NPM can be configured to use proxies or local company repositories using a local file, `.npmrc`. 
The packaging process from step 1 normally excludes this file, so it needs to be copied before dependency installation, and then removed. 

### DIFF
--- a/aws_lambda_builders/workflows/nodejs_npm/actions.py
+++ b/aws_lambda_builders/workflows/nodejs_npm/actions.py
@@ -111,3 +111,87 @@ class NodejsNpmInstallAction(BaseAction):
 
         except NpmExecutionError as ex:
             raise ActionFailedError(str(ex))
+
+class NodejsNpmrcCopyAction(BaseAction):
+
+    """
+    A Lambda Builder Action that copies NPM config file .npmrc
+    """
+
+    NAME = 'CopyNpmrc'
+    DESCRIPTION = "Copying configuration from .npmrc"
+    PURPOSE = Purpose.COPY_SOURCE
+
+    def __init__(self, artifacts_dir, source_dir, osutils):
+        """
+        :type artifacts_dir: str
+        :param artifacts_dir: an existing (writable) directory with project source files.
+            Dependencies will be installed in this directory.
+
+        :type source_dir: str
+        :param source_dir: directory containing project source files.
+
+        :type osutils: aws_lambda_builders.workflows.nodejs_npm.utils.OSUtils
+        :param osutils: An instance of OS Utilities for file manipulation
+        """
+
+        super(NodejsNpmrcCopyAction, self).__init__()
+        self.artifacts_dir = artifacts_dir
+        self.source_dir = source_dir
+        self.osutils = osutils
+
+    def execute(self):
+        """
+        Runs the action.
+
+        :raises lambda_builders.actions.ActionFailedError: when .npmrc copying fails
+        """
+
+        try:
+            npmrc_path = self.osutils.joinpath(self.source_dir, ".npmrc")
+            if self.osutils.file_exists(npmrc_path):
+                LOG.debug(".npmrc copying in: %s", self.artifacts_dir)
+                self.osutils.copy_file(npmrc_path, self.artifacts_dir)
+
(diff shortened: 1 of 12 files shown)


aws_lambda_builders/workflows/nodejs_npm/actions.py
```diff
@@ -111,3 +111,87 @@ def execute(self):
...
+        try:
+            npmrc_path = self.osutils.joinpath(self.source_dir, ".npmrc")
+            if self.osutils.file_exists(npmrc_path):
+                LOG.debug(".npmrc copying in: %s", self.artifacts_dir)
+                self.osutils.copy_file(npmrc_path, self.artifacts_dir)
+
+        except OSError as ex:
+            raise ActionFailedError(str(ex))
+
+class NodejsCleanUpAction(BaseAction):
```
jfuss: This name indicates we are cleaning up more than `.npmrc`. Can we scope the name down to be Npmrc, like how you have done for the Copy Action? Or was the goal to make one Npm Clean up action? 

Side thought: I wonder if this will be a pattern that develops (adding then removing the same set of files), to the point where it's worth codifying.

tests/unit/workflows/nodejs_npm/test_actions.py
```diff
@@ -78,3 +79,63 @@ def test_raises_action_failed_when_npm_fails(self, SubprocessNpmMock):
...
+                                       osutils=osutils)
+
+        with self.assertRaises(ActionFailedError):
+            action.execute()
+
+
+class TestNodejsCleanUpAction(TestCase):
+
+    @patch("aws_lambda_builders.workflows.nodejs_npm.utils.OSUtils")
+    def test_removes_npmrc_if_npmrc_exists(self, OSUtilMock):
```
jfuss: Can you also add a tests for when npmrc does not exist, to make sure that branch is also covered?

aws_lambda_builders/workflows/nodejs_npm/workflow.py
```diff
@@ -55,8 +55,13 @@ def __init__(self,
 
         npm_install = NodejsNpmInstallAction(artifacts_dir,
                                              subprocess_npm=subprocess_npm)
+
+        npm_copy_npmrc = NodejsNpmrcCopyAction(tar_package_dir, source_dir, osutils=osutils)
+
         self.actions = [
             npm_pack,
+            npm_copy_npmrc,
```
jfuss: Does the order matter at all? Does `npm_copy_npmrc` just need to be between `npm_pack` and 'npm_install` or must is happen before the copy source?

Trying to understand if there is some documentation we need to add here for the future.

gojko: it needs to come between npm_pack and npm_install because `npm pack` will exclude it from the package, but it's needed for `npm install` to correctly function (it may contain URLs to module repositories etc)

aws_lambda_builders/workflows/nodejs_npm/actions.py
gojko: there will be some other clean up actions to add later (such as `npm dedupe`). Would you prefer each step to be a separate action, or do we keep it all inside a single action?

on a related note, is it worth adding another `Purpose` value there? This isn't really `COPY_SOURCE`, but nothing else fit.

tests/unit/workflows/nodejs_npm/test_actions.py
simalexan: sure, I'll take a look at it

simalexan: done.

simalexan:
@jfuss I've added the test, @gojko gave good answers on the other two. If you think we should change something else, please do mention, would be happy to fix/add

simalexan:
@TheSriram Thank you! :) Waiting for approval of both of Jacob and you.

simalexan:
Thank you both @jfuss and @TheSriram !

simalexan:
@TheSriram thank you!

PR #54, opened 2018-12-18
Golang Dep Builder

sthulb:
*Issue #, if available:*

*Description of changes:*


aws_lambda_builders/actions.py
```diff
@@ -32,6 +32,9 @@ class Purpose(object):
     # Action is compiling source code
     COMPILE_SOURCE = "COMPILE_SOURCE"
 
+    # Action is removing source code
+    REMOVE_SOURCE = "REMOVE SOURCEs"
```
jfuss: Indention looks off here. It matches `COMPILE_SOURCE` but not `COPY_SOURCE`.

This should be
`REMOVE_SOURCE = "REMOVE_SOURCE` to match the schema of the exiting Purposes

aws_lambda_builders/actions.py
```diff
@@ -99,3 +102,19 @@ def __init__(self, source_dir, dest_dir, excludes=None):
 
     def execute(self):
         copytree(self.source_dir, self.dest_dir, ignore=shutil.ignore_patterns(*self.excludes))
+
+
+class FileCleanupAction(BaseAction):
```
jfuss: Match the schema of the other Actions here: `RemoveSourceAction`

aws_lambda_builders/actions.py
```diff
@@ -99,3 +102,19 @@ def __init__(self, source_dir, dest_dir, excludes=None):
...
+
+    NAME = "FileCleanup"
```
jfuss: `RemoveSource`

Not sure RemoveSource is the right name for everything in general.

aws_lambda_builders/utils.py
```diff
@@ -57,3 +57,23 @@ def copytree(source, destination, ignore=None):
             copytree(new_source, new_destination, ignore=ignore)
         else:
             shutil.copy2(new_source, new_destination)
+
+def removetree(source, pattern):
```
jfuss: Why is it important to to a regex here?

aws_lambda_builders/workflows/go_dep/workflow.py
```diff
@@ -0,0 +1,57 @@
...
+        tar_dest_dir = osutils.joinpath(scratch_dir, 'unpacked')
+        tar_package_dir = osutils.joinpath(tar_dest_dir, 'package')
+
+        handler = kwargs.get("options", {}).get("handler", None)
+
+        self.actions = [
+            CopySourceAction(tar_package_dir, artifacts_dir, excludes=self.EXCLUDED_FILES),
+            GoDepInstallAction(artifacts_dir, subprocess_dep),
+            GoBuildAction(artifacts_dir, handler, subprocess_go),
+            FileCleanupAction(artifacts_dir, r'/([\w]+\.go|Gopkg.toml)/gi')
```
jfuss: Why aren't we doing actions in the scratch directory and then doing a copy of the binary produced to the file resting place?

Doing regex's to clean up files isn't a process I think we should follow at all here. This is the reason we have a scratch directory.

.pylintrc
```diff
@@ -59,7 +59,7 @@ confidence=
 # --enable=similarities". If you want to run only the classes checker, but have
 # no Warning level messages displayed, use"--disable=all --enable=classes
 # --disable=W"
-disable=R0201,W0613,I0021,I0020,W1618,W1619,R0902,R0903,W0231,W0611,R0913,W0703,C0330,R0204,I0011,R0904,C0301
+disable=R0201,W0613,I0021,I0020,W1618,W1619,R0902,R0903,W0231,W0611,R0913,W0703,C0330,R0204,I0011,R0904,C0301,duplicate-code
```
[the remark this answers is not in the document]
sthulb: I’m just playing around for now to ensure things work. 

I wouldn’t worry about reviewing this too much for now.

aws_lambda_builders/workflows/go_dep/actions.py
```diff
@@ -0,0 +1,66 @@
...
+        self.base_dir = base_dir
+        self.source_path = source_path
+        self.output_path = output_path
+
+        self.subprocess_go = subprocess_go
+        self.env = env if not env is None else {}
+
+    def execute(self):
+        env = self.env
+        env.update({"GOOS": "linux", "GOARCH": "amd64"})
```
jfuss: This edits the OS environment Variables, which we do not want to do. 

There is a couple ways around this:
1. `self.env` should be a copy of the dictionary passed in 
2. what you pass into env should be a copy of the environment variables.

Ideally, I think we want both but either really works.

aws_lambda_builders/workflows/go_dep/actions.py
```diff
@@ -0,0 +1,66 @@
...
+
+        try:
+            self.subprocess_go.run(["build", "-o", self.output_path, self.source_path],
```
jfuss: How is the binary this produces named? We had a previous conversation on slack about this.

aws_lambda_builders/workflows/go_dep/subproc_exec.py
```diff
@@ -0,0 +1,90 @@
...
+                               cwd=cwd,
+                               env=env)
+
+        out, err = p.communicate()
+
+        if p.returncode != 0:
+            if sys.version_info[0] < 3:
+                raise ExecutionError(message=err.strip())
+            else:
+                raise ExecutionError(message=err.decode("utf8").strip())
```
jfuss: We should be able to just get away with `err.decode("utf8").strip()` for both Python2 and Python3. This matches what we are doing for the return statement on this function. Is there a reason we can't do this?

aws_lambda_builders/workflows/go_dep/utils.py
```diff
@@ -0,0 +1,42 @@
...
+        p = subprocess.Popen(command, stdout=stdout, stderr=stderr, env=env, cwd=cwd)
+        return p
+
+    @property
+    def pipe(self):
+        return subprocess.PIPE
+
+    @property
+    def environ(self):
+        return os.environ
```
jfuss: This should be a copy of the environ so consumers don't modify the customers env vars.

aws_lambda_builders/workflows/go_dep/workflow.py
```diff
@@ -0,0 +1,63 @@
...
+            osutils = OSUtils()
+
+        # project base name, where the Gopkg.toml and vendor dir are.
+        base_dir = osutils.abspath(osutils.dirname(manifest_path))
+        output_path = osutils.joinpath(osutils.abspath(artifacts_dir), handler)
+
+        subprocess_dep = SubprocessExec(osutils, "dep")
+        subprocess_go = SubprocessExec(osutils, "go")
+
+        self.actions = [
```
jfuss: We don't need to copy the source anywhere for this to work?

designs/Path_Resolver_DESIGN.md
```diff
@@ -0,0 +1,118 @@
+## Path Resolver for Builders
```
jfuss: Why is this being added in you PR?

aws_lambda_builders/workflows/go_dep/workflow.py
```diff
@@ -0,0 +1,63 @@
...
+
+        options = kwargs["options"] if "options" in kwargs else {}
+        handler = options.get("handler", None)
+
+        if osutils is None:
+            osutils = OSUtils()
+
+        # project base name, where the Gopkg.toml and vendor dir are.
+        base_dir = osutils.abspath(osutils.dirname(manifest_path))
+        output_path = osutils.joinpath(osutils.abspath(artifacts_dir), handler)
```
jfuss: Does this mean the binary is going to be under `.aws-sam/build/<logical_id>/<handler name>/`?

tests/integration/workflows/go_dep/data/src/remote-deps/Gopkg.toml
```diff
@@ -0,0 +1,34 @@
...
+#   source = "github.com/myfork/project2"
+#
+# [[override]]
+#   name = "github.com/x/y"
+#   version = "2.4.0"
+#
+# [prune]
+#   non-go = false
+#   go-tests = true
+#   unused-packages = true
```
jfuss: Are all these commented out lines needed?

tests/integration/workflows/go_dep/test_go_dep.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+class TestGoDep(TestCase):
+    TEST_DATA_FOLDER = os.path.join(os.path.dirname(__file__), "data")
+
+    def setUp(self):
+        self.artifacts_dir = tempfile.mkdtemp()
+        self.scratch_dir = tempfile.mkdtemp()
+
+        os.environ["GOPATH"] = self.TEST_DATA_FOLDER
```
jfuss: Don't edit the system's `os.environ`. Use a copy

Or at least unset the value in teardown if you need to edit the system.

aws_lambda_builders/workflows/go_dep/workflow.py
sthulb: Amazingly, no. we don't change anything for the source, other than ensuring we're have the deps, which the user will probably want anyway.

The version where we copy source was rather complex for the sake of it.

aws_lambda_builders/workflows/go_dep/workflow.py
sthulb: It'll be under whatever we specify as the artifact path in SAM CLI, which is likely to be the above (which is fine since the `CodeUri` property will define the path)

tests/integration/workflows/go_dep/data/src/remote-deps/Gopkg.toml
sthulb: no, but it's the default contents.

aws_lambda_builders/workflows/go_dep/workflow.py
jfuss: The convention is to place the code under the logical_id (not a subfolder). 

Can we match that expectation here?

sthulb: Surely, it's for whatever calls the aws-lambda-builder to define this, not this projects.

There's no guarantee that SAM CLI will always be the sole thing using this project.

sthulb: From what I can see in other builds, they write directly to the artefacts dir, the same way I do.

The go builder does nothing differently, the `output_path` is just `artifacts_dir/handler`

jfuss: Adding a subfolder is breaking the expectation, this has nothing to do with SAM CLI. We should be placing the artifacts in that folder directly not creating a subfolder. As a caller, I would need to understand this. This will actually break SAM CLI because the CodeUri is updated to be what the `artifacts_dir` is and this builder is creating a subfolder to place the artifacts in.

jfuss: `output_path` is the path to the binary not to a folder. I got those mixed up here.

jfuss:
@sthulb Can you rebase so we can get the build passing (you fixed the appveyor ruby in a different pr)?

I will make a pass at the PR now as well.

aws_lambda_builders/workflows/go_dep/subproc_exec.py
```diff
@@ -0,0 +1,88 @@
...
+        p = self.osutils.popen(invoke_bin,
+                               stdout=self.osutils.pipe,
+                               stderr=self.osutils.pipe,
+                               cwd=cwd,
+                               env=env)
+
+        out, err = p.communicate()
+
+        if p.returncode != 0:
+            if sys.version_info[0] < 3:
```
jfuss: Can you refactor this check into a function? That way we can guard the `out.strip()` as well?

To make this slightly more readable, can you use `sys.version_info.major` instead?

Also add a comment on why this is needed. I know we did some back and forth on this and while we understand the reasoning, it will be a great thing to document explicitly.

Side Note: We really appreciate the effort and diving into what was causing this to be needed!

aws_lambda_builders/workflows/go_dep/subproc_exec.py
```diff
@@ -0,0 +1,88 @@
...
+                raise ExecutionError(message=str(err.strip()))
+            else:
+                raise ExecutionError(message=str(err.decode("utf8").strip()))
+
+        return out.strip()
```
jfuss: In Py3, out will be bytes which does not match the return type of this function and other workflows. If you wrap what you did for `err` into a function and call this here, it will match everything.

aws_lambda_builders/workflows/go_dep/subproc_exec.py
```diff
@@ -0,0 +1,88 @@
...
+                               cwd=cwd,
+                               env=env)
+
+        out, err = p.communicate()
+
+        if p.returncode != 0:
+            if sys.version_info[0] < 3:
+                raise ExecutionError(message=str(err.strip()))
+            else:
+                raise ExecutionError(message=str(err.decode("utf8").strip()))
```
jfuss: The `str()` call here is not necessary. The `.decode` will return a string [1]. The `str()` might not be necessary in the `if` statement as well but not 100% sure.

[1] https://docs.python.org/3/library/stdtypes.html#bytes.decode

sthulb:
I'm not quite sure why ruby tests are failing for my build... I've not touched ruby in anyway in this branch.

The good news is, the Go tests all passed.

jfuss:
@sthulb Did you rebase with PR #66?

sthulb:
> @sthulb Did you rebase with PR #66?

Yep! I can see the bundler pinning in appveyor.yaml

PR #55, opened 2018-12-18
feat: Path Resolver and Validator

TheSriram:
- BaseWorkflow provides super simple resolver and a validator.
- Each Workflow can bring its own path resolver and validator.
- Each Workflow also can list the binaries to be used in the actions.
- workflow's run method is decorated to check for validaton of the
  language executable path.

The function "which" at <file> was copied from https://github.com/python/cpython/blob/3.7/Lib/shutil.py 
SPDX-License-Identifier: Python-2.0 
Copyright 2019 by the Python Software Foundation

*Issue #, if available:*

*Description of changes:*


TheSriram:
Will refactor and add unit tests, and bring in language specific path resolvers and validators in the next few commits.

aws_lambda_builders/workflows/python_pip/compat.py
```diff
@@ -1,13 +1,21 @@
...
-def pip_import_string():
-    import pip
-    pip_major_version = pip.__version__.split('.')[0]
+
+def pip_import_string(python_exe):
+    os_utils = OSUtils()
+    cmd = [
+        python_exe,
+        "-c",
+        "import pip; assert pip.__version__.split('.')[0] == 9"
```
jfuss: Shouldn't this be <= 9? Otherwise we try to do `from pip._internal import main` for any pip version that isn't 9, which I don't think the intention is correct.

9 should be a string here as well

TheSriram: This is a great catch, thanks! yes will update it.

aws_lambda_builders/workflow.py
```diff
@@ -22,6 +24,22 @@
...
 
+# TODO: Move sanitize out to its own class.
+def sanitize(func):
+    """
+    sanitize the executable path of the runtime specified by validating it.
+    :param func: Workflow's run method is sanitized
+    """
+
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs):
```
jfuss: Should we pass in self here and make that an explicit argument instead of getting the first value out of args?

aws_lambda_builders/workflow.py
```diff
@@ -138,6 +156,19 @@ def is_supported(self):
 
         return True
 
+    def get_executable(self):
+        """
+        Non specialized path resolver that just returns the first executable for the runtime on the path.
+        """
+        return PathResolver(runtime=self.runtime).path
```
jfuss: Should we make the attribute `path` more explicit? Something like `exec_path`?

Should we force `get_executable` into a strong contract by returning an object that has specific required methods or attributes, instead of just a string?

aws_lambda_builders/workflows/python_pip/actions.py
```diff
@@ -12,12 +13,20 @@ class PythonPipBuildAction(BaseAction):
...
         self.scratch_dir = scratch_dir
         self.runtime = runtime
-        self.package_builder = PythonPipDependencyBuilder(runtime=runtime)
+        self.runtime_path = runtime_path
+        self.pip = SubprocessPip(osutils=OSUtils(), python_exe=runtime_path)
+        self.pip_runner = PipRunner(python_exe=runtime_path, pip=self.pip)
+        self.dependency_builder = DependencyBuilder(osutils=OSUtils(), pip_runner=self.pip_runner,
+                                                    runtime=runtime)
+
+        self.package_builder = PythonPipDependencyBuilder(osutils=OSUtils(),
```
jfuss: `OSUtils()` is created 3 times here? Do we need new instances in each (`self.pip`, `self.pip_runner`, and `self.package_builder`).

Have you thought about whether `SubprocessPip`, `PipRunner` should be passed into the action here?

aws_lambda_builders/workflows/python_pip/validator.py
```diff
@@ -0,0 +1,65 @@
...
+        self.runtime = runtime
+        self.runtime_path = runtime_path
+
+    def has_runtime(self):
+        """
+        Checks if the runtime is supported.
+        :param string runtime: Runtime to check
+        :return bool: True, if the runtime is supported.
+        """
+        return self.runtime in self.SUPPORTED_RUNTIMES
```
jfuss: Sets are better for this comparison: https://stackoverflow.com/questions/2831212/python-sets-vs-lists

aws_lambda_builders/workflows/ruby_bundler/validator.py
```diff
@@ -0,0 +1,61 @@
...
+import os
+import subprocess
+
+from aws_lambda_builders.exceptions import MisMatchRuntimeError
+
+LOG = logging.getLogger(__name__)
+
+
+class RubyRuntimeValidator(object):
+    SUPPORTED_RUNTIMES = [
```
jfuss: set?

TheSriram:
Yep node support coming too. trying to sort out some test issues on python_pip workflow.

aws_lambda_builders/workflow.py
TheSriram: Sure we can do that.

aws_lambda_builders/workflow.py
TheSriram: what are other attributes did you have in mind?

aws_lambda_builders/workflows/python_pip/actions.py
TheSriram: Yeah, this place is a mess right now and needs clean up. but yes we need the pips to be intialized with the runtime_path, because we want the appropriate pip belonging to the chosen runtime.

aws_lambda_builders/workflow.py
jfuss: I didn't have any directly in mind but everywhere this method is doing some `PathResolver().path`. Instead, we can make `get_executable()` return the PathResolver and then when needed, we call the attribute or methods we need. This is what we are doing for the validation.

TheSriram:
Weird that appveyor is going through the python versions, and not finding the one that is in the Path.

.pylintrc
```diff
@@ -9,7 +9,7 @@
 
 # Add files or directories to the blacklist. They should be base names, not
 # paths.
-ignore=compat.py
+ignore=compat.py, utils.py, path_resolver.py, validator.py
```
jfuss: curious, why are these added?

.appveyor.yml
```diff
@@ -16,7 +16,8 @@ install:
 # To run Nodejs workflow integ tests
 - ps: Install-Product node 8.10
 
-- "set PATH=%PYTHON%\\Scripts;%PYTHON%\\bin;%PATH%"
+- "set PATH=%PYTHON%;%PYTHON%\\Scripts;%PYTHON%\\bin;%PATH%"
+- "echo %PATH%"
```
jfuss: When you rebase: I would move this echo down the script so we see everything in the PATH.

Makefile
```diff
@@ -3,8 +3,8 @@ init:
 
 test:
 	# Run unit tests
-	# Fail if coverage falls below 94%
-	LAMBDA_BUILDERS_DEV=1 pytest --cov aws_lambda_builders --cov-report term-missing --cov-fail-under 94 tests/unit tests/functional
+	# Fail if coverage falls below 93%
```
jfuss: Lowering the bar!? Any particular reason?

aws_lambda_builders/binary_path.py
```diff
@@ -0,0 +1,22 @@
...
+"""
+
+
+class BinaryPath(object):
+
+    def __init__(self, resolver, validator, binary, binary_path=None):
+        self.resolver = resolver
+        self.validator = validator
+        self.binary = binary
+        setattr(self, binary, binary)
```
jfuss: This is equivalent to the line above right?

aws_lambda_builders/workflow.py
```diff
@@ -22,6 +25,40 @@
...
+                try:
+                    valid_path = validator.validate(executable_path)
+                except MisMatchRuntimeError as ex:
+                    LOG.debug("Invalid executable for %s at %s",
+                              binary_path.binary, executable_path, exc_info=str(ex))
+                if valid_path:
+                    binary_path.binary_path = valid_path
+                    valid_paths.append(valid_path)
+                    break
+        if not valid_paths:
```
jfuss: if `len(self.binaries())` is 2 but only 1 is a valid path, we will still execute the workflow, which will fail.

aws_lambda_builders/workflow.py
```diff
@@ -126,6 +163,7 @@ def __init__(self,
 
         # Actions are registered by the subclasses as they seem fit
         self.actions = []
+        self.binaries = None
```
jfuss: Should this be an empty list instead of `None`?

aws_lambda_builders/workflow.py
```diff
@@ -22,6 +25,40 @@
...
+    """
+    sanitize the executable path of the runtime specified by validating it.
+    :param func: Workflow's run method is sanitized
+    """
+
+    @functools.wraps(func)
+    def wrapper(self, *args, **kwargs):
+        valid_paths = []
+        # NOTE: we need to access binaries to get paths and resolvers, before validating.
+        self.get_binaries()
```
jfuss: We call `self.get_binaries()` but don't use the output at all?

aws_lambda_builders/workflow.py
```diff
@@ -22,6 +25,40 @@
...
+        for binary_path in self.binaries:
```
jfuss: `self.binaries` can be `None` which will fail here. Probably better have a sane default here.

aws_lambda_builders/workflows/nodejs_npm/workflow.py
```diff
@@ -60,3 +60,9 @@ def __init__(self,
             CopySourceAction(tar_package_dir, artifacts_dir, excludes=self.EXCLUDED_FILES),
             npm_install,
         ]
+
+    def get_resolvers(self):
+        """
+        specialized path resolver that just returns the list of executable for the runtime on the path.
+        """
+        return [PathResolver(runtime=self.runtime, binary="node")]
```
jfuss: Thought node needed `npm`?

aws_lambda_builders/workflows/python_pip/actions.py
```diff
@@ -3,25 +3,37 @@
...
         self.artifacts_dir = artifacts_dir
         self.manifest_path = manifest_path
         self.scratch_dir = scratch_dir
         self.runtime = runtime
-        self.package_builder = PythonPipDependencyBuilder(runtime=runtime)
+        self.binaries = binaries
 
     def execute(self):
+        os_utils = OSUtils()
+        python_path = [binary.binary_path for binary in self.binaries if getattr(binary, self.LANGUAGE)][0]
```
jfuss: Not sure I understand this line of code. Could you explain this? My brain can't grasp this right now

aws_lambda_builders/workflows/python_pip/workflow.py
```diff
@@ -65,6 +67,22 @@ def __init__(self,
 
         self.actions = [
             PythonPipBuildAction(artifacts_dir, scratch_dir,
-                                 manifest_path, runtime),
+                                 manifest_path, runtime, binaries=self.get_binaries()),
             CopySourceAction(source_dir, artifacts_dir, excludes=self.EXCLUDED_FILES),
         ]
+
+    # def get_resolvers(self):
```
jfuss: Needed?

aws_lambda_builders/workflows/python_pip/workflow.py
```diff
@@ -65,6 +67,22 @@ def __init__(self,
...
+    #     """
+    #     specialized path resolver that just returns the list of executable for the runtime on the path.
+    #     """
+    #     return [PathResolver(runtime=self.runtime, binary=self.CAPABILITY.language)]
+
+    def get_validators(self):
+        return [PythonRuntimeValidator(runtime=self.runtime)]
+
+    # def get_binaries(self):
```
jfuss: Needed?

aws_lambda_builders/workflows/ruby_bundler/validator.py
```diff
@@ -0,0 +1,62 @@
...
+import logging
+import os
+import subprocess
+
+from aws_lambda_builders.exceptions import MisMatchRuntimeError
+
+LOG = logging.getLogger(__name__)
+
+
+class RubyRuntimeValidator(object):
```
jfuss: What's the purpose of this Validator? The ruby workflow uses bundler not ruby. Shouldn't we be validating the binaries the workflow uses?

TheSriram: Yeah agreed on this, we need a best attempt bundler validator instead.

aws_lambda_builders/workflows/python_pip/workflow.py
TheSriram: Can be removed.

aws_lambda_builders/workflows/python_pip/workflow.py
TheSriram: can be removed.

aws_lambda_builders/workflows/python_pip/actions.py
TheSriram: yes.
if binary_path.python is set, we get the associated binary_path, this is why I do setattr(self, binary, binary) above in the class.

so that binary_path.python is set.

aws_lambda_builders/workflows/nodejs_npm/workflow.py
TheSriram: by default, we look for the language in the workflow.. the workflow's language is filled as nodejs, but the executable is called node.

aws_lambda_builders/binary_path.py
TheSriram: No.

```
In [1]: class A(object):
   ...:     def __init__(self, a):
   ...:         self.a = a
   ...:         setattr(self, a , a)
   ...:

In [2]: A(a="python")
Out[2]: <__main__.A at 0x109ebcd68>

In [3]: obj = A(a="python")

In [4]: obj.a
Out[4]: 'python'

In [5]: obj.python
Out[5]: 'python'
```

Makefile
TheSriram: I was hovering at 93.76% and didnt have test coverage at that point. and I wanted to see the tick haha, this will go back up.

.appveyor.yml
TheSriram: sounds good.

.pylintrc
TheSriram: I had to keep utils.py because of the copying of the which command, the other two can go away.

aws_lambda_builders/workflow.py
TheSriram: good catch, it sets self.binaries, so I just used that directly.

PR #63, opened 2018-12-22
Feature: npm prebuild and postbuild

stojanovic:
*Issue #, if available:*
This implements additional build steps required by: https://github.com/awslabs/aws-sam-cli/issues/858. 

*Description of changes:*
Adding `sam:prebuild` and `sam:postbuild` npm scripts for Node.js projects, that will be executed before and after sam build command.

### DIFF
--- a/aws_lambda_builders/workflows/nodejs_npm/DESIGN.md
+++ b/aws_lambda_builders/workflows/nodejs_npm/DESIGN.md
@@ -78,7 +78,11 @@ Lambda environment.
 The general algorithm for preparing a node package for use on AWS Lambda
 is as follows.
 
-#### Step 1: Prepare a clean copy of the project source files
+#### Step 1: Execute prebuild scripts
+
+For situations where precompilation is required, such as typescript, users can define `sam:prebuild` npm script which will be executed in the source directory before packaging. (Note: development dependencies may be available at this point if the developer installs them in the source folder)
+
+#### Step 2: Prepare a clean copy of the project source files
 
 Execute `npm pack` to perform project-specific packaging using the supplied
 `package.json` manifest, which will automatically exclude temporary files, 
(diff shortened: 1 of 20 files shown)


jfuss:
@stojanovic and @gojko This is my fault in what I communicated in the converstation we had on slack. I see the value we could provide in this but this allows for arbitrary scripts to be run, which is a security concern for us. Since these post and pre builds just wrap the workflow, I think it is better for us to think about the generic solution and support this through that mechanism. If instead they were deeper in the workflows, there would be a different converstation on how we support these types of interactions within the workflow themselves. 

Maybe there is a case for supporting the transpilers that are commonly used in a different workflow so we can support the typescript use-cases more natively?

By the looks of this, customers using typescript can still use `sam build` (for example) but need to do the transpiling outside of the build command explicitly. My previous understand was we were blocking this support because the conversation needed to be after some specific action in the workflow.

Really sorry for the time I wasted for you on this, but I don't think we can accept this PR as is with the security concern.

gojko:
@jfuss npm is a security nightmare on its own, so I don't think that preventing people from running custom scripts inside the build is making it any less secure, but I understand your point :) I guess this will need to wait for some generic prebuild/postbuild implementation in SAM

j:
@jfuss generic support for typescript would be okay for typescript only users, but if you're building a (for example) typescript + react app (where you may be using "tsc" compiler AND "webpack" for front-end compilation), then generic typescript won't work for those folks.

Reality is, the JS world, people do everything in different ways.  So there really needs to be more management in how people build in the js ecosystem.

The way this library gets node_modules and packs everything is fine, however in a lot of webpack builds, people don't want to include anything in node_modules because webpack pulls it all in to a single file (most cases).  But this can be fixed by moving everything into a devDependency.

I would have loved to see AWS Sam builders be generic in the first place, just how apex up does their building (and they support any language lambda does).

Issue #64, opened 2018-12-27
Create Golang builder for Go Modules


volkangurel:
**Description:**
It would be great to have support in the SAM CLI to for go builds with [go modules](https://github.com/golang/go/wiki/Modules). That would allow SAM projects using go with go modules to have a standardized build process.

sanathkr:
We have a PR merged. Closing

PR #65, opened 2018-12-28
Add the Go Modules builder

volkangurel:
*Issue #, if available:*
#64 

*Description of changes:*
Add a builder for Go that assumes the given go project uses Go Modules, and so
can be built using standard go tooling.


aws_lambda_builders/workflows/go_modules/builder.py
```diff
@@ -0,0 +1,67 @@
...
+
+        :type config: :class:`lambda_builders.utils.Config` or None
+        :param config: To be determined. This is an optional config object
+            we can extend at a later date to add more options to how modules is
+            called.
+        """
+        env = {}
+        env.update(self.osutils.environ)
+        env.update({"GOOS": "linux", "GOARCH": "amd64"})
+        cmd = ["go", "build", "-o", self.osutils.joinpath(artifacts_dir_path, executable_name), source_dir_path]
```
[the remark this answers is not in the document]
TheSriram: Should be within a week, if there are no other major re-working required on it.

aws_lambda_builders/actions.py
```diff
@@ -92,10 +92,16 @@ class CopySourceAction(BaseAction):
 
     PURPOSE = Purpose.COPY_SOURCE
 
-    def __init__(self, source_dir, dest_dir, excludes=None):
+    def __init__(self, source_dir, dest_dir, only=None, excludes=None):
         self.source_dir = source_dir
         self.dest_dir = dest_dir
+        self.only = only
```
jfuss: Can you describe what you the use-case for this is vs just using the excludes? Should we allow only and excludes to be passed in? If so, how do they play with each other?

aws_lambda_builders/workflows/go_modules/utils.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+class OSUtils(object):
+    """
+    Wrapper around file system functions, to make it easy to
+    unit test actions in memory
+    """
+    @property
+    def environ(self):
+        return os.environ
```
jfuss: We should copy this (`os.envrion.copy()`), otherwise we risk changing Env Vars on the machine.

aws_lambda_builders/workflows/go_modules/workflow.py
```diff
@@ -0,0 +1,45 @@
...
+            artifacts_dir,
+            scratch_dir,
+            manifest_path,
+            runtime=runtime,
+            **kwargs)
+
+        if osutils is None:
+            osutils = OSUtils()
+
+        executable_name = osutils.basename(source_dir)
```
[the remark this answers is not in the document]
jfuss: In the Go Dep workflow, the executable name is defined to match the handler specified in the function. This detail should remain the same here otherwise the name of the binary could differ from what is in the template.

For reference: https://github.com/awslabs/aws-lambda-builders/pull/54/files#diff-c11c930189e8062aced3a130ff5ff0efR48

aws_lambda_builders/workflows/go_modules/workflow.py
```diff
@@ -0,0 +1,45 @@
...
+        builder = GoModulesBuilder(osutils)
+        self.actions = [
+            GoModulesBuildAction(source_dir, artifacts_dir, executable_name, builder),
+            CopySourceAction(source_dir, artifacts_dir, only=[executable_name]),
```
jfuss: Is this needed? `go build` can take an output dir. Why not do that as well here? 

Reference: https://github.com/awslabs/aws-lambda-builders/pull/54/files#diff-c11c930189e8062aced3a130ff5ff0efR55

volkangurel: It's not, the first call is actually placing things in the right place, so I don't need to `CopySourceAction` call. Fixing.

aws_lambda_builders/workflows/go_modules/utils.py
volkangurel: Fixed

aws_lambda_builders/workflows/go_modules/workflow.py
volkangurel: Fixed.

aws_lambda_builders/workflows/go_modules/workflow.py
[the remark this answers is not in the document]
volkangurel: Makes sense, fixed!

aws_lambda_builders/workflows/go_modules/builder.py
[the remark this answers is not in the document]
volkangurel: I've made the changes necessary for working with path resolvers and validators.

aws_lambda_builders/workflows/go_modules/workflow.py
```diff
@@ -0,0 +1,55 @@
...
+
+        builder = GoModulesBuilder(osutils, runtime_path=self.get_executable())
+        self.actions = [
+            GoModulesBuildAction(source_dir, output_path, builder),
+        ]
+
+    def get_executable(self):
+        return GoPathResolver(runtime=self.runtime).exec_path
+
+    def get_validator(self):
```
TheSriram: This has slightly changed in #55 , with get_validators(), get_resolvers() and binaries. So I would say go ahead and take a look at that.

jfuss:
@volkangurel Are you submitting this on behalf of Coinbase? If so, can you confirm you have approval from the company?

aws_lambda_builders/workflows/go_modules/validator.py
```diff
@@ -0,0 +1,57 @@
...
+        expected_major_version = self.runtime.replace(self.language, "").split('.')[0]
+
+        p = subprocess.Popen([self.runtime_path, "version"],
+                             cwd=os.getcwd(),
+                             stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        out, _ = p.communicate()
+
+        mismatched = p.returncode != 0 \
+            or len(out.split()) < 3 \
+            or out.split()[2].replace(self.language, "").split('.')[0] != expected_major_version
```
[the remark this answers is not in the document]
volkangurel: Yes, great point, I'll add that to the validator.

aws_lambda_builders/workflows/go_modules/validator.py
```diff
@@ -0,0 +1,63 @@
...
+        expected_major_version = self.runtime.replace(self.language, "").split('.')[0]
+
+        p = subprocess.Popen([runtime_path, "version"],
+                             cwd=os.getcwd(),
+                             stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        out, _ = p.communicate()
+
+        mismatched = p.returncode != 0 \
+            or len(out.split()) < 3 \
+            or out.split()[2].decode().replace(self.language, "").split('.')[0] != expected_major_version
```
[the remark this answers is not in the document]
volkangurel: Yep!

aws_lambda_builders/workflows/go_modules/validator.py
```diff
@@ -0,0 +1,63 @@
...
+
+        expected_major_version = self.runtime.replace(self.language, "").split('.')[0]
+
+        p = subprocess.Popen([runtime_path, "version"],
+                             cwd=os.getcwd(),
+                             stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        out, _ = p.communicate()
+
+        mismatched = p.returncode != 0 \
+            or len(out.split()) < 3 \
```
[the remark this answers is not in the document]
volkangurel: Where `out = "go version go1.10.3 darwin/amd64"`, the `len(out.split()) < 3` expression will return `False`, but that's expected since this is a chained or expression where any truthy part would set `mismatched` to `True` and raise an exception later. I can definitely make this clearer to avoid confusion.

requirements/base.txt
```diff
@@ -1 +1,2 @@
 six~=1.11
+whichcraft~=0.5.2
```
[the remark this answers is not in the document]
volkangurel: Yep!

volkangurel:
@jfuss I'm happy to submit it either on behalf of Coinbase or myself. Our company policy is:
"Beyond the constraints of the Employee Handbook, upstream Open Source contributions do not require explicit approval by Coinbase.", so I can confirm I have approval from the company if that's required.

aws_lambda_builders/workflows/go_modules/validator.py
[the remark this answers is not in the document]
jfuss: Ahh this was a guard for getting the version out of the `go version` command. I didn't realize this initially.

aws_lambda_builders/workflows/go_modules/validator.py
```diff
@@ -0,0 +1,68 @@
...
+                             cwd=os.getcwd(),
+                             stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        out, _ = p.communicate()
+
+        if p.returncode == 0:
+            out_parts = out.decode().split()
+            if len(out_parts) >= 3:
+                version_parts = [int(x) for x in out_parts[2].replace(self.LANGUAGE, "").split('.')]
+                if len(version_parts) == 3:
+                    if version_parts[0] == expected_major_version and version_parts[1] >= min_expected_minor_version:
```
jfuss: It would be nice to have sem version comparison somewhere. I am ok with this for now, I think we will see more of this and either should look at using a library or abstract this into a method any workflow could use.

aws_lambda_builders/workflows/go_modules/validator.py
```diff
@@ -0,0 +1,68 @@
...
+                        return self._valid_runtime_path
+
+        # otherwise, raise mismatch exception
+        raise MisMatchRuntimeError(language=self.LANGUAGE,
+                                   required_runtime=self.runtime,
+                                   runtime_path=runtime_path)
+
+    @property
+    def validated_runtime_path(self):
+        return self._valid_runtime_path if self._valid_runtime_path is not None else None
```
jfuss: `self._valid_runtime_path` defaults to None. This can be simplified to just return `self._valid_runtime_path`.

aws_lambda_builders/workflows/go_modules/workflow.py
```diff
@@ -0,0 +1,51 @@
...
+                 manifest_path,
+                 runtime=None,
+                 osutils=None,
+                 **kwargs):
+
+        super(GoModulesWorkflow, self).__init__(
+            source_dir,
+            artifacts_dir,
+            scratch_dir,
+            manifest_path,
```
jfuss: Just making sure I understand correctly. `go build` doesn't need to have `go.mod` file passed into the command. So that is the reason why this workflow doesn't need to use the manifest_path?

aws_lambda_builders/workflows/go_modules/workflow.py
```diff
@@ -0,0 +1,51 @@
...
+            runtime=runtime,
+            **kwargs)
+
+        if osutils is None:
+            osutils = OSUtils()
+
+        options = kwargs.get("options") or {}
```
jfuss: Not a huge deal but you can do `kwargs.get("options", {})`.

aws_lambda_builders/workflows/go_modules/validator.py
volkangurel: Good call! I'll fix this. I can also fix where I copied this from: https://github.com/awslabs/aws-lambda-builders/blob/f53dc86b9094bdae5fdc36d702b4d8c093c06ed7/aws_lambda_builders/workflows/python_pip/validator.py#L73.

aws_lambda_builders/workflows/go_modules/workflow.py
volkangurel: Right, `go build` walks up the directory tree to look for `go.mod`, no need to pass it in.

aws_lambda_builders/workflows/go_modules/workflow.py
volkangurel: I did it this way because if `options` is passed in as a kwarg set to `None`, then this would throw an exception. Using an `or` ensures we get a dict.

aws_lambda_builders/workflows/go_modules/validator.py
jfuss: Yes please. :)

aws_lambda_builders/workflows/go_modules/validator.py
volkangurel: I agree, I can build a small library function that can be shared among workflows and put it in https://github.com/awslabs/aws-lambda-builders/blob/develop/aws_lambda_builders/utils.py.

aws_lambda_builders/workflows/go_modules/workflow.py
jfuss: We need an explicit None check then. if `kwargs` is `None` then this will fail with an `AttributeError: 'NoneType' object has no attribute 'get'`. Can you add a None check here?

I did this in a 3.7 interpreter:
```
>>> kwargs = None
>>> kwargs.get("something") or {}
  File "<stdin>", line 1, in <module>
AttributeError: 'NoneType' object has no attribute 'get'
```

aws_lambda_builders/workflows/go_modules/validator.py
jfuss: I don't want to block the PR on this. We need to do this "provide utils easily to workflows" in general. To this point, we have copied common code across the workflows. This was to keep workflows completely independent but (as you can probably tell with OSUtils being everywhere), we need a better way to do it. I lump this into there. :)

aws_lambda_builders/workflows/go_modules/workflow.py
jfuss: Nevermind. That will not happen. What you are talking about is the value of `kwargs = {'options': None}` not `kwargs = None`. I get it now (apparently I need another coffee :))

volkangurel: Right, sorry my wording wasn't the best :)

jfuss:
@volkangurel Please confirm this contribution is under the terms of the Apache 2.0 license. Thanks.

PR #66, opened 2019-01-03
Updates bundler in AppVeyor

sthulb:
*Issue #, if available:*

*Description of changes:*
Pins bundled version on AppVeyor. 

AppVeyor drops out with:
```
ERROR:  Error installing bundler:
	The last version of bundler (>= 0) to support your Ruby & RubyGems was 1.17.3. Try installing it with `gem install bundler -v 1.17.3`
	bundler requires RubyGems version >= 3.0.0. The current RubyGems version is 2.7.8. Try 'gem update --system' to update RubyGems itself.
```

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -20,7 +20,8 @@ install:
 - "%PYTHON%\\python.exe -m pip install -r requirements/dev.txt"
 - "%PYTHON%\\python.exe -m pip install -e ."
 - "set PATH=C:\\Ruby25-x64\\bin;%PATH%"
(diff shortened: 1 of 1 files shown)


jfuss:
Should we just update RubyGems instead of pinning?

I don't understand ruby enough to understand which is better: @awood45?

sthulb:
I tried updating rubygems first. It added a considerable amount of time to the build.

jfuss:
@sthulb I see. I am ok pinning this. We don't have any min version requirement so this seems fine to pin given the build time.

awood45:
This approach is reasonable.

PR #67, opened 2019-01-04
Support for Rust Lambda function with Cargo

sapessi:
*Description of changes:* Added a new `rust_cargo` workflow. Uses `cargo build --release` by default to compile the function and automatically renames the executable to `bootstrap`. If the processing is running on a non-linux system the build process automatically attempts to target `x86_64-unknown-linux-musl`. The copy source steps looks for both the `bootstrap` executable and a `runtime_release` metadata file if it is available.


aws_lambda_builders/workflows/rust_cargo/workflow.py
```diff
@@ -0,0 +1,37 @@
+"""
+Python PIP Workflow
```
Phrohdoh: Copy + paste?

This is the Rust/Cargo workflow.

aws_lambda_builders/workflows/rust_cargo/cargo.py
```diff
@@ -0,0 +1,170 @@
+"""
+Cargo validator utility
+"""
+import os
+import logging
+
+import toml
```
Phrohdoh: From Travis:

> E   ModuleNotFoundError: No module named 'toml'

sapessi:
I'll work on unit tests over the next few days

jesskfullwood:
Has this PR stalled? If so I would be interested in taking on

sanathkr:
@jesskfullwood  Yes, it would be great if you can pick up this PR. It just some work to become compatible with latest interface of the Workflow/Builder classes. You can follow the other workflows as an example for the pattern. It should be straightforward.

sanathkr:
@jesskfullwood To make this work with SAM CLI, you need to make a code change to automatically pick up the Rust builder based on the presence of Cargo manifest as I explained here - https://github.com/awslabs/aws-sam-cli/issues/1099#issuecomment-488877333 

Let me know how we can help you with this

jackric:
@jesskfullwood @sanathkr  is this stalled? I want it in SAM; I may take it on

ramosbugs:
I've been wondering the same thing! Started working on reviving this PR last night, and hoping to have a new PR out early next week. But if someone beats me to it, great :-)

It sounds like the main missing piece is test coverage? Hopefully the Workflow/Builder interface changes alluded to above will become obvious as I continue trying to get this to work.

And then once this change gets merged we need a separate PR for `aws-sam-cli` to make it usable? That part seems to require primarily changing https://github.com/awslabs/aws-sam-cli/blob/98be672418f76bf1cc54db7528988b8a1c6a6b52/samcli/lib/build/workflow_config.py#L213 to
```
"provided": ManifestWorkflowSelector([PROVIDED_MAKE_CONFIG, RUST_CARGO_CONFIG])
```
@sanathkr Does that seem to match what you had in mind above?

sriram-mv:
there is now a makefile builder workflow that can be directly https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/building-custom-runtimes.html

ramosbugs:
great, that seems like a more flexible solution than special-casing Rust. I'll just use that, thanks!

PR #69, opened 2019-01-09
Gradle builder for Java

dagnir:
*Issue #, if available:*

*Description of changes:*
This adds the design document to enable `sam build` for Gradle based Java projects.


aws_lambda_builders/workflows/java_gradle/DESIGN.md
```diff
@@ -0,0 +1,117 @@
...
+We leverage Gradle to do all the heavy lifting for executing the `build.gradle`
+script which will resolve and download the dependencies and build the project.
+To create the distribution ZIP, we insert a new Gradle `task` that packages the
+compiled classes, resouces, and dependencies into a ZIP.
+
+### sam build
+
+#### Step 1: Check Java version and emit warning
+
+Check whether the local JDK version is <= Java 8, and if it is not, emit a
```
abrooksv: Should be `<`?

aws_lambda_builders/workflows/java_gradle/DESIGN.md
```diff
@@ -0,0 +1,117 @@
...
+properly configured (i.e. using `targetCompatibility`) or b) the project is
+build within the Lambda-compatibile container.
+
+#### Step 2: Synthesize a new build script
+
+There is no standard task in Gradle to create a distribution ZIP (or uber JAR).
+We must synthesize a new build script that we will run to actually produce the
+Lambda artifact. This is so we can insert a new task that will assemble the
+distribution ZIP. The new script is identical to the customer's original
+`build.gradle` except that it imports our own `sam-ext.gradle` file provided by
```
abrooksv: Did you get to look at the initScript system to see if we could leave their file alone? Otherwise we have to also worry about .kts files

dagnir: I did. I basically tried to adapt the given `sam-ext.gradle` as an init script but since we declare `dependsOn` on `build` and `runtimeClasspath`, it gives an error since they're not in scope until `build.gradle` is executed. Happy to discuss if you know a way around this.

+1 on finding `.kts` files for this approach.

aws_lambda_builders/workflows/java_gradle/DESIGN.md
```diff
@@ -0,0 +1,117 @@
...
+    into('lib') {
+        from configurations.runtimeClasspath
+    }
+}
+```
+
+#### Step 3: Build and package
+
+```sh
+gradle -b build.gradle.sam-build samBuildZip
```
abrooksv: Do we want to use gradlew? So that we don't have to worry about Gradle versions having to match.

aws_lambda_builders/workflows/java_gradle/DESIGN.md
dagnir: We should only warn if the version is > 8, since Lambda should be able to execute bytecode produced by 8 or lower.

aws_lambda_builders/workflows/java_gradle/DESIGN.md
abrooksv: You may just have to use the actual API instead of relying on things in scope. Like `tasks.findByName("build")`, but I admit I have not had a chance to look

aws_lambda_builders/workflows/java_gradle/DESIGN.md
dagnir: That would necessitate a preexisting `gradlew` file though right? Maybe we can look for that first and fallback to a local `gradle` executable if we can't find it.

abrooksv: It's standard practice to check the gradlew system into Git

abrooksv: But for last ditch we could search the path

dagnir: Right but they may not have generated one right (even if the majority of projects do)?

aws_lambda_builders/workflows/java_gradle/DESIGN.md
abrooksv: Hmmm not sure that is right either.... What if they do

```
    sourceCompatibility = JavaVersion.VERSION_1_8
    targetCompatibility = JavaVersion.VERSION_1_8
```

Then they can use JDK 12 if they wanted

aws_lambda_builders/workflows/java_gradle/DESIGN.md
abrooksv: Ya, gradlew should have precedence and fall back to search if you need to.

aws_lambda_builders/workflows/java_gradle/DESIGN.md
dagnir: Yeah, it's not a very reliable check, hence just a warning that it *may* not work. This situation is called out in the challenges section.

I would vote for emitting a warning and ensuring that's it calls out the need to set target and source compat if they're using > 8. Otherwise the warning produced by the JDK when an incompatible bytecode version is much more confusing IMO. Something like "Unsupported class file major version"

abrooksv: Alright, its also not a 1 way door.

aws_lambda_builders/workflows/java_gradle/DESIGN.md
```diff
@@ -0,0 +1,117 @@
...
+## Challenges
+
+Java bytecode can only run on the same or newer version of the JVM for which
+it was compiled for. For example Java 8 bytecode can run a JVM that is at
+least version 8, but bytecode targetting Java 9 cannot run on a Java 8 VM.
+This is further complicated by the fact that a newer JDK can generate code to
+be run on an older VM if configured using the `targetCompatibility` and
+`sourceCompatibility` properties of the Java plugin. Therefore, it is not
+sufficient to check the version of the local JDK, nor is it possible to check
+the value set for `targetCompatibility` or `sourceCompatibility` since it can
```
TheSriram: with #55 , you can provide your own validator to check for this before the workflow begins to run.

dagnir: Can you elaborate on which part #55 can help with? `targetCompatibility` and `sourceCompatibility` would be set at build time within `build.gradle`.

TheSriram: Ah, I just meant it as a check on the JDK version that was found when building for the java8 runtime.

aws_lambda_builders/workflows/java_gradle/DESIGN.md
```diff
@@ -0,0 +1,117 @@
+# Java - Gradle Lambda Builder
```
abrooksv: Got a bigger hypothetical question.

Java build systems (gradle / maven) rely a lot on internal dependencies and sub-projects (sub-modules). How do we want to leverage that?

Example project layout:
```
gradlew
template.yaml
build.gradle
settings.gradle

lambda1/src/main/java
lambda1/build.gradle

lambda2/src/main/java
lambda2/build.gradle

common/src/main/java
common/build.gradle
```

Where template.yaml defines 2 lambda functions:
1. Lambda1 with codeUri: lambda1
2. Lambda1 with codeUri: Lambda2

settings.gradle defines 2 sub-projects:
```
rootProject.name = 'FooService'
include 'lambda1'
include 'lambda2'
```

Lambda1 has a compile dependency on `project(":common")

Should sam build actually search up for the root and then run `./gradlew :Lambda1:build :lambda1:package` to build Lamda1 (and same for Lambda2)

Or do we want to force people to have to go and make 2 full Gradle projects (which is not idiomatic) and foce people to do `common/gradelw build install && sam build` so that common code gets compiled. I would say this is bad, so we may need to tweak the workflow design to allow for this...

(In maven world, it can build just the sub-module's pom.xml)

TheSriram: As of this moment, sam build works by by looking at each lambda function/ resource  in isolation and building it.

Look at below code to see how sam build works.

https://github.com/awslabs/aws-sam-cli/blob/develop/samcli/lib/build/app_builder.py#L112

If there are common dependencies, the manifest that is passed in needs to reflect it. The actions in the workflow need to know which dependency to build from where.

dagnir: Hmm okay, the manifest being `build.gradle` in this case. I'm not sure that they *is* a way for it to reflect that, since as @abrooksv pointed out, only the root project `settings.gradle` declares that.

dagnir: Okay actually it may actually not be very difficult. We should be able to leverage the built-in [`getPath`](https://docs.gradle.org/current/javadoc/org/gradle/api/Project.html#getPath()) API

abrooksv: Gradle will handle building other projects that it needs based on the dependency graph (compile project(":common")), the question to SAM is:

How much is mounted in the container? 
1. Is it the entire project? or just the Lambda1 folder?
2. If 1 is everything, do we have a path to it somehow in the workflow?

dagnir: I think there's still some open questions here.

We can only rely on `getProjectDir` if we can run the init script, but that requires a `gradle`/`gradlew` executable. Since we give preference to bundled `gradlew`, we need to be able to find the root directory of the project to search for it.

Is there any way for SAM to give us the root of the project? i.e. if `template.yaml` must be in the root of the project, can SAM provide it to the build function?

dagnir: If it comes to it, we can emulate the root project resolution Gradle does I suppose https://github.com/gradle/gradle/blob/master/subprojects/core/src/test/groovy/org/gradle/initialization/layout/BuildLayoutFactoryTest.groovy

jfuss: Let's not get hung up on what SAM needs or what is mounted. SAM is a caller of this and we should define what is needing without knowing or caring about what SAM does. For example, a customer can use this feature without the docker-lambda build container without ever using SAM CLI at all. We need to think about this in context of what is provided to the [Workflow](https://github.com/awslabs/aws-lambda-builders/blob/develop/aws_lambda_builders/workflow.py#L76)

aws_lambda_builders/workflows/java_gradle/DESIGN.md
```diff
@@ -0,0 +1,117 @@
...
+uber JAR.
+
+Another solution is to create a distribution ZIP containing the customer's
+classes and resources and include their dependency JARs under a `lib` directory.
+This keeps the customers' classes and resources separate from their
+dependencies' to avoid any file collisions. However, this incurs some overhead
+as the ZIP must be unpacked before the code can run.
+
+To avoid the problem of colliding files, we will choose the second option and
+create distribution ZIP's by default. We can further provide the `--uber-jar`
```
jfuss: Let's try and focus from what the aws-lambda-builders scope is and not options added to the CLI. This seems specific to SAM CLI, and while SAM CLI uses this we need to design for any caller.

aws_lambda_builders/workflows/java_gradle/DESIGN.md
```diff
@@ -0,0 +1,117 @@
...
+option for customers to force the creation of an uber JAR if necessary. This
+distribution will only contain the libraries needed at runtime (i.e. the runtime
+classpath libraries).
+
+Finally, a `--use-container` option will be provided to address potential issues
```
jfuss: SAM CLI specific. Where this runs doesn't need to be called out in the doc.

aws_lambda_builders/workflows/java_gradle/DESIGN.md
```diff
@@ -0,0 +1,117 @@
...
+artifact may not run in Lambda.
+
+Gradle projects are configured using `build.gradle` build scripts. These are
+executable files authored in either Groovy or since 5.0, Kotlin, and using the
+Gradle DSL. This presents a similar problem to `setup.py` in the Python world in
+that arbitrary logic can be executed during build time that could affect both
+how the customer's artifact is built, and which dependencies are chosen.
+
+To solve the issue where the build environment could affect the `build.gradle`
+script, a `--use-container` option will be provided, to give the option of
```
jfuss: SAM CLI specific. We shouldn't need to call out where this runs in the doc.

aws_lambda_builders/workflows/java_gradle/DESIGN.md
```diff
@@ -0,0 +1,117 @@
...
+building in an environment that closely mimics the Lambda environment.
+
+## Implementation
+
+We leverage Gradle to do all the heavy lifting for executing the `build.gradle`
+script which will resolve and download the dependencies and build the project.
+To create the distribution ZIP, we insert a new Gradle `task` that packages the
+compiled classes, resouces, and dependencies into a ZIP.
```
jfuss: Will we need to unzip this or is there a way to produce unzipped artifacts?

aws_lambda_builders/workflows/java_gradle/DESIGN.md
```diff
@@ -0,0 +1,117 @@
...
+
+### sam build
```
jfuss: This shouldn't be SAM specific, instead how about Build Steps

aws_lambda_builders/workflows/java_gradle/DESIGN.md
jfuss: Not 100% sure this is the right thing todo. Editing the script seems wrong. Can we provide a plugin that customers should use and then expect it to be there? 

I don't know a ton about gradle but will do some more reading to see what other options we may have here.

aws_lambda_builders/workflows/java_gradle/DESIGN.md
jfuss: Not sure I understand this search the path piece here. Why do we need to search for anything? Why wouldn't the gradlew file be provided in the `source_dir`?

aws_lambda_builders/workflows/java_gradle/DESIGN.md
```diff
@@ -0,0 +1,117 @@
...
+#### Step 3: Build and package
+
+```sh
+gradle -b build.gradle.sam-build samBuildZip
+```
+
+The `samBuildZip` task depends on the `build` task so we just run the
+`samBuildZip` task.
+
+### sam build --use-container
```
jfuss: SAM detail remove.

aws_lambda_builders/workflows/java_gradle/DESIGN.md
```diff
@@ -0,0 +1,117 @@
...
+}
+```
+
+#### Step 3: Build and package
+
+```sh
+gradle -b build.gradle.sam-build samBuildZip
+```
+
+The `samBuildZip` task depends on the `build` task so we just run the
```
jfuss: Does this mean a customer must have a `build` target in their gradle file? Or is this the build that is provided by gradle?

Sorry I am not up on gradle lingo.

dagnir: `build` is a standard task yes.

aws_lambda_builders/workflows/java_gradle/DESIGN.md
dagnir: Will change. FYI I followed the example from https://github.com/awslabs/aws-lambda-builders/blob/develop/aws_lambda_builders/workflows/ruby_bundler/DESIGN.md#sam-build

jfuss: Ahh. We should change that. The (Pip,Python)[https://github.com/awslabs/aws-lambda-builders/blob/develop/aws_lambda_builders/workflows/python_pip/DESIGN.md], (NPM,Nodejs)[https://github.com/awslabs/aws-lambda-builders/blob/develop/aws_lambda_builders/workflows/nodejs_npm/DESIGN.md] both have minimal SAM CLI mentions. This is due to SAM CLI just being a caller. I am happy to leave examples that use SAM CLI to describe callers points of view but in this repo we should focus on the interface of the (JSONRPC protocol)[https://github.com/awslabs/aws-lambda-builders/blob/develop/aws_lambda_builders/__main__.py#L55]

aws_lambda_builders/workflows/java_gradle/DESIGN.md
dagnir: > Not 100% sure this is the right thing todo. Editing the script seems wrong. Can we provide a plugin that customers should use and then expect it to be there?

We'll be switching to an init script which will be a separate file per @abrooksv's suggestion.

> Not sure I understand this search the path piece here. Why do we need to search for anything? Why wouldn't the gradlew file be provided in the source_dir?

Per @abrooksv's example, in a multi-project build, you would be a some sub-directory if you're building a single lambda, whereas `gradlew` will be at the root.

```sh
my-gradle-project
├── common
│   └── build.gradle
├── lambda1
│   └── build.gradle
├── lambda2
│   └── build.gradle
├── build.gradle
├── gradlew
├── settings.gradle
└── template.yaml
```

Maybe "search" was the wrong word to use. I just meant that we assume that `gradle` is on the PATH.

aws_lambda_builders/workflows/java_gradle/DESIGN.md
dagnir: Not sure in which context you're referring to but I tested this approach and `sam local start-api` works fine with the ZIP and this is supported Lambda https://docs.aws.amazon.com/lambda/latest/dg/deployment-package-v2.html

jfuss: Zips do work but that means we have to unzip on every invoke of that functions. If we have unzipped artifacts, customers will not endure this cost.

dagnir: Is that `sam local` only? I believe in Lambda it's just on cold start.

aws_lambda_builders/workflows/java_gradle/DESIGN.md
dagnir: I'll go ahead and remove this bit for now.

aws_lambda_builders/workflows/java_gradle/DESIGN.md
dagnir: Removed.

aws_lambda_builders/workflows/java_gradle/DESIGN.md
```diff
@@ -47,71 +41,128 @@ Gradle DSL. This presents a similar problem to `setup.py` in the Python world in
...
+on whether the `options` provided to the build command contains a map called
+`artifact_mapping`.
+
+When this map is **not** present, the semantics of `source_dir` and
+`artifact_dir` are not changed.
+
+If this map is present, then `source_dir` is treated as the root directory of
+the parent project. Additionally, `artifact_dir` will not be the path under
+which the artifact will be copied to, but just the parent of the final
+directory. `artifact_mapping` will be a mapping from a source subdirectory under
```
abrooksv: If the `artifact_mapping` is provided, should we optimize the build by doing 

```
for each mapping in artifact_mapping:
    cmd =+ " $mapping.folderName:build"

$GRADLE_EXECUTABLE --init-script /$SCRATCH_DIR/lambda-build-init.gradle $cmd
```

This does lead to issues where people make the sub-project name not match the folder (why people would I do not know...)
In that case we may have to make init script manipulare the task graph based on the project root of each sub-project. It's possible, just question around "should we"

aws_lambda_builders/workflows/java_gradle/DESIGN.md
```diff
@@ -47,71 +41,128 @@ Gradle DSL. This presents a similar problem to `setup.py` in the Python world in
...
 
 ## Implementation
 
-We leverage Gradle to do all the heavy lifting for executing the `build.gradle`
-script which will resolve and download the dependencies and build the project.
-To create the distribution ZIP, we insert a new Gradle `task` that packages the
-compiled classes, resouces, and dependencies into a ZIP.
+### Source directory, and artifact directory semantics
+
+To enable the multi build projects where a Gradle project can contain multiple
```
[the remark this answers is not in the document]
jfuss: This will be hidden. Using SAM CLI as an example:

A customer needing this could:
`sam build -m ./build.gradle`

SAM CLI will then conform to the details here. So we will provide the correct `source_dir` and `artifact_dir`. For the use-container, we will need to mount all of ProjectB (only way this will work, I think).

SAM CLI will only do this operation if -m is provided and the runtime is Java. Otherwise, SAM CLI will assume the Project A is what is desired (aka building the individual functions only).

aws_lambda_builders/workflows/java_gradle/DESIGN.md
jfuss: We can always optimize later if this is really needed.

aws_lambda_builders/workflows/java_gradle/DESIGN.md
```diff
@@ -0,0 +1,168 @@
...
+
+An interesting challenge is dealing with single build and multi build projects.
+Consider the following different projects structures:
+
+**Project A**
+```
+ProjectA
+├── build.gradle
+├── gradlew
+├── src
```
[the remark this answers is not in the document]
dagnir: Yes, the project is considered the function code; similarly for the multi-build example, `lambda1` and `lambda2` would each have a `src` subdirectory.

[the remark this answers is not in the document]
jfuss: Can you add a note that says Project A is the function? It was super clear in Project B but I was mapping that over to Project A which caused some confusion for me.

aws_lambda_builders/workflows/java_gradle/workflow.py
```diff
@@ -0,0 +1,72 @@
...
+        self.actions = [
+            JavaGradleBuildAction(source_dir,
+                                  artifacts_dir,
+                                  artifact_mapping,
+                                  subprocess_gradle,
+                                  scratch_dir,
+                                  self.os_utils)
+        ]
+
+    def _determine_gradle_exec(self):
```
jfuss: You should follow the validator and path resolver we just [introduced for this](https://github.com/awslabs/aws-lambda-builders/blob/develop/aws_lambda_builders/workflow.py#L29). Your workflow can define what binaries you need in a [dictionary](https://github.com/awslabs/aws-lambda-builders/blob/develop/aws_lambda_builders/workflow.py#L197). The sanitize method will then validate all binaries and set the first one that is validated. Since order matters here, you can use an OrderedDict to force the resolving of gradlew first.

aws_lambda_builders/workflows/java_gradle/workflow.py
```diff
@@ -0,0 +1,72 @@
...
+        """
+        Determine the correct Gradle executable to use. The project's included gradlew script takes precedence.
+
+        :return: The Gradle executable to use for the build.
+        """
+        gradlew_name = 'gradlew.bat' if self.os_utils.is_windows() else 'gradlew'
+        gradlew = os.path.join(self.source_dir, gradlew_name)
+        if os.path.exists(gradlew):
+            return gradlew
+        # assume Gradle is available on the PATH
```
jfuss: What happens if Gradle isn't in the PATH? If you use the validator/path resolver, the workflow will run if it is available otherwise fail. We want this to fail fast if the binaries are available.

aws_lambda_builders/workflows/java_gradle/workflow.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def _resolve_artifact_mapping(self):
+        """
+        Creates the map from the source of a single Lambda function (under source_dir), to the sub-directory under
+        artifacts_dir where the function's artifact is to be placed.
+
+        :return: The artifact mapping.
+        """
+        if self.options is None or self.options['artifact_mapping'] is None:
+            return {'.': '.'}
```
jfuss: If `artifact_mapping` is not provided, shouldn't the resting place for the artifact be in the `artifact_dir`? Does `{'.':'.'}` signify this?

aws_lambda_builders/workflows/java_gradle/workflow.py
```diff
@@ -0,0 +1,72 @@
...
+        return 'gradle'
+
+    def _resolve_artifact_mapping(self):
+        """
+        Creates the map from the source of a single Lambda function (under source_dir), to the sub-directory under
+        artifacts_dir where the function's artifact is to be placed.
+
+        :return: The artifact mapping.
+        """
+        if self.options is None or self.options['artifact_mapping'] is None:
```
jfuss: Can be simplified to if `self.options or self.options.get('artifact_mapping')`. This covers the case that `self.options = {}` otherwise `self.options['artifact_mapping']` throws a `KeyError`. We should have a test for this case as well.

aws_lambda_builders/workflows/java_gradle/actions.py
```diff
@@ -0,0 +1,59 @@
...
+        except Exception as ex:
+            raise ActionFailedError(str(ex))
+
+    def _build_project(self, init_script_file):
+        try:
+            self.subprocess_gradle.build(self.source_dir, init_script_file)
+        except GradleExecutionError as ex:
+            raise ActionFailedError(str(ex))
+
+    def _copy_artifacts(self):
```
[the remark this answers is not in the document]
dagnir: +1 will do.

aws_lambda_builders/workflows/java_gradle/workflow.py
dagnir: > Does `{'.':'.'}` signify this?

Yep the mappings are relative to `source_dir` and `artifacts_dir` respectively which is what we want for single build projects.

tests/integration/workflows/java_gradle/testdata/multi-build/with-deps-inter-module/common/build.gradle
```diff
@@ -0,0 +1,11 @@
+plugins {
+    id 'java'
+}
+
+repositories {
+    mavenCentral()
+}
+
+dependencies {
+    implementation 'software.amazon.awssdk:annotations:2.1.0'
```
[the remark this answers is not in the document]
dagnir: Done

aws_lambda_builders/workflows/java_gradle/resources/lambda-build-init.gradle
```diff
@@ -0,0 +1,74 @@
...
+import java.nio.file.SimpleFileVisitor
+import java.util.zip.ZipOutputStream
+import java.util.zip.ZipEntry
+
+gradle.taskGraph.afterTask{ t ->
+    if (t.name != 'build') {
+        return;
+    }
+
+    def mainSourcesDirs = t.project.sourceSets.main.output.files
```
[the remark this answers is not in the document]
dagnir: +1, done.

aws_lambda_builders/workflows/java_gradle/resources/lambda-build-init.gradle
```diff
@@ -33,12 +33,21 @@ def createArtifactDir(project) {
...
-def createLambdaArtifact(artifactPath, mainSourceDirs, classPathJars) {
+def createLambdaArtifact(artifactPath, artifactJars, classPathJars) {
     artifactPath.withOutputStream({ artifactOs ->
         def zipOs = new ZipOutputStream(artifactOs)
 
-        mainSourceDirs.each {
-            copyTree(zipOs, it.toPath());
+        artifactJars.each {
+            it.withInputStream({ jis ->
+                def ZipInputStream zipIs = new ZipInputStream(jis)
```
abrooksv: you shouldnt need the type after `def`

tests/integration/workflows/java_gradle/test_java_gradle.py
```diff
@@ -0,0 +1,146 @@
...
+
+        lambda2_expected_contents = ['aws/lambdabuilders/Lambda2_Main.class', 'lib/annotations-2.1.0.jar']
+        self.assert_zip_contains(lambda2_zip_path, lambda2_expected_contents)
+
+    def test_build_single_build_with_deps_broken(self):
+        source_dir = os.path.join(self.SINGLE_BUILD_TEST_DATA_DIR, 'with-deps-broken')
+        manifest_path = os.path.join(source_dir, 'build.gradle')
+        with self.assertRaises(WorkflowFailedError) as raised:
+            self.builder.build(source_dir, self.artifacts_dir, self.scratch_dir, manifest_path, runtime=self.runtime)
+        self.assertTrue(raised.exception.args[0].startswith('JavaGradleWorkflow:GradleBuild - Gradle Failed'))
```
abrooksv: Does the Gradle logs get shown back to the user?

TheSriram:
There is a PR out for searching additional paths when looking for executables that are to be used by the workflow. this will help for locating `gradlew` at different paths. #77 . #69 needs to integrate that.

dagnir:
@TheSriram Looks like it's #78? Is it mostly ready to go? Or should I wait for it to be merged?

aws_lambda_builders/workflows/java_gradle/gradle_resolver.py
```diff
@@ -0,0 +1,30 @@
...
+    def __init__(self, source_dir, os_utils, path_resolver=None):
+        self.source_dir = source_dir
+        self.os_utils = OSUtils() if not os_utils else os_utils
+        self.binary = 'gradle'
+        self.executables = [self.binary]
+        self.path_resolver = PathResolver(binary='gradle', runtime=None) if not path_resolver else path_resolver
+
+    @property
+    def exec_paths(self):
+        if self._includes_gradlew():
```
sanathkr: You can simplify this more by simply using `PathResolver('gradlew') or PathResolver('gradle')` to get paths for both :)

aws_lambda_builders/workflows/java_gradle/gradle_resolver.py
```diff
@@ -0,0 +1,30 @@
...
+            return [self._gradlew_path()]
+        return self.path_resolver.exec_paths
+
+    def _gradlew_path(self):
+        gradlew_name = 'gradlew.bat' if self.os_utils.is_windows() else 'gradlew'
```
sanathkr: Is it always 'gradlew.bat' for windows? or do they use .cmd or .exe extensions?

aws_lambda_builders/workflows/java_gradle/gradle_validator.py
```diff
@@ -0,0 +1,60 @@
...
+import re
+
+from aws_lambda_builders.exceptions import LambdaBuilderError
+from .utils import OSUtils
+
+LOG = logging.getLogger(__name__)
+
+
+class GradleBinaryValidatorError(LambdaBuilderError):
+    MESSAGE = "gradle executable found failed to return version"
```
sanathkr: Can you make this more descriptive? This error message is opaque

aws_lambda_builders/workflows/java_gradle/gradle_validator.py
```diff
@@ -0,0 +1,60 @@
...
+class GradleBinaryValidator(object):
+    MAJOR_VERSION_WARNING = "%s is using a JVM with major version %s which is newer than 8 that is supported by AWS " \
+                            "Lambda. The compiled function code may not run in AWS Lambda unless the project has " \
+                            "been configured to be compatible with Java 8 using 'targetCompatibility' in Gradle."
+
+    def __init__(self, os_utils=None, log=None):
+        self.language = 'java'
+        self._valid_binary_path = None
+        self.os_utils = OSUtils() if not os_utils else os_utils
+        self.log = LOG if not log else log
```
sanathkr: this is not a good pattern in Python. using LOG is the way to go.

aws_lambda_builders/workflows/java_gradle/gradle_validator.py
```diff
@@ -0,0 +1,60 @@
...
+        if version[0] == '1':
+            return version[1]
+        # Starting with Java 9, the major version is first
+        return version[0]
+
+    def _get_jvm_string(self, gradle_path):
+        p = self.os_utils.popen([gradle_path, '-version'], stdout=self.os_utils.pipe, stderr=self.os_utils.pipe)
+        stdout, _ = p.communicate()
+        if p.returncode != 0:
+            raise GradleBinaryValidatorError()
```
sanathkr: Since this validation is just a "warning", is it still worth hard failing here?

aws_lambda_builders/workflows/java_gradle/gradle_validator.py
```diff
@@ -0,0 +1,60 @@
...
+        for l in stdout.splitlines():
+            l_dec = l.decode()
+            if l_dec.startswith('JVM'):
```
sanathkr: How reliable is this parsing? ie. will a future gradle version change this?

tests/integration/workflows/java_gradle/test_java_gradle.py
```diff
@@ -0,0 +1,138 @@
...
+import tempfile
+
+from zipfile import ZipFile
+from unittest import TestCase
+
+from aws_lambda_builders.builder import LambdaBuilder
+from aws_lambda_builders.exceptions import WorkflowFailedError
+
+
+class TestJavaGradle(TestCase):
```
sanathkr: Love the extensive coverage of integration tests! Thanks a lot for taking the time. Really helps pushing the boundary for code quality

.appveyor.yml
```diff
@@ -36,6 +37,9 @@ install:
 - "go version"
 - "go env"
 
+# setup Gradle
+- "choco install gradle"
```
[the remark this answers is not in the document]
dagnir: This is for the test scenarios where `gradlew` isn't included in the project and we use the one on the PATH.

aws_lambda_builders/workflows/java_gradle/DESIGN.md
```diff
@@ -0,0 +1,184 @@
...
+Wrapper file within the root of the project. This essentially locks in the
+version of Gradle for the project and uses an executable that is independent of
+any local installations. This helps ensure that builds are always consistent
+over different environments.
+
+The `gradlew` script, if it is included, will be located at the root of the
+project. We will rely on the invoker of the workflow to supply the path to the
+`gradlew` script.
+
+We give precedence to this `gradlew` file, and if isn't found, we use the
```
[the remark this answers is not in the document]
dagnir: +1 will do.

aws_lambda_builders/workflows/java_gradle/workflow.py
```diff
@@ -0,0 +1,68 @@
...
+    def get_validators(self):
+        return [GradleBinaryValidator(self.os_utils)]
+
+    @property
+    def build_output_dir(self):
+        if self.build_dir is None:
+            self.build_dir = os.path.join(self.scratch_dir, self._compute_scratch_subdir())
+        return self.build_dir
+
+    def _compute_scratch_subdir(self):
```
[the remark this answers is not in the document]
dagnir: No, the default directory where Gradle stores compiled class files and the like is `$SRC_DIR/build` (the `buildDir`), so to avoid writing them in the source directory, we set it to somewhere in `scratch_dir`. I went into this a little in Step 4, but I'll clarify.

aws_lambda_builders/workflows/java_gradle/gradle_resolver.py
dagnir: I don't believe this will work right now for `gradlew` case though right? `gradlew` is included in the project root, so it won't be on the `PATH`.

aws_lambda_builders/workflows/java_gradle/gradle_resolver.py
dagnir: Yes, it's always `gradlew.bat`; `gradlew`/`gradlew.bat` are not binaries, they're simply scripts that download the correct binaries to run and then execute it with the provided params.

aws_lambda_builders/workflows/java_gradle/gradle_validator.py
dagnir: Will do!

aws_lambda_builders/workflows/java_gradle/gradle_validator.py
dagnir: This was simply so I could test that the validator was actually logging the warning (by passing in a mock log object). This is the pattern we use in the Java SDK. Is there a more idiomatic way to test this in Python?

aws_lambda_builders/workflows/java_gradle/gradle_validator.py
dagnir: Nope; will replace with a warning.

aws_lambda_builders/workflows/java_gradle/gradle_validator.py
dagnir: I went all the way back to gradle 1.0 and the output of `-v` seems to have remained the same so I think it should be pretty reliable

aws_lambda_builders/workflows/java_gradle/gradle_resolver.py
dagnir: Incorporated #78

aws_lambda_builders/workflows/java_gradle/DESIGN.md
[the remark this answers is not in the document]
dagnir: Using #78

aws_lambda_builders/workflows/java_gradle/DESIGN.md
```diff
@@ -164,21 +164,32 @@ $GRADLE_EXECUTABLE --project-cache-dir $SCRATCH_DIR/gradle-cache \
...
+ensures that we are not writing anything under the source directory.
+
+One important detail here is that the init script may create *multiple*
+subdirectories under `scratch_dir`, one for each project involved in building
+the lambda located at `source_dir`. Going back to the `ProjectB` example, if
+we're building `lambda1`, this also has the effect of building `common` because
+it's a declared dependency in its `build.gradle`. So, within `scratch_dir` will
+be a sub directory for each project that gets built as a result of building
+`source_dir`; in this case there will be one for each of `lambda1` and `common`.
+The init file uses some way of mapping the source root of each project involved
```
sanathkr: Thanks for the explanation. This makes sense.

aws_lambda_builders/workflows/java_gradle/gradlew_resolver.py
```diff
@@ -0,0 +1,29 @@
+"""
+Gradle executable resolution
+"""
+
+from aws_lambda_builders.path_resolver import PathResolver
+
+
+class GradlewResolver(object):
+    DUMMY_PATH = object()
```
sanathkr: Can't you just use None instead?

aws_lambda_builders/workflows/java_gradle/gradle_validator.py
sanathkr: Makes sense. Usually you don't test for log statements, but in this case it makes sense.

aws_lambda_builders/workflows/java_gradle/workflow.py
```diff
@@ -51,18 +54,30 @@ def __init__(self,
         ]
 
     def get_resolvers(self):
-        return [GradleResolver(self.source_dir, self.os_utils)]
+        gradlew_resolver = GradlewResolver(binary=self._wrapper_name(),
```
sanathkr: hang on, why are you using internal `_wrapper_name`? I don't quite understand the interaction between this and GradlewResolver

aws_lambda_builders/workflows/java_gradle/gradlew_resolver.py
```diff
@@ -0,0 +1,29 @@
...
+        else:
+            self.path_resolver = path_resolver
+
+    @property
+    def exec_paths(self):
+        try:
+            return self.path_resolver.exec_paths
+        except ValueError as e:
+            # gradlew is optional so we're okay with not finding the executable
+            if str(e).startswith("Path resolution for runtime"):
```
sanathkr: oh no! If you were to do hacks at this level, we shoudl either fix the Gradle implementation design or the Path resolver design.

aws_lambda_builders/workflows/java_gradle/resources/lambda-build-init.gradle
```diff
@@ -0,0 +1,103 @@
...
+        return;
+    }
+
+    assertExpectedBuildDir(t.project)
+
+    def artifactJars = t.project.configurations.archives.artifacts.files.files
+
+    def runtimeCpJars = t.project.configurations.runtimeClasspath.files
+
+    def zipArtifact = createArtifactDir(t.project).resolve("${t.project.name}.zip")
```
sanathkr: Actually, do you really need to create a zip here? Can you instead write everything into a directory that SAM CLI will eventually zip everything up using `sam package`? 

In terms of workflow, `sam build` -> `sam local invoke` -> `sam package` -> `sam deploy`, writing a zip here means the `local invoke` will unzip and rezip eventually during `package`.

aws_lambda_builders/workflows/java_gradle/workflow.py
```diff
@@ -0,0 +1,83 @@
...
+                            dependency_manager="gradle",
+                            application_framework=None)
+
+    INIT_FILE = "lambda-build-init.gradle"
+
+    def __init__(self,
+                 source_dir,
+                 artifacts_dir,
+                 scratch_dir,
+                 manifest_path,
```
sanathkr: You should also verify that this manifest file exists and fail the workflow if it does not. We don't validate this in SAM CLI side because the builder is the authority on manifests.

sanathkr: Similar to this https://github.com/awslabs/aws-lambda-builders/blob/develop/aws_lambda_builders/workflows/python_pip/packager.py#L37-L40

aws_lambda_builders/workflows/java_gradle/gradlew_resolver.py
dagnir: Providing `gradlew` is optional, but `PathResolver` throws a `ValueError` if it's not found.

aws_lambda_builders/workflows/java_gradle/resources/lambda-build-init.gradle
dagnir: Hmm but where do we draw the line between SAM and the builders? If customers want to just use the builders library, it would be more convenient if they didn't need to ZIP if themselves

aws_lambda_builders/workflows/java_gradle/workflow.py
dagnir: The resolver needs to know whether it's looking for `gradlew.bat` or `gradlew` based on the OS.

aws_lambda_builders/workflows/java_gradle/gradlew_resolver.py
dagnir: I tried just returning `None` but I get a workflow error when running from here: https://github.com/awslabs/aws-lambda-builders/blob/681d65b76c925980a158fd767cc5716488df0474/aws_lambda_builders/workflow.py#L56-L58

aws_lambda_builders/workflows/java_gradle/workflow.py
dagnir: +1

aws_lambda_builders/workflows/java_gradle/gradlew_resolver.py
dagnir: Thinking about it more, it will be cleaner to just create a custom path resolver for `gradlew` + `gradle`

aws_lambda_builders/workflows/java_gradle/resources/lambda-build-init.gradle
sanathkr: The scope of builder is to "compile" the code into deployable artifacts. How you package the deployable artifacts is the responsibility of a packager (which currently is in SAM CLI). It is zip if you want to deploy to Lambda, but we could also package artifacts for say publishing to Serverless Apps Repo which would be a different format. We might package each function separately or all functions together, but those are decisions the packager should make.

dagnir: Makes sense. Do I just treat `artifacts_dir` as if it were the root of the ZIP? e.g.

```
artifact_dir/com/amazonaws/Foo.class
artifact_dir/com.amazonaws/Bar.class
lib/dependency1.jar
```
?

aws_lambda_builders/workflows/java_gradle/actions.py
```diff
@@ -72,7 +76,13 @@ def execute(self):
     def _copy_artifacts(self):
         lambda_build_output = os.path.join(self.build_dir, 'build', 'distributions', 'lambda-build')
         try:
-            for f in self.os_utils.listdir(lambda_build_output):
-                self.os_utils.copy(os.path.join(lambda_build_output, f), self.artifacts_dir)
+            self.os_utils.makedirs(self.artifacts_dir)
```
sanathkr: os.scandir is not available in Python 2.7. Can you use this osutils helper instead?

https://github.com/awslabs/aws-lambda-builders/blob/develop/aws_lambda_builders/workflows/python_pip/utils.py#L56-L66

Copy/paste it in your code for now.

tests/integration/workflows/java_gradle/test_java_gradle.py
```diff
@@ -0,0 +1,119 @@
...
+    def test_build_single_build_with_test_deps_test_jars_not_included(self):
+        source_dir = os.path.join(self.SINGLE_BUILD_TEST_DATA_DIR, 'with-test-deps')
+        manifest_path = os.path.join(source_dir, 'build.gradle')
+        self.builder.build(source_dir, self.artifacts_dir, self.scratch_dir, manifest_path, runtime=self.runtime)
+        expected_files = [p('aws', 'lambdabuilders', 'Main.class'), p('lib', 'annotations-2.1.0.jar')]
+
+        self.assert_artifact_contains_files(expected_files)
+        self.assert_artifact_not_contains_file(p('lib', 's3-2.1.0.jar'))
+
+    def test_build_single_build_with_deps_gradlew(self):
```
TheSriram: Nice! executable search paths :)

PR #72, opened 2019-01-22
Fix local package installation

stealthycoin:
When packaging the dependencies only the first local package would be
included in the final bundle. This change adds support for any number
of local directory links to be treated as buildable dependencies. Tests
were also added for the case of 1 or 2 local directory links specified
in the requirements.txt.


sanathkr:
@TheSriram can you review?

Issue #77, opened 2019-02-08
Add support for specific executable lookups for workflows


TheSriram:
It would be great to directly specify the path of the executable required for executing the workflow, instead of relying on looking up the the executable in precedence order in $PATH

TheSriram:
This issue has been addressed by #78

PR #78, opened 2019-02-08
feat: allow path resolver to look at additional paths

TheSriram:
- search for executables on additional paths that are not on `$PATH`
- additional paths take precedence before searching for executables
  on `$PATH`
- rev protocol version to 0.2 with introduction of additional search
  paths.

*Issue #, if available:*

https://github.com/awslabs/aws-lambda-builders/issues/77

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -23,7 +23,7 @@ install:
 - "%PYTHON%\\python.exe -m pip install -e ."
 - "set PATH=C:\\Ruby25-x64\\bin;%PATH%"
 - "gem --version"
-- "gem install bundler -v 1.17.3 --no-ri --no-rdoc"
+- "gem install bundler -v 1.17.3"
 - "bundler --version"
 - "echo %PATH%"
 
--- a/aws_lambda_builders/__init__.py
+++ b/aws_lambda_builders/__init__.py
@@ -2,4 +2,4 @@
 AWS Lambda Builder Library
 """
 __version__ = '0.0.5'
-RPC_PROTOCOL_VERSION = "0.2"
+RPC_PROTOCOL_VERSION = "0.1"
--- a/aws_lambda_builders/__init__.py
+++ b/aws_lambda_builders/__init__.py
@@ -2,4 +2,4 @@
 AWS Lambda Builder Library
 """
 __version__ = '0.0.5'
-RPC_PROTOCOL_VERSION = "0.1"
+RPC_PROTOCOL_VERSION = "0.2"
--- a/aws_lambda_builders/__main__.py
+++ b/aws_lambda_builders/__main__.py
@@ -54,7 +54,9 @@ def _parse_version(version_string):
     if VERSION_REGEX.match(version_string):
         return float(version_string)
     else:
-        raise ValueError("Protocol Version does not match : {}".format(VERSION_REGEX.pattern))
+        ex = "Protocol Version does not match : {}".format(VERSION_REGEX.pattern)
+        LOG.debug(ex)
+        raise ValueError(ex)
 
 
 def version_compatibility_check(version):
--- a/aws_lambda_builders/__main__.py
+++ b/aws_lambda_builders/__main__.py
@@ -10,10 +10,11 @@ import sys
 import json
 import os
 import logging
+import re
 
 from aws_lambda_builders.builder import LambdaBuilder
 from aws_lambda_builders.exceptions import WorkflowNotFoundError, WorkflowUnknownError, WorkflowFailedError
-
+from aws_lambda_builders import RPC_PROTOCOL_VERSION as lambda_builders_protocol_version
 
 log_level = int(os.environ.get("LAMBDA_BUILDERS_LOG_LEVEL", logging.INFO))
 
@@ -23,6 +24,8 @@ logging.basicConfig(stream=sys.stderr,
                     format='%(message)s')
 
 LOG = logging.getLogger(__name__)
+
+VERSION_REGEX = re.compile("^([0-9])+.([0-9]+)$")
 
 
 def _success_response(request_id, artifacts_dir):
@@ -44,6 +47,22 @@ def _error_response(request_id, http_status_code, message):
             "message": message
         }
     })
+
+
+def _parse_version(version_string):
+
+    if VERSION_REGEX.match(version_string):
+        return float(version_string)
+    else:
+        raise ValueError("Protocol Version does not match : {}".format(VERSION_REGEX.pattern))
+
+
+def version_compatibility_check(version):
+    if _parse_version(lambda_builders_protocol_version) < version:
+        ex = "Incompatible Protocol Version : {}, " \
+             "Current Protocol Version: {}".format(version, lambda_builders_protocol_version)
+        LOG.error(ex)
+        raise ValueError(ex)
 
 
 def _write_response(response, exit_code):
@@ -77,11 +96,20 @@ def main():  # pylint: disable=too-many-statements
         response = _error_response(request_id, -32601, "Method unavailable")
         return _write_response(response, 1)
 
+    try:
+        protocol_version = _parse_version(params.get("__protocol_version"))
+        version_compatibility_check(protocol_version)
+
+    except ValueError:
+        response = _error_response(request_id, 505, "Unsupported Protocol Version")
+        return _write_response(response, 1)
+
     capabilities = params["capability"]
     supported_workflows = params.get("supported_workflows")
 
     exit_code = 0
     response = None
+
     try:
         builder = LambdaBuilder(language=capabilities["language"],
                                 dependency_manager=capabilities["dependency_manager"],
@@ -93,6 +121,7 @@ def main():  # pylint: disable=too-many-statements
                       params["artifacts_dir"],
                       params["scratch_dir"],
                       params["manifest_path"],
+                      executable_search_paths=params['executable_search_paths'],
                       runtime=params["runtime"],
                       optimizations=params["optimizations"],
                       options=params["options"])
--- a/aws_lambda_builders/__main__.py
+++ b/aws_lambda_builders/__main__.py
@@ -60,6 +60,13 @@ def _parse_version(version_string):
 
 
 def version_compatibility_check(version):
+    # The following check is between current protocol version vs version of the protocol
+    # with which aws-lambda-builders is called.
+    # Example:
+    # 0.2 < 0.2 comparison will fail, don't throw a value Error saying incompatible version.
+    # 0.2 < 0.3 comparison will pass, throwing a ValueError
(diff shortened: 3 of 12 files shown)


jfuss:
Are there any integration tests we can add to make sure we don't break this behavior in the future?

aws_lambda_builders/__init__.py
```diff
@@ -2,4 +2,4 @@
 AWS Lambda Builder Library
 """
 __version__ = '0.0.5'
-RPC_PROTOCOL_VERSION = "0.1"
+RPC_PROTOCOL_VERSION = "0.2"
```
jfuss: I don't think we want to bump this right now. Since we release aws-lambda-builders and build it into a docker image before we release SAM CLI, this breaks all customers that are not using `--skip-pull-image` for all runtimes `sam build` supports. SAM CLI does not have a mechanism to pin to a docker image, so customers are left hanging until we do a SAM CLI release. Until we have a good mechanism to release this package without impacting customers using SAM CLI or releasing together, we can't update this version.

TheSriram: will revise the version back to 0.1, we need a plan for updating the protocol version. I'll capture that as a github issue.

aws_lambda_builders/utils.py
```diff
@@ -68,13 +68,12 @@ def copytree(source, destination, ignore=None):
...
-    `mode` defaults to os.F_OK | os.X_OK. `path` defaults to the result
-    of os.environ.get("PATH"), or can be overridden with a custom search
-    path.
+def which(cmd, mode=os.F_OK | os.X_OK, additional_search_paths=None):  # pragma: no cover
+    """Given a command, mode, and an additional search paths list, return the paths which
+    conforms to the given mode on the PATH with the prepended additional search paths,
+    or None if there is no such file.
+    `mode` defaults to os.F_OK | os.X_OK. the default search `path` defaults
+    to the result of os.environ.get("PATH")
     Note: This function was backported from the Python 3 source code.
```
sanathkr: docstring for the new parameter

aws_lambda_builders/utils.py
```diff
@@ -93,10 +92,11 @@ def _access_check(fn, mode):
...
 
-    if path is None:
-        path = os.environ.get("PATH", os.defpath)
-    if not path:
-        return None
+    path = os.environ.get("PATH", os.defpath)
+
+    if additional_search_paths:
+        for pth in additional_search_paths:
+            path.insert(0, pth)
```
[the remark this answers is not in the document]
sanathkr: Also, it is confusing here because looks like `path` is a string and additional_search_paths is a list. I would first convert path to list before concatenating

tests/functional/test_cli.py
```diff
@@ -65,6 +65,7 @@ def test_run_hello_workflow(self, flavor):
                 "runtime": "ignored",
                 "optimizations": {},
                 "options": {},
+                "additional_paths": []
```
sanathkr: this is called `additional_search_paths` right?

sanathkr: Also, you should modify __main__.py to parse and pass the value to builder

aws_lambda_builders/__init__.py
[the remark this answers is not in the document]
sanathkr: Also, there is implementation to compares versions and returns the necessary error. You should implement it. It will be good to implement it within the __main__.py so only the CLI interface is affected

aws_lambda_builders/utils.py
```diff
@@ -68,13 +68,12 @@ def copytree(source, destination, ignore=None):
...
 
 
-def which(cmd, mode=os.F_OK | os.X_OK, path=None):  # pragma: no cover
-    """Given a command, mode, and a PATH string, return the paths which
-    conforms to the given mode on the PATH, or None if there is no such
-    file.
-    `mode` defaults to os.F_OK | os.X_OK. `path` defaults to the result
-    of os.environ.get("PATH"), or can be overridden with a custom search
-    path.
+def which(cmd, mode=os.F_OK | os.X_OK, additional_search_paths=None):  # pragma: no cover
```
sanathkr: do you know if anyone is using the `path` parameter?

aws_lambda_builders/__init__.py
sanathkr: Actually not. 

There are two versions at play here 
- Builder library version within SAM CLI. Think of this as a client
- Buildre library within the container. Think of this as the server

Client issues a `build` RPC command to server.

SAM CLI Builder (client) |  Builder Version in Container (Server) | Compatibility
---------|-----------------|--------------
0.1 | 0.2 | YES - New version of server is backwards compat with old version of client
0.2 | 0.1 | NO - New client is not compatible with old server

We auto-upgrade Docker containers. So if someone gets newer version of container aka server, SAM CLI will continue to work. But if someone upgrades SAM CLI but uses --skip-pull-image to stick with older container version, their existing builds will break. This is because newer client has behavior that older servers don't support.

The thought is, we will try to be back compat as much as possible. So checking if `client version <= server version` is good enough. At some point we will intentionally break compat when we will rev major version.

Does this make sense?

aws_lambda_builders/__main__.py
```diff
@@ -46,6 +49,22 @@ def _error_response(request_id, http_status_code, message):
     })
 
 
+def _parse_version(version_string):
+
+    if VERSION_REGEX.match(version_string):
+        return float(version_string)
```
sanathkr: didn't realize you could do this :) Hope it is safe

aws_lambda_builders/__main__.py
```diff
@@ -46,6 +49,22 @@ def _error_response(request_id, http_status_code, message):
...
+    else:
+        raise ValueError("Protocol Version does not match : {}".format(VERSION_REGEX.pattern))
```
sanathkr: Can you LOG.debug here? We should be able to debug this

aws_lambda_builders/__main__.py
```diff
@@ -46,6 +49,22 @@ def _error_response(request_id, http_status_code, message):
...
+
+
+def version_compatibility_check(version):
+    if _parse_version(lambda_builders_protocol_version) < version:
```
sanathkr: This should be `<=` - you want "0.2 == 0.2" to be a pass

TheSriram:
@sanathkr we are definitely revving the version. That was an old comment, disregard that.

aws_lambda_builders/__main__.py
TheSriram: should be safe, we do a regex match before we get here.

aws_lambda_builders/__main__.py
TheSriram: sure

aws_lambda_builders/__init__.py
jfuss: @sanathkr Thanks for the detailed explanation. I had a miss understanding of the SAM CLI interactions.

aws_lambda_builders/__main__.py
TheSriram: No this is a condition check when an error is thrown (Incompatible version), I have added func tests to show it works.

(current lamdda builders version < lambda builder version from client)

0.2 < 0.2 comparison will fail, therefore we dont throw a value Error saying incompatible version.
0.2 < 0.3 comparison will pass, throwing a ValueError
0.2 < 0.1 comparison will fail, therefore we dont throw a value Error saying incompatible version

tests/functional/test_cli.py
```diff
@@ -50,17 +50,19 @@ def tearDown(self):
         shutil.rmtree(self.artifacts_dir)
 
     @parameterized.expand([
-        ("request_through_stdin"),
-        ("request_through_argument")
+        ("request_through_stdin", lambda_builders_protocol_version),
```
sanathkr: can you add a test that fails compat check?

aws_lambda_builders/__main__.py
sanathkr: Ah okay, I didn't realize this was the case. Thanks for clarification. Can you add a quick comment explaining why this is the case

aws_lambda_builders/utils.py
```diff
@@ -68,15 +68,27 @@ def copytree(source, destination, ignore=None):
...
 
 
-def which(cmd, mode=os.F_OK | os.X_OK, path=None):  # pragma: no cover
-    """Given a command, mode, and a PATH string, return the paths which
-    conforms to the given mode on the PATH, or None if there is no such
-    file.
-    `mode` defaults to os.F_OK | os.X_OK. `path` defaults to the result
-    of os.environ.get("PATH"), or can be overridden with a custom search
-    path.
+def which(cmd, mode=os.F_OK | os.X_OK, executable_search_paths=None):  # pragma: no cover
```
sanathkr: Can you update Path Resolver design doc to explain what this is? https://github.com/awslabs/aws-lambda-builders/blob/develop/designs/Path_Resolver_DESIGN.md

tests/functional/test_cli.py
TheSriram: yes the test is present below.

aws_lambda_builders/utils.py
TheSriram: Sure, shall I do that in another PR?

tests/functional/test_cli.py
TheSriram: https://github.com/awslabs/aws-lambda-builders/pull/78/files#diff-736a032bebd5a17dd2e352d0b55f76acR115

.appveyor.yml
```diff
@@ -23,7 +23,7 @@ install:
 - "%PYTHON%\\python.exe -m pip install -e ."
 - "set PATH=C:\\Ruby25-x64\\bin;%PATH%"
 - "gem --version"
-- "gem install bundler -v 1.17.3 --no-ri --no-rdoc"
+- "gem install bundler -v 1.17.3"
```
sanathkr: yeah, this is some recent breaking change with gem

aws_lambda_builders/utils.py
sanathkr: Works

tests/functional/test_cli.py
```diff
@@ -94,4 +106,52 @@ def test_run_hello_workflow(self, flavor):
             contents = fp.read()
 
         self.assertEquals(contents, self.expected_contents)
+        shutil.rmtree(self.scratch_dir)
+
+    @parameterized.expand([
+        ("request_through_stdin"),
+        ("request_through_argument")
+    ])
+    def test_run_hello_workflow_incompatible(self, flavor):
```
sanathkr: This test is good for now, but it will need an upgrade when we change the version check logic or bump beyond 2.0.

PR #79, opened 2019-02-08
refactor(go): Updated handler to output_executable_name for go workflows

jfuss:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/aws_lambda_builders/workflows/go_dep/workflow.py
+++ b/aws_lambda_builders/workflows/go_dep/workflow.py
@@ -45,7 +45,7 @@ class GoDepWorkflow(BaseWorkflow):
                                             **kwargs)
 
         options = kwargs["options"] if "options" in kwargs else {}
-        handler = options.get("output_executable_name", None)
+        handler = options.get("artifact_executable_name", None)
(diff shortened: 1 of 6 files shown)


aws_lambda_builders/workflows/go_dep/workflow.py
```diff
@@ -45,7 +45,7 @@ def __init__(self,
                                             **kwargs)
 
         options = kwargs["options"] if "options" in kwargs else {}
-        handler = options.get("handler", None)
+        handler = options.get("output_executable_name", None)
```
sanathkr: Sorry for the nitpick. "output" doesn't convey as much. Can we call it "function_executable_name" or "lambda_executable_name" or something more specific?

jfuss: `function_executable_name` communicates this builder is only for building functions. When we start supporting layers, this may not be true anymore. I am ok with `lambda_executable_name` but I think `artifact_executable_name` is clearer (stolen from what you suggested in the SAM CLI Go Mod PR).

Issue #4, opened 2018-11-08
Feature Request: Pass arbitrary options to the build actions


sanathkr:
**Description:**
Based on experience building `claudia pack`, the team learnt that people kept wanting to add certain tweaks (eg skip the .npmrc file, skip security audits, don’t create package.lock etc). NPM keeps adding these steps into their builds. So they introduced the ability to pass arbitrary flags to NPM.

**Pros**: Easy to support these one off cases
**Cons**: This breaks one of the tenets -  "make this builder super easy to reason about". Arbitrary flags make the builder hard to reason about because the output one person gets on his machine will be different from output another person get, because of the flags they have configured.

Ideally, we should implement 80% of these requested features in the build action. But I am opening this feature request to collect use cases for passing arbitrary options to build actions.

**Request by build actions**:
1. node-npm

sanathkr:
This isn't a big deal for PIP because most arguments to PIP can be passed as environment variables.

sriram-mv:
Here is an example for using makefile for custom runtimes, but a makefile builder can still be used for officially supported lambda runtimes.

https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/building-custom-runtimes.html

This will allow passing of all options deemed necessary by the end user.

PR #80, opened 2019-02-11
Add support for .NET Core build

normj:
*Description of changes:*
Add support for building .NET Core Lambda function as part of SAM build.


aws_lambda_builders/workflows/dotnet_clipackage/DESIGN.md
```diff
@@ -0,0 +1,86 @@
...
+      -f     | --framework                    Target framework to compile, for example netcoreapp2.1.
+      --msbuild-parameters                    Additional msbuild parameters passed to the 'dotnet publish' command. Add quotes around the value if the value contains spaces.
+      -pl    | --project-location             The location of the project, if not set the current directory will be assumed.
+      -cfg   | --config-file                  Configuration file storing default values for command line arguments.
+      -pcfg  | --persist-config-file          If true the arguments used for a successful deployment are persisted to a config file.
+      -o     | --output-package               The output zip file name
+      -dvc   | --disable-version-check        Disable the .NET Core version check. Only for advanced usage.
+```
+
+Currently **--framework** is the only required parameter which tells the underlying build process what version of .NET Core to build for.
```
TheSriram: How is this infomation passed in? is it decided by the builder based on the runtime?

aws_lambda_builders/workflows/dotnet_clipackage/DESIGN.md
```diff
@@ -0,0 +1,86 @@
...
+
+The implementation is broken up into 2 steps. The first action is to make sure the Amazon.Lambda.Tools Global Tool
+is installed. The second action is to execute the `dotnet lambda package` command. 
+
+#### Step 1: Install Amazon.Lambda.Tools
+
+The tool is installed by executing the command `dotnet tool install -g Amazon.Lambda.Tools` This will install the
+tool from [NuGet](https://www.nuget.org/packages/Amazon.Lambda.Tools/) the .NET package management system. 
+
+To keep the tool updated the command `dotnet tool update -g Amazon.Lambda.Tools` will be executed if the install
```
TheSriram: Nice!

aws_lambda_builders/workflows/dotnet_clipackage/DESIGN.md
normj: For most users --framework will be already set in the aws-lambda-tools-defaults.json file so they won't have to do anything. If it is not set then the user would need to specify it on the sam build command line which would get passed into the builder. I tried to follow the examples I saw in the other builders for grabbing options that were passed on the commandline.

aws_lambda_builders/workflows/dotnet_clipackage/DESIGN.md
```diff
@@ -0,0 +1,86 @@
...
+.NET Core CLI is a required tool for .NET Core developers to build any .NET Core project.
+
+The .NET Core CLI handles the validation that the correct version of .NET Core is installed and errors out when there is 
+not a correct version. 
+
+### Challenges
+
+#### Output
+
+The output of `dotnet lambda package` command is a zip archive that consumers can then deploy to Lambda. For SAM build
```
sanathkr: Yes, this is a requirement mainly stemming from the ability to customize the zip file later during package. See this discussion - https://github.com/awslabs/aws-lambda-builders/pull/69#discussion_r256537324

aws_lambda_builders/workflows/dotnet_clipackage/DESIGN.md
sanathkr: So this is currently not supported in SAM CLI, although the correct plumbing exists thru `options` property. See https://github.com/awslabs/aws-lambda-builders/issues/4 

How big of a requirement is this? Do you know the fraction of users that use parameters vs .json file to pass this value?

normj: It is not big requirement since the framework will most likely be set through the *.json file. It is a good thing to have for the future. As long as the code is doing it in the way SAM envisions this being handled in the future is all that I require now. We should make sure SAM init version of a .NET Core project includes the json file like all of the rest of our .NET tooling does.

I also might update the Amazon.Lambda.Tools to auto detect the framework based on the project file. If/when I do this work SAM build would pull that change in automatically due to the nature of this PR always updating the version of Amazon.Lambda.Tools.

aws_lambda_builders/workflows/dotnet_clipackage/actions.py
```diff
@@ -0,0 +1,86 @@
...
+
+from aws_lambda_builders.actions import BaseAction, Purpose, ActionFailedError
+from .dotnetcli import DotnetCLIExecutionError
+
+LOG = logging.getLogger(__name__)
+
+class GlobalToolInstallAction(BaseAction):
+
+    """
+    A Lambda Builder Action which runs bundle install in order to build a full Gemfile.lock
```
TheSriram: docstring update

aws_lambda_builders/workflows/dotnet_clipackage/actions.py
```diff
@@ -0,0 +1,86 @@
...
+            try:
+                self.subprocess_dotnet.run(
+                    ['tool', 'update', '-g', 'Amazon.Lambda.Tools'],
+                )
+            except DotnetCLIExecutionError as ex:
+                raise ActionFailedError("Error configuring the Amazon.Lambda.Tools .NET Core Global Tool: " + str(ex))
+
+class RunPackageAction(BaseAction):
+    """
+    A Lambda Builder Action which vendors dependencies to the vendor/bundle directory.
```
TheSriram: docstring update

aws_lambda_builders/workflows/dotnet_clipackage/actions.py
```diff
@@ -0,0 +1,86 @@
...
+    def __init__(self, source_dir, subprocess_dotnet, artifacts_dir, options):
+        super(RunPackageAction, self).__init__()
+        self.source_dir = source_dir
+        self.subprocess_dotnet = subprocess_dotnet
+        self.artifacts_dir = artifacts_dir
+        self.options = options
+
+    def execute(self):
+        try:
+            LOG.debug("Running bundle install --deployment in %s", self.source_dir)
```
TheSriram: log statement needs update

tests/integration/workflows/dotnet_cli-package/test_dotnet.py
```diff
@@ -0,0 +1,66 @@
...
+
+        self.assertEquals(expected_files, output_files)
+
+    def test_require_parameters(self):
+        source_dir = os.path.join(self.TEST_DATA_FOLDER, "RequireParameters")
+
+        self.builder.build(source_dir, self.artifacts_dir, self.scratch_dir,
+                           source_dir,
+                           runtime=self.runtime,
+                           options={"--framework": "netcoreapp2.1", "--configuration": "Debug"})
```
TheSriram: we currently need a way to support args support in aws-sam-cli, i.e flags passed to `sam build`

aws_lambda_builders/workflows/dotnet_clipackage/workflow.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+class DotnetCliPackageWorkflow(BaseWorkflow):
+
+    """
+    A Lambda builder workflow that knows to build and package .NET Core Lambda functions
+    """
+    NAME = "DotnetCliPackageBuilder"
+
+    CAPABILITY = Capability(language="dotnet",
```
TheSriram: Do we need to add resolver and validator for the executables used by the workflow?

tests/integration/workflows/dotnet_cli-package/test_dotnet.py
normj: Is it fine to leave this hear till sam build as support for passing args?

aws_lambda_builders/workflows/dotnet_clipackage/workflow.py
normj: Do you have an example of what you mean?

aws_lambda_builders/workflows/dotnet_clipackage/actions.py
normj: Updated

aws_lambda_builders/workflows/dotnet_clipackage/actions.py
normj: Updated

aws_lambda_builders/workflows/dotnet_clipackage/actions.py
normj: Updated

tests/integration/workflows/dotnet_cli-package/test_dotnet.py
TheSriram: Yeah think thats fine.

aws_lambda_builders/workflows/dotnet_clipackage/workflow.py
TheSriram: https://github.com/awslabs/aws-lambda-builders/blob/develop/aws_lambda_builders/workflows/java_gradle/workflow.py#L54

https://github.com/awslabs/aws-lambda-builders/blob/d60920060e2f52688a1b84d8c6d635dd8d4799fc/aws_lambda_builders/workflows/java_gradle/workflow.py#L57

Those are the examples.

normj: I added a resolver to make sure the dotnet CLI is installed. 

A validator is not needed because the dotnet CLI has its own built in versioning system to match the available installed .NET SDK with the selected project. A validator would just duplicate that work and possibly get in the way.

aws_lambda_builders/workflows/dotnet_clipackage/DESIGN.md
TheSriram: Yeah, in the future when we add add `sam build --debug` mode, we would want to pass this over to lambda-builders, and the dotnet workflow to pick it up as the configuration.

This would also yield a conversation on which is higher priority, the flag or the information in the json config file.

normj: The existing Amazon.Lambda.Tools .NET Core Global tool that this wraps around already has the rule that if a value is specified in the command line it overrides any setting in the json file. I would expect that rule to apply from sam build as well.

TheSriram:
This looks good to me! All gates pass. Just a follow up question, do we need a test with any tough to build dependencies in the builder?

aws_lambda_builders/workflows/dotnet_clipackage/dotnetcli.py
```diff
@@ -0,0 +1,63 @@
...
+                             stdout=self.os_utils.pipe,
+                             stderr=self.os_utils.pipe,
+                             cwd=cwd)
+
+        out, err = p.communicate()
+
+        # The package command contains lots of useful information on how the package was created and
+        # information when the package command was not successful. For that reason the output is
+        # always written to the output to help developers diagnose issues.
+        sys.stdout.write(out.decode('utf8').strip())
```
TheSriram: Also can we write this to Logs instead?

normj: Are logs written to stdout by default during SAM build? I want users to see that information.

normj:
Closing in favor of a new clean branch

PR #81, opened 2019-02-11
[WIP] fix(mod): Allow Go Modules Builder to run in a read only container

jfuss:
*Issue #, if available:*

*Description of changes:*
SAM CLI mounts the source directory as 'ro' within the container. Due to this, Go can't write go.sum file into the directory. To support this use-case, I modified the Go Mod builder to copy the source into the scratch directory and then after building to copy the `go.sum` file to the artifact directory. Go recommends to commit the `go.sum`, so ideally this file is written to where a customer would commit this file.

*Area that need discussion:*
Writing `go.sum` to the artifact directory isn't ideally what we want but is a way to surface this file back to customers. I believe we want this to write back to the source directory to allow customers to commit this file as Go suggests but this because difficult for the container use-case SAM CLI uses (due to the mount being readonly). 
Options: 
1. Try and write this file but gracefully handle the failure of writing. This is to allows `go.sum` to be written to source directory if able to.
2. Never copy `go.sum` and ignore what Go suggests
3. Write the file to the artifact directory and let customers move it to where they want.

I would like to do Option 1, as it follows the Go patterns and what developers expect.

### DIFF
--- a/aws_lambda_builders/workflows/go_modules/actions.py
+++ b/aws_lambda_builders/workflows/go_modules/actions.py
@@ -2,6 +2,8 @@
 Action to build a Go project using standard Go tooling
 """
 
+from shutil import copy2
+
 from aws_lambda_builders.actions import BaseAction, Purpose, ActionFailedError
 from .builder import BuilderError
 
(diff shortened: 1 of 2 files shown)


sanathkr:
I am not enthusiastic about copying entire source code to scratch dir. This is going to severely slow down builds, especially non-container builds where this extra step is not necessary. 

I am more inclined towards Option 1 (w/o copy source to scratch), if we can ask gomod to write the file but fail gracefully if it cannot. 

Is this possible?

sanathkr:
The other alternative is to copy source into the container (using `docker cp`) instead of bind mounting it. This is even slower but at least is not a hack and doesn't impact non-container builds

jfuss:
This isn't needed anymore. It could be worth saving the `go.sum` file and putting it into the artifacts directory but isn't necessary right now.

Closing

PR #85, opened 2019-02-14
backcompat: cli

TheSriram:
- bugfix: make executable_search_paths computable if not specified and add it as
      `None`.

- explicitly check with a test with older RPC version and removal of
      `executable_search_paths` param.

- This ensures that if the container is upgraded to the latest version of builders, and we build on the container from a older version of SAM CLI, back compat does not break.


sanathkr:
Lgtm

PR #87, opened 2019-02-18
Add design doc and implementations for Java Maven builder

zoewangg:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/aws_lambda_builders/workflows/java_maven/DESIGN.md
+++ b/aws_lambda_builders/workflows/java_maven/DESIGN.md
@@ -7,7 +7,7 @@ projects managed using the Maven build tool.
 
 ## Challenges
 
-- Java Version compatibility mentioned in the java_grade Design doc.
+- Java Version compatibility mentioned in the [Gradle Lambda Builder] Design doc.
 
 - Building Multimodule project
 
@@ -52,18 +52,7 @@ assumes aws-lambda-builders have knowledge of the root package path for multimod
 
 ### Build Workflow
 
-We leverage Maven to do all the heavy lifting for executing the
-`mvn package` which will resolve and download the dependencies and
-build the project.
-
-#### Step 1: Copy source project to scratch directory
-
-By default, Maven stores its build-related metadata in a `target`
-directory under the source directory and there is no way to change the output 
-directory from command line. To avoid writing anything under `source_dir`, 
-we copy the source project to scratch directory and build it from there.
-
-#### Step 2: Check Java version and emit warning
+#### Step 1: Check Java version and emit warning
 
 Check whether the local JDK version is <= Java 8, and if it is not, emit a
 warning that the built artifact may not run in Lambda unless a) the project is
@@ -73,26 +62,66 @@ built within a Lambda-compatibile environment like `lambci`.
 We use Maven to check the actual JVM version Maven is using in case it has been 
 configured to use a different one than can be found on the PATH.
 
-#### Step 3: Build and package
+#### Step 2: Build and package
+
+We leverage Maven to do all the heavy lifting for executing the`mvn package` which
+will resolve and download the dependencies and build the project. Generate java classes 
+will be located in `target/classes`. Then we use `mvn dependency:copy-dependenceis` to copy
+the dependencies and the dependencies will be located in `target/dependency` under the 
+source directory in `scratch_dir`.
+
+##### Single project
 
 ```sh
-# Single Project
-mvn package
+mvn package 
 mvn dependency:copy-dependencies
-
-# Multimodule Project
-cd /path/to/rootdirectory
-mvn install -pl :MODULE_TO_BUILD -also-make
-mvn dependency:copy-dependencies -pl :MODULE_TO_BUILD
 ```
 
-Generate java classes will be located in `target/classes` and dependencies 
-will be located in `target/dependency` under the source directory in `scratch_dir`.
+##### Multimodule project
 
-#### Step 4: Copy to artifact directory
+```bash
+cd /path/to/rootdirectory
+mvn install -pl :$SOURCE_DIRECTORY_NAME --also-make
+mvn dependency:copy-dependencies -pl :$SOURCE_DIRECTORY_NAME
+```
+
+Here `$SOURCE_DIRECTORY_NAME` is the name of the source directory of the project. Maven
+will build the dependencies of that project in the reactor and then build the project itself.
+Note that `mvn install` is being used for multimodule project and this is because 
+`dependency:copy-dependencies`  requires all dependencies being installed in the local repository.
+
+```bash
+# Example commands to build ProjectB/lambda1
+cd ..
+mvn install -pl :lambda1 --also-make # build common first then lambda1
+mvn dependency:copy-dependencies -pl :lambda1
+```
+
+#### Step 3: Copy to artifact directory
 
 The workflow implementation is aware of the mapping scheme used to map a
 `source_dir` to the correct directory under `scratch_dir` (described in step 4),
 so it knows where to find the built Lambda artifact when copying it to
 `artifacts_dir`. They will be located in
 `$SCRATCH_DIR/<mapping for source_dir>/target`.
+
+#### Step 4: Clean up target directory
+
+By default, Maven stores its build-related metadata in a `target`
+directory under the source directory and we will run `mvn clean` to clean up
+the directory.
+
+##### Single project
+
+```sh
+mvn clean 
+```
+
+##### Multimodule project
+
+```sh
+cd /path/to/rootdirectory
+mvn clean
+```
+
+[Gradle Lambda Builder]:https://github.com/awslabs/aws-lambda-builders/blob/develop/aws_lambda_builders/workflows/java_gradle/DESIGN.md
--- a/aws_lambda_builders/workflows/java_maven/DESIGN.md
+++ b/aws_lambda_builders/workflows/java_maven/DESIGN.md
@@ -52,7 +52,17 @@ assumes aws-lambda-builders have knowledge of the root package path for multimod
 
 ### Build Workflow
 
-#### Step 1: Check Java version and emit warning
+#### Step 1: Copy source project to scratch directory
+
+By default, Maven stores its build-related metadata in a `target`
+directory under the source directory and there is no way to change the output 
+directory from command line. To avoid writing anything under `source_dir`, 
+we copy the source project to scratch directory and build it from there.
+
+*TODO*: For multi-module project, we need to know the root directory
+in order to copy the whole project.
+
+#### Step 2: Check Java version and emit warning
 
 Check whether the local JDK version is <= Java 8, and if it is not, emit a
 warning that the built artifact may not run in Lambda unless a) the project is
@@ -62,7 +72,7 @@ built within a Lambda-compatibile environment like `lambci`.
 We use Maven to check the actual JVM version Maven is using in case it has been 
 configured to use a different one than can be found on the PATH.
 
-#### Step 2: Build and package
+#### Step 3: Build and package
 
 We leverage Maven to do all the heavy lifting for executing the`mvn package` which
 will resolve and download the dependencies and build the project. Built java classes 
@@ -70,19 +80,19 @@ will be located in `target/classes`. Then we use `mvn dependency:copy-dependence
 the dependencies and the dependencies will be located in `target/dependency` under the 
 source directory.
 
-##### Single project
+##### Single-module project
 
-```sh
-mvn package 
-mvn dependency:copy-dependencies
+```bash
+mvn clean package 
+mvn dependency:copy-dependencies -DincludeScope=compile
 ```
 
-##### Multimodule project
+##### Multi-module project
 
 ```bash
 cd /path/to/rootdirectory
-mvn install -pl :$SOURCE_DIRECTORY_NAME --also-make
-mvn dependency:copy-dependencies -pl :$SOURCE_DIRECTORY_NAME
+mvn clean install -pl :$SOURCE_DIRECTORY_NAME --also-make
+mvn dependency:copy-dependencies -pl :$SOURCE_DIRECTORY_NAME -DincludeScope=compile
 ```
 
 Here `$SOURCE_DIRECTORY_NAME` is the name of the source directory of the project. Maven
@@ -93,32 +103,13 @@ Note that `mvn install` is being used for multimodule project and this is becaus
 ```bash
 # Example commands to build ProjectB/lambda1
 cd ..
-mvn install -pl :lambda1 --also-make # build common first then lambda1
-mvn dependency:copy-dependencies -pl :lambda1
+mvn clean install -pl :lambda1 --also-make # build common first then lambda1
+mvn dependency:copy-dependencies -pl :lambda1 -DincludeScope=compile
 ```
 
-#### Step 3: Copy to artifact directory
+#### Step 4: Copy to artifact directory
 
-Built Java classes and depedenceis are copied from `source_dir/target/classes` and `source_dir/target/dependency`
-respectively to the artifact directory.
-
-#### Step 4: Clean up target directory
-
-By default, Maven stores its build-related metadata in a `target`
-directory under the source directory and we run `mvn clean` to clean up
-the directory.
-
-##### Single project
-
-```sh
-mvn clean 
-```
-
-##### Multimodule project
-
-```sh
-cd /path/to/rootdirectory
-mvn clean
-```
+Built Java classes and dependencies are copied from `scratch_dir/target/classes` and `scratch_dir/target/dependency`
+to `artifact_dir` and `artifact_dir/lib` respectively.
 
 [Gradle Lambda Builder]:https://github.com/awslabs/aws-lambda-builders/blob/develop/aws_lambda_builders/workflows/java_gradle/DESIGN.md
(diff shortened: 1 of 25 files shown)


aws_lambda_builders/workflows/java_maven/DESIGN.md
```diff
@@ -0,0 +1,98 @@
...
+#### Step 3: Build and package
+
+```sh
+# Single Project
+mvn package
+mvn dependency:copy-dependencies
+
+# Multimodule Project
+cd /path/to/rootdirectory
+mvn install -pl :MODULE_TO_BUILD -also-make
```
TheSriram: How is module to build established? can you expand on that? when do we consider a build complete, when we build all the modules?

so we do need to run

```
mvn install -pl :MODULE_TO_BUILD -also-make
```

for all modules present in the codebase?

zoewangg: Yeah, I'll add more details. 

The module name should be same as the `source_dir` name. If the customer wants to build `ProjectB/lambda1`, the module name should be `lambda1`. 

We only need to run the command for the module that we want to build and it will build the dependencies under the same parent for that module first and then the module itself.

So for `ProjectB/lambda1`,  below are the commands we run:
```bash
mvn install -pl :lambda1 --also-make  # It would build `common` first and then `lambda1`
mvn dependency:copy-dependencies -pl :lambda1 # copy the dependencies of `lambda1`.
```

aws_lambda_builders/workflows/java_maven/DESIGN.md
```diff
@@ -0,0 +1,98 @@
...
+│   └── pom.xml
+├── pom.xml
+└── template.yaml
+```
+
+Building Project A is relatively simple since we just need to issue `mvn
+package` and place the built classes and dependency jars within the artifact directory.
+
+Building `ProjectB/lambda1` requires maven to build `lambda1` module from
+the root pom directory and use `--also-make` option to build necessary dependencies
```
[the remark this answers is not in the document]
zoewangg: Yeah, finding the root path is the challenging part. There's no easy way to find the root path from maven especially for the projects that have multiple layer of submodules. I assume `use-container` would also need to find the parent package when mounting.  I understand this is still under design, but is that something SAM could pass to the builder as well?

aws_lambda_builders/workflows/java_maven/DESIGN.md
```diff
@@ -0,0 +1,124 @@
...
+built within a Lambda-compatibile environment like `lambci`.
+
+We use Maven to check the actual JVM version Maven is using in case it has been 
+configured to use a different one than can be found on the PATH.
+
+#### Step 2: Build and package
+
+We leverage Maven to do all the heavy lifting for executing the`mvn package` which
+will resolve and download the dependencies and build the project. Built java classes 
+will be located in `target/classes`. Then we use `mvn dependency:copy-dependenceis` to copy
```
sanathkr: this `target` folder should not be within the `lambda` directory but instead within a scratch directory that you will get as part of the workflow. One of the design goals is to prevent the builder from writing to the Function's source as much as possible.

aws_lambda_builders/workflows/java_maven/DESIGN.md
```diff
@@ -0,0 +1,124 @@
...
+
+```sh
+mvn package 
+mvn dependency:copy-dependencies
+```
+
+##### Multimodule project
+
+```bash
+cd /path/to/rootdirectory
```
sanathkr: You cannot differentiate between Single Project & Multimodule project within the workflow. Lambda Builder does not have that information. 

Can't you do the same operation for both?

aws_lambda_builders/workflows/java_maven/DESIGN.md
```diff
@@ -0,0 +1,124 @@
...
+mvn install -pl :$SOURCE_DIRECTORY_NAME --also-make
+mvn dependency:copy-dependencies -pl :$SOURCE_DIRECTORY_NAME
+```
+
+Here `$SOURCE_DIRECTORY_NAME` is the name of the source directory of the project. Maven
```
sanathkr: You do not get a path to `$SOURCE_DIRECTORY_NAME` within the Lambda Builder. If you cannot reference `common/pom.xml` from within `lambda1/pom.xml` then is this multi-module case even technically feasible?

aws_lambda_builders/workflows/java_maven/DESIGN.md
```diff
@@ -0,0 +1,124 @@
...
+```bash
+# Example commands to build ProjectB/lambda1
+cd ..
+mvn install -pl :lambda1 --also-make # build common first then lambda1
+mvn dependency:copy-dependencies -pl :lambda1
+```
+
+#### Step 3: Copy to artifact directory
+
+Built Java classes and depedenceis are copied from `source_dir/target/classes` and `source_dir/target/dependency`
```
sanathkr: again, they should not be written directly to `source_dir`. Can you ask maven to write to `scratch_dir/target/` instead?

aws_lambda_builders/workflows/java_maven/DESIGN.md
```diff
@@ -0,0 +1,124 @@
...
+respectively to the artifact directory.
+
+#### Step 4: Clean up target directory
```
sanathkr: You don't need to cleanup if you use `scratch_dir` because SAM CLI will do that for you.

aws_lambda_builders/workflows/java_maven/DESIGN.md
```diff
@@ -0,0 +1,124 @@
...
+
```
sanathkr: Also, do you want to run a `mvn clean` **before** building so you always get clean builds? What does `mvn clean` do besides deleting the `target` folder?

aws_lambda_builders/workflows/java_maven/DESIGN.md
zoewangg: I see. I was actually using `CopySourceAction` to copy the source code to the scratch directory in my first commit. Maven doesn't support change the target directory from command line, so we'd need to copy the whole project and it could take long time especially for large project. Will update.

aws_lambda_builders/workflows/java_maven/DESIGN.md
zoewangg: For multimodule project, we need to run the build command from the **root** directory and build using `--also-make` option because Maven cannot find its way back to the parent. The challenging part is finding the root directory. I assume use-container would also need to find the parent package when mounting. I understand this is still under design, but is that something SAM could pass to the builder as well?

```
cd ..
mvn install -pl :lambda1 --also-make 
mvn dependency:copy-dependencies -pl :lambda1
```

aws_lambda_builders/workflows/java_maven/DESIGN.md
zoewangg: `$SOURCE_DIRECTORY_NAME` is just the name of the source directory, so it would be `os.path.basename(os.path.dirname(source_dir))`.

Multimodule case is possible if the builder has knowledge of the project root directory

aws_lambda_builders/workflows/java_maven/DESIGN.md
zoewangg: Maven doesn't support changing the target directory from command line, so we'd need to copy the whole project to `scratch_dir`.

aws_lambda_builders/workflows/java_maven/DESIGN.md
zoewangg: Yeah, will remove

aws_lambda_builders/workflows/java_maven/DESIGN.md
zoewangg: Yes, will add `clean`. All it does is just deleting the `target` folder.

aws_lambda_builders/workflows/java_maven/DESIGN.md
zoewangg: For multimodule project, builder also needs to know the root directory in order to copy the whole project.

aws_lambda_builders/workflows/java_maven/DESIGN.md
```diff
@@ -0,0 +1,94 @@
...
+By default, Maven stores its build-related metadata in a `target`
+directory under the source directory and there is no way to change the output 
+directory from command line. To avoid writing anything under `source_dir`, 
+we copy the source project to scratch directory and build it from there.
+
+#### Step 2: Check Java version and emit warning
+
+Check whether the local JDK version is <= Java 8, and if it is not, emit a
+warning that the built artifact may not run in Lambda unless a) the project is
+properly configured (i.e. using `targetCompatibility`) or b) the project is
```
dagnir: Not sure if there's a config in Maven with the same name, but I think the option is different (it's `target`) https://maven.apache.org/plugins/maven-compiler-plugin/examples/set-compiler-source-and-target.html

zoewangg: Yup. fixed

aws_lambda_builders/workflows/java_maven/DESIGN.md
```diff
@@ -0,0 +1,94 @@
...
+#### Step 3: Build and package
+
+We leverage Maven to do all the heavy lifting for executing the`mvn package` which
+will resolve and download the dependencies and build the project. Built java classes 
+will be located in `target/classes`. Then we use `mvn dependency:copy-dependenceis` to copy
+the dependencies and the dependencies will be located in `target/dependency` under the 
+source directory.
+
+```bash
+MODULE_NAME=$(mvn -q -Dexec.executable=echo -Dexec.args='${project.artifactId}' exec:exec --non-recursive)
```
TheSriram: Does this work in windows too? just checking.

aws_lambda_builders/workflows/java_maven/DESIGN.md
```diff
@@ -0,0 +1,94 @@
...
+especially for the projects with multiple level of submodules. Building multi-module project is
+out of scope for the current version.
+
+## Implementation
+
+### Build Workflow
+
+#### Step 1: Copy source project to scratch directory
+
+By default, Maven stores its build-related metadata in a `target`
```
TheSriram: This will cause slowdown on large projects, is there any other alternative to this?

aws_lambda_builders/workflows/java_maven/DESIGN.md
```diff
@@ -0,0 +1,94 @@
...
+directory under the source directory and there is no way to change the output 
+directory from command line. To avoid writing anything under `source_dir`, 
+we copy the source project to scratch directory and build it from there.
+
+#### Step 2: Check Java version and emit warning
```
TheSriram: This is awesome!

aws_lambda_builders/workflows/java_maven/maven_validator.py
```diff
@@ -0,0 +1,64 @@
+"""
+Maven Binary Validation
```
TheSriram: Is this radically different from the gradle validator? If so, then thats fine. Im  just trying to see if they work in very similar ways.

aws_lambda_builders/workflows/java_maven/DESIGN.md
zoewangg: Yes, the tests passed on windows.

aws_lambda_builders/workflows/java_maven/DESIGN.md
zoewangg: Unfortunately no. Maven doesn't support changing the `target` directory from command line. We'd need to update the pom.xml file to change it.

aws_lambda_builders/workflows/java_maven/maven_validator.py
zoewangg: It works similarly to Gradle validator. It basically runs `mvn -version` and checks the printed out jvm version.

aws_lambda_builders/workflows/java_maven/DESIGN.md
sanathkr: One crazy idea: Can the workflow generate a pom.xml which customizes the build directory folder [1] and somehow "include" the Lambda function's pom.xml? The workflow can generate this file within scratch directory, and run `mvn` on this pom which would also run on the function's pom.

I know the gradle workflow does something similar. So is this possible?

[1] Generated pom.xml would have something like this https://stackoverflow.com/a/12598554

zoewangg: Technically we could do this by injecting a ` <profile>` to a copy of customer's pom.xml file. However, unlike Gradle, Maven doesn't support it natively, i.e, we'd need to do some xml parsing and validating in order to insert the xml blob to the right place. Not sure if that's what we want.

sanathkr: - Is generating the a pom to be the parent of function's pom not possible?
- Do we really need to parse the XML? Can we insert something to the end-ish by using regex to find the end tag? 

I really don't want to be in the business of parsing XMLs, but I want to make sure we have exhausted all possibilities before we think about it.

zoewangg: > Is generating the a pom to be the parent of function's pom not possible?

There are more files to update with this approach.

- If the project is single-module, we need to add the following module to the customer's pom.
```
    <parent>
        <artifactId>fake-parent</artifactId>
        <groupId>groupId</groupId>
        <version>VERSION</version>
    </parent>
```
- If the project is multi-module project itself, on top of what we do in the above, we'd also need to update the project's original parent pom with
```
 <modules>
        <module>fake-parent</module>
<modules>
```

> Do we really need to parse the XML? Can we insert something to the end-ish by using regex to find the end tag?

It requires some parsing.  `<profile/>` has to be inside of `<profiles/>`, we need to firstly check if customer has profiles and if yes, insert it after the last `</profile>` tag and before `</profiles>`

If not, we need to insert 
```
<profiles>
  <profile>...</profile>
</profiles>
```

zoewangg: We can have more discussions on this and optimize this later and I don't think it should be be blocker for the java maven builder. What do you think?

aws_lambda_builders/workflows/java_maven/actions.py
```diff
@@ -0,0 +1,98 @@
...
+                 os_utils):
+        self.scratch_dir = scratch_dir
+        self.artifacts_dir = artifacts_dir
+        self.os_utils = os_utils
+
+    def execute(self):
+        self._copy_artifacts()
+
+    def _copy_artifacts(self):
+        lambda_build_output = os.path.join(self.scratch_dir, 'target', 'classes')
```
jfuss: Should we have `exists` checks here if these are required? I ran into this path not being created by maven, which ended up being user error but had little to no information on how or what went wrong.

Maybe something like:
```
if not exists:
  LOG.info("Required directory(s) [X,Y] were not found. Please check pom.xml is located in the right directory.")
```

This might not encompass everything but drives the point of providing more information to customers.

aws_lambda_builders/workflows/java_maven/actions.py
```diff
@@ -0,0 +1,98 @@
...
+        dependency_output = os.path.join(self.scratch_dir, 'target', 'dependency')
+
+        try:
+            self.os_utils.copytree(lambda_build_output, self.artifacts_dir)
+            if self.os_utils.exists(dependency_output):
```
jfuss: Should we debug log if this directory doesn't exist? This could provide insight into transient errors or miss configurations in the pom.xml?

aws_lambda_builders/workflows/java_maven/maven.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def retrieve_module_name(self, scratch_dir):
+        args = ['-q', '-Dexec.executable=echo', '-Dexec.args=${project.artifactId}',
+                'exec:exec', '--non-recursive']
+        ret_code, stdout, stderr = self._run(args, scratch_dir)
+        if ret_code != 0:
+            raise MavenExecutionError(message=stderr.decode('utf8').strip())
+        return stdout.decode('utf8').strip()
+
+    def build(self, scratch_dir, module_name):
```
jfuss: Might be a good idea to `LOG.info(stdout) or LOG.debug(stdout)`. When the build fails, it could be useful to inspect maven output to see what happened.

aws_lambda_builders/workflows/java_maven/workflow.py
```diff
@@ -0,0 +1,58 @@
+"""
+Java Maven Workflow
+"""
+from aws_lambda_builders.workflow import BaseWorkflow, Capability
+from aws_lambda_builders.workflows.java_gradle.utils import OSUtils
```
jfuss: We don't want to be importing other workflows modules. Even if it is the same, please copy into a maven specific OSUtils. We know there is reuse across workflows and want to solve this all at once. If workflows start importing other workflow classes, it will make it difficult to untangle the imports. Each workflow should be isolated from each other.

aws_lambda_builders/workflows/java_maven/maven_validator.py
```diff
@@ -0,0 +1,64 @@
...
+"""
+
+import logging
+import re
+
+from aws_lambda_builders.workflows.java_gradle.utils import OSUtils
```
jfuss: Create a maven specific OSUtils.

aws_lambda_builders/workflows/java_maven/maven_resolver.py
```diff
@@ -0,0 +1,23 @@
+"""
+Maven executable resolution
+"""
+
+from aws_lambda_builders.workflows.java_gradle.utils import OSUtils
```
jfuss: Maven specific

aws_lambda_builders/workflows/java_maven/DESIGN.md
sanathkr: Understood. Let's skip this optimization for now. I was just trying to see if there are other options. We can always make it better later

PR #88, opened 2019-02-22
[WIP] Design for generic builder interface

sam-goodwin:
Proposal to extend `aws-lambda-builders` to become a generic library for building all AWS projects.

Still more details to be worked out, esp. around the UX experience and implementation details. Looking forward to discussion from those with more context.

Supports: https://github.com/awslabs/aws-cdk/issues/1435

### DIFF
--- a/designs/Generic_Builder_Interface.md
+++ b/designs/Generic_Builder_Interface.md
@@ -1,6 +1,6 @@
 Generic Builder Interface
 =========================
-This proposal suggests we evolve `aws-lambda-builders` to become a generic library for building all AWS projects. The `aws-lambda-builders` project stands as a testament to the breadth of this problem domain, revealing an opportunity for teams across AWS to contribute their respective workflows to a single library.
+This proposal suggests we evolve `aws-lambda-builders` to become a generic library for building all AWS code project types such as Lambda and ECS, providing a common library through which customer facing tools such as the CDK, SAM and ECS's CLI build and deploy customer's code.
 
 What is the problem?
 --------------------
@@ -25,7 +25,9 @@ Taking a dependency on Docker is justified as follows:
 
 Success criteria for the change
 -------------------------------
-A single library and CLI exists, encapsulating the logic of building deployable artifacts from source without inspecting the project's internal structure (e.g. looking for `pom.xml` or `build.gradle`). This CLI may be installed directly into the developer's environment or consumable as encapsulated Docker images specific to various langauges/environments.
+A single library and CLI exists, encapsulating the logic to build deployable artifacts from arbitrary code workspaces. It should attempt to infer boiler-plate details such as maven vs gradle for building java, requiring only the language target from the caller. "This is a java 1.8 project, build it for me".
+
+This CLI may be installed directly into the developer's environment or consumable as encapsulated Docker images specific to various langauges/environments.
 
 User Experience Walkthrough
 ---------------------------
@@ -40,7 +42,7 @@ const myFunction = new lambda.JavaFunction(stack, 'MyJavaFunction', {
 });
 ```
 
-Then as part of the CLI command, `cdk synth` or `cdk deploy`, the toolkit will run a docker command to build the project into the `cdk.out` folder. In this example, we know to select the image for Java 1.8 because we know the function's runtime. 
+The CLI command `cdk synth` or `cdk deploy` will run a docker command to build the project into the `cdk.out` folder. In this example, we know to select the image for Java 1.8 because we know the function's runtime. 
 
 The payload is a stripped down version of the existing RPC format,
 ```json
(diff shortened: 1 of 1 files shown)


designs/Generic_Builder_Interface.md
```diff
@@ -0,0 +1,92 @@
+Generic Builder Interface
+=========================
+This proposal suggests we evolve `aws-lambda-builders` to become a generic library for building all AWS code project types such as Lambda and ECS, providing a common library through which customer facing tools such as the CDK, SAM and ECS's CLI build and deploy customer's code.
+
+What is the problem?
+--------------------
+Tools like the [Cloud Development Kit (CDK) CLI](https://github.com/awslabs/aws-cdk) and the [Elastic Contianer Service (ECS) CLI](https://github.com/aws/amazon-ecs-cli) require a single library to encapsulate the logic of assembling a project's source code into an artifact deployable to AWS. Since the CDK covers the entire surface area of AWS, not just serverless use-cases like SAM, the scope of `aws-lambda-builders` must be extended to include non-lambda target environments such as Docker images for ECR. This also asks questions of branding, perhaps requiring a new project named `aws-builders`.
```
eladb: I would change "must be extended" to "could be extended"

eladb:
Copy @clareliguori @SoManyHs @rix0rrr

designs/Generic_Builder_Interface.md
```diff
@@ -0,0 +1,92 @@
...
+
+User Experience Walkthrough
+---------------------------
+
+*TODO: More details/use-cases*
+
+For the CDK use-case, we wish to instantiate a language-specific function type and only have to point to the project's location:
+
+```ts
+const myFunction = new lambda.JavaFunction(stack, 'MyJavaFunction', {
```
rix0rrr: Couldn't we make use of the newly centralized build/language detection logic to get rid of the need of specifying `Java` here?

rix0rrr: Also, strictly speaking, shouldn't this be called `JvmFunction`? :)

designs/Generic_Builder_Interface.md
```diff
@@ -0,0 +1,92 @@
...
+*TODO: Explain how this feature will be implemented. Highlight the components
+of your implementation, relationships* *between components, constraints,
+etc.*
+
+Open Issues
+-----------
+1. Do we want to rename this project or create a new package encapsulating the generic functionality?
+2. What to do about `pip install`? Is there a better way to expose the CLI for local installation or inclusion in a `Dockerfile`?
+3. Which environments/configurations should we publish Docker containers for? E.g. one container for NPM, one for Java, one for Python, etc. etc.
+4. Do we want to provide hooks for developers to customize a build when their use-case is not supported?
```
rix0rrr: Lack of customization hooks seems like a nonstarter. What if a project build requires additional installed tool dependencies, or private package repositories, or ANYTHING, really. Let's give people the ability to build what they need from the start, and add incremental sugar later.

designs/Generic_Builder_Interface.md
```diff
@@ -0,0 +1,92 @@
...
+5. What impact does supporting non-lambda targets have on the current architecture?
+
+Task Breakdown
+--------------
+
+-   \[x\] Send a Pull Request with this design document
+-   \[ \] Extract project inference from SAM CLI to `aws-lambda-builders`
+-   \[ \] Update RPC to support the higher-level use cases (project inference)
+-   \[ \] Publish a Docker image for each language we support (javascript, java, python, go mod, go dep)
```
rix0rrr: Build a pipeline for all images we vend!

designs/Generic_Builder_Interface.md
eladb: There's a different APIs for defining the build project for different languages. Different way to specify dependencies/handlers/runtime versions. I like the strong typing.

designs/Generic_Builder_Interface.md
eladb: customization hooks are usually built into the language specific build system. Maven has like a million hooks, npm has hooks for pre-install post-install, etc.

sanathkr: @rix0rrr #4 is tracking passing arbitrary options to build tools. We didn't enable this from the beginning because we wanted builds to work out-of-box with majority of cases without additional work.

clareliguori:
Agree with @eladb, I would start without targeting ecs-cli.  I'm not actually sure it makes sense for ecs-cli to leverage this library: the ecs-cli is intended to fit into a user's existing docker-compose workflow, so they use docker-compose directly to assemble everything.  In which case, the user does something like
```
docker-compose build
docker-compose push
ecs-cli -f my-compose-file.yml service up
```

For CDK +  ECS, there is already a Docker "asset" type for building and pushing the Docker image.

sam-goodwin:
Preliminary work in the CDK can be seen here: https://github.com/awslabs/aws-cdk/pull/1884

Would love your feedback @sanathkr

PR #96, opened 2019-03-11
fix: Remove MODULE_NAME command from Maven Workflow

jfuss:
On Windows, the echo command is not valid in most cases. The Maven Workflow
was doing some operations that help support the multiple project build
use-case. Given this sometimes fails on Windows, the PR updates the Maven
Worklfow to only install and copy dependencies. This removes the need
to echo and therefore allows Windows Maven builds to succeed without
needing to run within the container.

*Issue #, if available:*
Will solve https://github.com/awslabs/aws-sam-cli/issues/1053, when released within SAM CLI.

*Description of changes:*

### DIFF
--- a/aws_lambda_builders/workflows/java_maven/DESIGN.md
+++ b/aws_lambda_builders/workflows/java_maven/DESIGN.md
@@ -71,20 +71,16 @@ configured to use a different one than can be found on the PATH.
 
 #### Step 3: Build and package
 
-We leverage Maven to do all the heavy lifting for executing the`mvn package` which
+We leverage Maven to do all the heavy lifting for executing the`mvn clean install` which
 will resolve and download the dependencies and build the project. Built java classes 
 will be located in `target/classes`. Then we use `mvn dependency:copy-dependenceis` to copy
 the dependencies and the dependencies will be located in `target/dependency` under the 
 source directory.
 
 ```bash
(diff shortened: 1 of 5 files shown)


aws_lambda_builders/workflows/java_maven/DESIGN.md
```diff
@@ -71,21 +71,17 @@ configured to use a different one than can be found on the PATH.
...
 will resolve and download the dependencies and build the project. Built java classes 
 will be located in `target/classes`. Then we use `mvn dependency:copy-dependenceis` to copy
 the dependencies and the dependencies will be located in `target/dependency` under the 
 source directory.
 
 ```bash
-MODULE_NAME=$(mvn -q -Dexec.executable=echo -Dexec.args='${project.artifactId}' exec:exec --non-recursive)
-mvn clean package -pl :MODULE_NAME -am
-mvn dependency:copy-dependencies -DincludeScope=compile -pl :MODULE_NAME
+mvn clean install
```
TheSriram: Just checking, this works on windows too right?

jfuss: Ya. I patched the windows machine that was failing before this change and was able to build locally on the windows machine with Maven.

jfuss:
@TheSriram Can you please re-review? I had to handle some merge conflicts and would prefer another set of eyes before merging.

PR #98, opened 2019-03-13
fix: integ tests (make source dir read only)

sriram-mv:
integration tests need to check that source directory is in read-only mode, so that no artifacts are produced on the source_dir.

Dotnet and Go have been left out of this PR, because their specific builders do not adhere to this rule.


TheSriram:
Agree about the switch, Architecture of the tests might need to be changed a bit, to know about the source dir beforehand, during setUp maybe.

sanathkr:
pytest fixtures can be easily used to abstract this away from all tests - https://docs.pytest.org/en/2.7.3/fixture.html#using-fixtures-from-classes-modules-or-projects

@TheSriram Can you try this to simplify the readonly change?

TheSriram:
@sanathkr I didnt know about that. that looks really good actually! will try to simplify this.

Issue #104, opened 2019-03-21
Java version can not be matched correctly


LiuZhiyan:
**Description:**

`sam build` failed on Oracle java 12 environment (on mac).

**Steps to reproduce the issue:**
1. Setup an Oracle java 12 development environment on mac. (maven and sam-cli were installed by brew)
2. Execute `sam build` command in project directory.

**Observed result:**

$ sam build
2019-03-20 20:08:43 Building resource 'HelloWorldFunction'
Traceback (most recent call last):
  File "/usr/local/bin/sam", line 11, in <module>
    load_entry_point('aws-sam-cli==0.13.0', 'console_scripts', 'sam')()
  File "/usr/local/Cellar/aws-sam-cli/0.13.0/libexec/lib/python3.7/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/Cellar/aws-sam-cli/0.13.0/libexec/lib/python3.7/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/local/Cellar/aws-sam-cli/0.13.0/libexec/lib/python3.7/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/Cellar/aws-sam-cli/0.13.0/libexec/lib/python3.7/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/Cellar/aws-sam-cli/0.13.0/libexec/lib/python3.7/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/Cellar/aws-sam-cli/0.13.0/libexec/lib/python3.7/site-packages/click/decorators.py", line 64, in new_func
    return ctx.invoke(f, obj, *args[1:], **kwargs)
  File "/usr/local/Cellar/aws-sam-cli/0.13.0/libexec/lib/python3.7/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/Cellar/aws-sam-cli/0.13.0/libexec/lib/python3.7/site-packages/samcli/commands/build/command.py", line 95, in cli
    skip_pull_image, parameter_overrides)  # pragma: no cover
[...]

PR #105, opened 2019-03-21
fix: match correct java version when only major version provided (#104)

LiuZhiyan:
Use non-greedy match in the regular expression, to handle the case of `mvn -version` provided the major part of java version only.


LiuZhiyan:
@TheSriram, thanks for your review, would you like to merge this? pls let me know if there are any other works need to be added to this pr.

PR #106, opened 2019-03-24
Support for local dependencies for Node.js build command

stojanovic:
*Issue #, if available:*
Issue is not opened.

*Description of changes:*
Right now, Node.js builder can resolve dependencies from NPM or git, but it doesn't know how to resolve local dependencies. This PR allows users to have local dependencies, and even nested local dependencies in Node.js. For example, your package.json file can contain something similar to this:

```js
{
  "name": "awesome-serverless-function",
  "version": "1.0.0",
  "dependencies": {
    "minimal-request-promise": "^1.3.0", // NPM dependency
    "logger": "file:../logger" // local dependency
  }
}
```


TheSriram:
This requires 2 ways to think about it. Supporting the local use case (no mounting here, so it just works!) and then supporting this to work inside a container. Any local dependencies that are not mounted along with source code will break within the container, because they wont resolve.

peterabarry:
It would be helpful if some workaround was covered here, as well as, perhaps disambiguating from layers, or other alternatives. Claudia pack is suggested as a workaround but it not that clear how exactly to use that in a workflow.

jfuss:
There was some conversations about this on this [issue](https://github.com/awslabs/aws-sam-cli/issues/756) but will add details here as well. 

This is something we want to support but we have a some challenges due to how we currently build functions in SAM CLI, which is what is holding this up not because we don't want it merged.

SAM CLI uses this library in order to build Lambda functions. SAM CLI mounts code into the container for each function. For the local dependency case (in the container), there is no way to build with local dependencies unless all the code is mounted, not just the CodeUri. To support this we need to change how we mount code into the container, most likely mount the project instead and execute `sam build` in the container. This fundamentally changes the current architecture. Now this could work in the non-container build but comes with a usability concern. That is customers who need local dependencies in build would fail if they try to do `--use-container` and cause further confusion on "this build works in the non-container build but not the container build“ (which is feedback we get on our current python build experience. To truly support building with local dependencies, we want to enable both use-cases.

As workarounds, you will need to build outside of `sam build`. You can use tools like Claudia or Webpack to achieve this in the meantime.

stojanovic:
I'll close this PR as it will not be merged anyway.

jfuss:
@stojanovic 

> I'll close this PR as it will not be merged anyway.

That isn't completely true. Our intention is to merge this but we need to have additional work done in SAM CLI to make sure we can fully support the local dependency case in the `--use-container` case. Otherwise, it becomes another place for confusion.

stojanovic:
It's 11 months old right now, and I guess it will be more than 18 months old when you can merge it, so I am sure it will be more complex to resolve conflicts than to rewrite it. I am no longer using the `sam build` command, anyway.

PR #109, opened 2019-04-08
fix(dotnet) : set absolute paths on the artifacts zip

TheSriram:
*Issue #, if available:*

https://github.com/awslabs/aws-sam-cli/issues/1101

*Description of changes:*

With windows testing, seems to be working, putting it out there, before I investigate more.


jfuss:
This PR does solve it but solves it at the workflow level. I submitted a PR in SAM CLI (https://github.com/awslabs/aws-sam-cli/pull/1110) which makes sure the build_dir is always resolved. I gave a full explanation https://github.com/awslabs/aws-sam-cli/issues/1101#issuecomment-481317178 for the more curious.  

Closing this PR in favor of the fix in SAM CLI, as it will fix it for all workflows in the future.

PR #112, opened 2019-04-25
fix output handling

abetaev:
Hi,

*Issue*
If `prepack` script is defined `npm` seems to output more than expected in original code.

*Description of changes:*
`npm pack` outputs filename as last line of stdout, while in code the whole its output was assigned to `tarfile_name` var. This fix makes it right.

Thanks,
Alex

p.s. By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.


TheSriram:
Does this change work in either cases? if prepack is defined or not defined?

mhart:
Lol, just ran into this – and fixed with it the same fix. Was just about to submit a PR.

It also happens if you have a `prepare` script defined.

This fix is the right fix – npm can output a bunch of stuff to stdout if it runs user scripts during the pack process, and aws-lambda-builders really messes up the tar file name in that case (and errors).

PR #117, opened 2019-05-31
fix: WindowsError keyword is undefined on non-Windows platforms

sanathkr:
WindowsError is a subclass of `OSError`. It is defined and raised only
in Windows. Hence the exception statement will return a
`WindowsError keyword is undefined` in other platforms.

EDIT: Usually this error is encountered when `CodeUri: /folder/does/not/exist`. The builder tries to copy the files and fails because the path does not exist. On Windows, the exception handler works, but on Mac/Linux Python interpreter does not find the variable `WindowsError` and crashes. So it leads to a vague `WindowsError` on Mac/Linux.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/aws_lambda_builders/utils.py
+++ b/aws_lambda_builders/utils.py
@@ -36,7 +36,7 @@ def copytree(source, destination, ignore=None):
         try:
             # Let's try to copy the directory metadata from source to destination
             shutil.copystat(source, destination)
-        except WindowsError as ex:  # pylint: disable=undefined-variable
+        except OSError as ex:  # pylint: disable=undefined-variable
             # Can't copy file access times in Windows
             LOG.debug("Unable to copy file access times from %s to %s", source, destination, exc_info=ex)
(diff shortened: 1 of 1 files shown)


aws_lambda_builders/utils.py
```diff
@@ -36,7 +36,7 @@ def copytree(source, destination, ignore=None):
         try:
             # Let's try to copy the directory metadata from source to destination
             shutil.copystat(source, destination)
-        except WindowsError as ex:  # pylint: disable=undefined-variable
+        except OSError as ex:  # pylint: disable=undefined-variable
```
jfuss: Do we still need the pytlint disable here?

TheSriram:
Can you add context on where you encountered this? will help for why this was fixed.

sanathkr:
- Removed pylint ignore
- Updated Description with more context on the error.

PR #118, opened 2019-06-06
fix: Better error message for binary validation failed

sanathkr:
*Issue #, if available:*

*Description of changes:*
https://github.com/aws/aws-toolkit-jetbrains/issues/828 - Doesn't fix the issue, but provides better error messages.

### DIFF
--- a/aws_lambda_builders/workflow.py
+++ b/aws_lambda_builders/workflow.py
@@ -61,7 +61,7 @@ def sanitize(func):
         if len(self.binaries) != len(valid_paths):
             raise WorkflowFailedError(workflow_name=self.NAME,
                                       action_name=None,
-                                      reason='Binary validation failed!')
+                                      reason='Unable to find a valid executable for {}'.format(self.runtime))
         func(self, *args, **kwargs)
     return wrapper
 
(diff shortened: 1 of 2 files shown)


aws_lambda_builders/workflow.py
```diff
@@ -61,7 +61,7 @@ def wrapper(self, *args, **kwargs):
         if len(self.binaries) != len(valid_paths):
             raise WorkflowFailedError(workflow_name=self.NAME,
                                       action_name=None,
-                                      reason='Binary validation failed!')
+                                      reason='Unable to find a valid executable for {}'.format(self.runtime))
```
TheSriram: This isnt necessarily the runtime, it could be the package manager, gradle, maven as well.

sanathkr: Hmm.. Is there a better message we can use?

TheSriram: well self.binaries could be used instead of self.runtime I think, it has total list of binaries that are needed to be verified.

jfuss: @sanathkr @TheSriram Did we ever land on what this should be? Would be great to wrap this up and get it into an upcoming release.

PR #127, opened 2019-08-21
Use Bespoke Tool Path for `dotnet-lambda` Commands

chrisoverzero:
There are situations in which the user's installation of
"Amazon.Lambda.Tools" can be as good as read-only. This installs the
builder's copy of the tool to a known location which is separate from
the user's default global installation.

Resolves: #126

*Issue #, if available:*

#126

*Description of changes:*

Adds the concept of a "tool directory" to the settings for dotnetcli. It's currently set to the value of the scratch directory. By doing this, the user's installation of "Amazon.Lambda.Tools" can be treated as read-only, whether it actually is or not.

### DIFF
--- a/aws_lambda_builders/workflows/dotnet_clipackage/actions.py
+++ b/aws_lambda_builders/workflows/dotnet_clipackage/actions.py
@@ -22,21 +22,22 @@ class GlobalToolInstallAction(BaseAction):
     DESCRIPTION = "Install or update the Amazon.Lambda.Tools .NET Core Global Tool."
     PURPOSE = Purpose.COMPILE_SOURCE
 
-    def __init__(self, subprocess_dotnet):
+    def __init__(self, subprocess_dotnet, tool_dir):
         super(GlobalToolInstallAction, self).__init__()
         self.subprocess_dotnet = subprocess_dotnet
+        self.tool_dir = tool_dir
 
     def execute(self):
         try:
             LOG.debug("Installing Amazon.Lambda.Tools Global Tool")
             self.subprocess_dotnet.run(
-                ['tool', 'install', '-g', 'Amazon.Lambda.Tools'],
+                ['tool', 'install', '--tool-path', self.tool_dir, 'Amazon.Lambda.Tools'],
             )
         except DotnetCLIExecutionError as ex:
             LOG.debug("Error installing probably due to already installed. Attempt to update to latest version.")
             try:
                 self.subprocess_dotnet.run(
-                    ['tool', 'update', '-g', 'Amazon.Lambda.Tools'],
+                    ['tool', 'update', '--tool-path', self.tool_dir, 'Amazon.Lambda.Tools'],
                 )
             except DotnetCLIExecutionError as ex:
                 raise ActionFailedError("Error configuring the Amazon.Lambda.Tools .NET Core Global Tool: " + str(ex))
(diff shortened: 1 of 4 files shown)


aws_lambda_builders/workflows/dotnet_clipackage/dotnetcli.py
```diff
@@ -47,9 +48,17 @@ def run(self, args, cwd=None):
 
         LOG.debug("executing dotnet: %s", invoke_dotnet)
 
+        if tool_dir is None:
+            env = None
+        else:
+            env = SubprocessDotnetCLI._merge(
```
chrisoverzero: I'm far from a Python expert, so if I've made a hash of this, please do let me know.

aws_lambda_builders/workflows/dotnet_clipackage/workflow.py
```diff
@@ -40,11 +40,13 @@ def __init__(self,
 
         options = kwargs["options"] if "options" in kwargs else {}
         subprocess_dotnetcli = SubprocessDotnetCLI(os_utils=OSUtils())
-        dotnetcli_install = GlobalToolInstallAction(subprocess_dotnet=subprocess_dotnetcli)
+        dotnetcli_install = GlobalToolInstallAction(subprocess_dotnet=subprocess_dotnetcli,
+                                                    tool_dir=scratch_dir)
```
chrisoverzero: I could find no more appropriate value, and I didn't feel right minting one.

sriram-mv:
you can use your own custom makefile builder for dotnet runtimes to add additional arguments to the binaries that are being used for building.

Here is an example for using makefile for custom runtimes, but a makefile builder can still be used for officially supported lambda runtimes.

chrisoverzero:
> you can use your own custom makefile builder […]

@sriram-mv I’m sure I can, but that doesn’t change how the builder outright fails on Docker images in which `dotnet-lambda` is already installed. The `sam build` process assumes that it _owns_ the installation of `dotnet-lambda` – even upgrading it if it’s out of date – when that’s commonly not so. 

Even if this approach isn’t the best one (and it felt like a kludge as I wrote it), that assumption of ownership is too fragile.

aws_lambda_builders/workflows/dotnet_clipackage/dotnetcli.py
CoshUS: We have dropped support for running Lambda builder in Python 2.
Let's directly use `env.copy().update(...)` instead.

aws_lambda_builders/workflows/dotnet_clipackage/actions.py
```diff
@@ -22,21 +22,22 @@ class GlobalToolInstallAction(BaseAction):
...
         super(GlobalToolInstallAction, self).__init__()
         self.subprocess_dotnet = subprocess_dotnet
+        self.tool_dir = tool_dir
 
     def execute(self):
         try:
             LOG.debug("Installing Amazon.Lambda.Tools Global Tool")
             self.subprocess_dotnet.run(
-                ['tool', 'install', '-g', 'Amazon.Lambda.Tools'],
+                ['tool', 'install', '--tool-path', self.tool_dir, 'Amazon.Lambda.Tools'],
```
CoshUS, line 34: This would still be a global tool but with a custom path.
> As a global tool in a custom location (also known as a tool-path tool).
>
>The tool binaries are installed in a location that you specify. You can invoke the tool from the installation directory or by providing the directory with the command name or by adding the directory to the PATH environment variable. One version of a tool is used for all directories on the machine.

Maybe we can use local tools instead and check whether a global Amazon.Lambda.Tools is already installed.
> As a local tool (applies to .NET Core SDK 3.0 and later).
>
>The tool binaries are installed in a default directory. You invoke the tool from the installation directory or any of its subdirectories. Different directories can use different versions of the same tool.

We should also check whether a global version of the Amazon.Lambda.Tools is already installed and skip over if needed.

CoshUS (changes requested):
Thanks for the PR! Added a few comments.
Using local tools should also fully resolve https://github.com/aws/aws-lambda-builders/issues/205

hawflau:
Closing due to inactivity. Feel free to reopen the PR if there's any update

PR #129, opened 2019-10-02
Change Go builders to match SAM CLI parameters and behavior

rts-rob:
*Issue #, if available:* N/A

*Description of changes:*

This change is required in order to enforce consistency with the parameters passed by SAM CLI to the other builders. After this PR is merged I will submit a separate PR to [awslabs/aws-sam-cli](https://github.com/awslabs/aws-sam-cli) that enables support for `sam build` for the `go1.x` runtime using either dep or modules.

Both Go builders expect an `options` dictionary to be passed in `**kwargs`; however, SAM CLI does not pass this dictionary. In order to maintain consistent behavior in SAM CLI, this PR updates the artifact name to match the source directory name. This behavior is consistent with the existing Makefile created by `sam init -r go1.x`.

This PR has been tested in Python 2.7, 3.6, and 3.7 with the proposed changes in SAM CLI to enable `sam build`.

### DIFF
--- a/aws_lambda_builders/workflows/go_dep/workflow.py
+++ b/aws_lambda_builders/workflows/go_dep/workflow.py
@@ -44,15 +44,12 @@ class GoDepWorkflow(BaseWorkflow):
                                             runtime=runtime,
                                             **kwargs)
 
-        options = kwargs["options"] if "options" in kwargs else {}
-        handler = options.get("artifact_executable_name", None)
-
         if osutils is None:
             osutils = OSUtils()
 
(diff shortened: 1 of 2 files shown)


jfuss:
@rts-rob For SAM CLI to support, there isn't any change that is required on this project. We do need to change SAM CLI though, which I have prototype out before. I can dig up the pr and my branch if needed.

These changes will break integration with SAM CLI, as we need the handler to name the binary `go build` produces.

rts-rob:
@jfuss - Thanks for picking this up so quickly! Understood; will move the modification over to aws-sam-cli.

PR #145, opened 2020-01-20
feat:Add Go Modules Debug Build Flags

awood45:
Required to do IDE debugging of Go functions.

### DIFF
--- a/aws_lambda_builders/workflows/go_modules/builder.py
+++ b/aws_lambda_builders/workflows/go_modules/builder.py
@@ -49,7 +49,7 @@ class GoModulesBuilder(object):
         cmd = [runtime_path, "build"]
         if self.mode and self.mode.lower() == BuildMode.DEBUG:
             LOG.debug("Debug build requested: Setting configuration to Debug")
-            cmd += ["-gcflags='-N -l'"]
+            cmd += ["-gcflags='all=-N -l'"]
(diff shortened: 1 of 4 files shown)


awood45:
My local functional tests (which should be unrelated to my change) are failing in a way that seems unrelated, so watching AppVeyor closely.

aws_lambda_builders/workflows/go_modules/builder.py
```diff
@@ -44,7 +46,11 @@ def build(self, source_dir_path, output_path):
         env.update(self.osutils.environ)
         env.update({"GOOS": "linux", "GOARCH": "amd64"})
         runtime_path = self.binaries[self.LANGUAGE].binary_path
-        cmd = [runtime_path, "build", "-o", output_path, source_dir_path]
+        cmd = [runtime_path, "build"]
+        if self.mode and self.mode.lower() == BuildMode.DEBUG:
```
c2tarun: Are we sure that BuildMode.DEBUG is always lower? I would suggest to make it BuildMode.DEBUG.lower() as well.

awood45: Following a pattern from the .NET builder, so the `aws_lambda_builders/workflow.py::BuildMode` enum values are contracted to be lower case, but input is not.

PR #146, opened 2020-01-21
fix: better error messages on build failures

sriram-mv:
Why is this change necessary?

* Generic "Binary Validation Failed" does not add enough value

How does it address the issue?

* Showcases which binary has failed and shows which paths were looked
at.

What side effects does this change have?

* None.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/aws_lambda_builders/workflow.py
+++ b/aws_lambda_builders/workflow.py
@@ -47,7 +47,10 @@ def sanitize(func):
         for binary, binary_path in binaries_copy.items():
             invalid_paths[binary] = []
             validator = binary_path.validator
-            exec_paths = binary_path.resolver.exec_paths if not binary_path.path_provided else binary_path.binary_path
+            try:
+                exec_paths = binary_path.resolver.exec_paths if not binary_path.path_provided else binary_path.binary_path
+            except ValueError as ex:
+                raise WorkflowFailedError(workflow_name=self.NAME, action_name="Resolver", reason=str(ex))
             for executable_path in exec_paths:
                 try:
                     valid_path = validator.validate(executable_path)
@@ -64,7 +67,7 @@ def sanitize(func):
             validation_failed_binaries = set(self.binaries.keys()).difference(valid_paths.keys())
             message = ""
             for validation_failed_binary in validation_failed_binaries:
-                message = "Binary validation failed for {}, Invalid paths : {}".format(
+                message = "Binary validation failed for {0}, searched for {0} in following locations  : {1} which did not satisfy constraints".format(
                     validation_failed_binary, invalid_paths[validation_failed_binary]
                 )
             raise WorkflowFailedError(workflow_name=self.NAME, action_name="Validation", reason=message)
--- a/aws_lambda_builders/workflow.py
+++ b/aws_lambda_builders/workflow.py
@@ -43,28 +43,27 @@ def sanitize(func):
         valid_paths = {}
         invalid_paths = {}
         # NOTE: we need to access binaries to get paths and resolvers, before validating.
-        binaries_copy = self.binaries
-        for binary, binary_path in binaries_copy.items():
+        for binary, binary_checker in self.binaries.items():
             invalid_paths[binary] = []
-            validator = binary_path.validator
             try:
                 exec_paths = (
-                    binary_path.resolver.exec_paths if not binary_path.path_provided else binary_path.binary_path
+                    binary_checker.resolver.exec_paths
+                    if not binary_checker.path_provided
+                    else binary_checker.binary_path
                 )
             except ValueError as ex:
                 raise WorkflowFailedError(workflow_name=self.NAME, action_name="Resolver", reason=str(ex))
             for executable_path in exec_paths:
                 try:
-                    valid_path = validator.validate(executable_path)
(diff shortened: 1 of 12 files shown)


sriram-mv:
Ran make black and that touched a bunch of files.

* Need to add extra tests

aws_lambda_builders/workflow.py
```diff
@@ -40,25 +40,34 @@ def sanitize(func):
...
         for binary, binary_path in binaries_copy.items():
+            invalid_paths[binary] = []
             validator = binary_path.validator
             exec_paths = binary_path.resolver.exec_paths if not binary_path.path_provided else binary_path.binary_path
             for executable_path in exec_paths:
-                valid_path = None
                 try:
                     valid_path = validator.validate(executable_path)
+                    if valid_path:
+                        valid_paths[binary] = valid_path
```
jfuss: Is `valid_paths` value of a key only a `str` and not a list, like `invalid_paths` because we consider only one path to be valid and end the search?

aws_lambda_builders/workflow.py
```diff
@@ -40,25 +40,34 @@ def sanitize(func):
...
+                if valid_paths.get(binary, None):
+                    binary_path.binary_path = valid_paths[binary]
                     break
         self.binaries = binaries_copy
         if len(self.binaries) != len(valid_paths):
-            raise WorkflowFailedError(workflow_name=self.NAME, action_name=None, reason="Binary validation failed!")
+            validation_failed_binaries = set(self.binaries.keys()).difference(valid_paths.keys())
+            message = ""
+            for validation_failed_binary in validation_failed_binaries:
+                message = "Binary validation failed for {}, Invalid paths : {}".format(
```
jfuss: Not sure this explains to customers what we want. To me, this says "the workflow needed binary X but it has all these invalid paths". But it is not clear to me what invalid paths are and why it matters, or even why we searched that path. Something like "We searched for binary X in these locations Y but did not find X that satisfied our constraints." We will still need a way to tell customers what the constraints are but it gets us closer to being descriptive to the customers so they can understand and fix the problem.

jfuss:
What issues is this solving? There seems to be two things here (pip not being installed and better error messages around searching for the binary). I am having trouble understand how this is solving the customer need of the second.

aws_lambda_builders/workflow.py
sriram-mv: Yes, if we encounter a valid path. we end the search.

aws_lambda_builders/workflow.py
sriram-mv: Fair enough, with the error message, we can be more explicit.

aws_lambda_builders/workflows/python_pip/compat.py
```diff
@@ -1,10 +1,16 @@
...
 from aws_lambda_builders.workflows.python_pip.utils import OSUtils
 
 
 def pip_import_string(python_exe):
     os_utils = OSUtils()
+    cmd = [python_exe, "-c", "import pip;"]
+    p = os_utils.popen(cmd, stdout=os_utils.pipe, stderr=os_utils.pipe)
+    _, _ = p.communicate()
+    if not p.returncode == 0:
+        raise MissingPipError(python_path=python_exe)
```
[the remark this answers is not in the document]
sriram-mv: Can be addressed.

sriram-mv:
" I am having trouble understand how this is solving the customer need of the second." - Yes, the verbiage of the error message can be improved, but it does convey more information than before.

tests/integration/workflows/python_pip/test_python_pip.py
```diff
@@ -64,7 +64,7 @@ def test_mismatch_runtime_python_project(self):
                 runtime=self.runtime_mismatch[self.runtime],
```
sanathkr: Is it possible to integration test this change?

aws_lambda_builders/workflow.py
```diff
@@ -40,25 +40,39 @@ def sanitize(func):
...
+                if valid_paths.get(binary, None):
+                    binary_checker.binary_path = valid_paths[binary]
                     break
-        self.binaries = binaries_copy
         if len(self.binaries) != len(valid_paths):
-            raise WorkflowFailedError(workflow_name=self.NAME, action_name=None, reason="Binary validation failed!")
+            validation_failed_binaries = set(self.binaries.keys()).difference(valid_paths.keys())
+            messages = []
+            for validation_failed_binary in validation_failed_binaries:
+                message = "Binary validation failed for {0}, searched for {0} in following locations  : {1} which did not satisfy constraints".format(
```
sanathkr: Good update. I think we can still make this better. Can we make the message more descriptive?

sriram-mv:
```
Building resource 'HelloWorldFunction'

Build Failed
Error: PythonPipBuilder:Validation - Binary validation failed for python, searched for python in following locations  : ['/Users/srirammv/workspace/python_versions/play/bin/python', '/usr/bin/python'] which did not satisfy constraints for runtime: python3.8. Do you have python for runtime: python3.8 on your PATH?
```

Does that look better?

PR #150, opened 2020-01-23
fix: npm output handling

sriram-mv:
Hi,

*Issue*
If `prepack` script is defined `npm` seems to output more than expected in original code.

*Description of changes:*
`npm pack` outputs filename as last line of stdout, while in code the whole its output was assigned to `tarfile_name` var. This fix makes it right.

Thanks,
Alex

p.s. By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.


sriram-mv:
Couldnt push to https://github.com/awslabs/aws-lambda-builders/pull/112

PR #153, opened 2020-01-30
Allow python_pip builder to install relative packages

### LINKED ISSUE #152: Allow python_pip builder to install relative packages
**Description:**
It is helpful for applications to sometimes structure themselves as multiple, separate packages. `sam build` currently can't require those local packages because the pip subprocess is run outside of the requirements.txt's directory.

**Steps to reproduce the issue:**
1. Init a sam project. Create a sub-package `mypackage` with its own setup.py.
2. Add `./mypackages` to your requriements.txt
3. Run `sam build`

**Observed result:**
Sam cannot find the local package `./mypackage`

Note: this does work currently when `sam build --use-container`

**Expected result:**
Sam would find, and install the local package `/.mypackage`, regardless of building in a container or not.

**Additional environment details (Ex: Windows, Mac, Amazon Linux etc)**

sbbowers:
Local requirements.txt references are relative to the process's cwd.
This changes the cwd during pip install, so `sam build` processes
relative to the location of requirements.txt, in the same way
`sam build --use-container` does.

Semantic differences between the purpose of requirements.txt and
setup.py:
* setup.py defines a project/libray and its specific dependancies;
supports additional project features: runtime version restriction,
extra_requires, entrypoints
* requirements.txt is a list of things to install, defining an
application's environment

This change provides support for installing packages using setup.py.
If your application is also a package, including "." in your
requirements.txt will defer dependancies resloution to your package's
setup.py.

This change also provides support for local sub-packages.  If your
application contains local sub-packages, listing them in the form of
"./<packagename>" in your requirements.txt will allow you to resolve
and install each package's dependancies.

Fixes #152

### DIFF
--- a/aws_lambda_builders/workflows/python_pip/packager.py
+++ b/aws_lambda_builders/workflows/python_pip/packager.py
@@ -7,6 +7,7 @@ import re
 import subprocess
 import logging
 from email.parser import FeedParser
+from os import chdir, getcwd, path
 
 
 from .compat import pip_import_string
@@ -328,9 +329,12 @@ class DependencyBuilder(object):
         # the dependency graph. Return the set of all package objects
         # which will serve as the master list of dependencies needed to deploy
         # successfully.
+        orig_cwd = getcwd()
+        chdir(path.dirname(requirements_filename))
         self._pip.download_all_dependencies(requirements_filename, directory)
         deps = {Package(directory, filename) for filename
                 in self._osutils.get_directory_contents(directory)}
+        chdir(orig_cwd)
         LOG.debug("Full dependency closure: %s", deps)
         return deps
 


sbbowers:
I believe the build failure is unrelated to the change. Please advise.

jfuss:
@sbbowers Thanks for the PR. The builds are unrelated to your change but I think appveyor has new patch versions of 3.6 and 2.7. We can handle this in a different pr.

I understand the reasoning behind this change but I am not sure its the right solution. This will actually change the behavior of the builder and break any customers who are currently relying on the resolving to happen from the CWD (this is how pip also works if I recall correctly). What you are running into is the differences on how `sam build` works in the container vs on your local file system. The big difference between `sam build --use-container` and `sam build` work is that in the container case, we only mount the `CodeUri` of a function. This means we we actually only every operator on the CodeUri and will never have access to any of the other modules. As you point out, this does have a side effect, in that the CWD is slightly different which causes a change in how dependencies are resolved in the `requirements.txt`. 

I think what we want to do is change how we do container type builds to match what we do locally. The team has talked about enabling locally dependencies in the container by mounting the CWD or where the `template.yaml` file is and then executing `sam build` inside the container instead of the `lambda-builders` cli. This we allow us to support the local dependency use-case and align the way we build local and in the container.

jfuss:
Closing due to no response. I think #250 solved this but need to verify. There is an open issue for this as well, so will handle the verification with the issue.

PR #161, opened 2020-03-21
Add the --unsafe-perm flag to to npm install

orf:
*Description of changes:*

NPM will not execute any script hooks if it is running as root, which is common in docker environments (i.e `sam build --use-container`). The `postinstall` script hook is commonly used by many libraries (especially ones involving native bindings), and it is highly confusing as to why hooks are not run when executing `sam build --use-container` versus `sam build` locally. 

Adding ﻿--unsafe-perm enables this.

### DIFF
--- a/aws_lambda_builders/workflows/nodejs_npm/actions.py
+++ b/aws_lambda_builders/workflows/nodejs_npm/actions.py
@@ -105,7 +105,7 @@ class NodejsNpmInstallAction(BaseAction):
             LOG.debug("NODEJS installing in: %s", self.artifacts_dir)
 
             self.subprocess_npm.run(
-                ["install", "-q", "--no-audit", "--no-save", "--production" "--unsafe-perm"], cwd=self.artifacts_dir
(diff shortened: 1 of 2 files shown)


aws_lambda_builders/workflows/nodejs_npm/actions.py
```diff
@@ -105,7 +105,7 @@ def execute(self):
             LOG.debug("NODEJS installing in: %s", self.artifacts_dir)
 
             self.subprocess_npm.run(
-                ["install", "-q", "--no-audit", "--no-save", "--production"], cwd=self.artifacts_dir
+                ["install", "-q", "--no-audit", "--no-save", "--production" "--unsafe-perm"], cwd=self.artifacts_dir
```
jfuss:
```suggestion
                ["install", "-q", "--no-audit", "--no-save", "--production", "--unsafe-perm"], cwd=self.artifacts_dir
```

sriram-mv:
closing since #163 is merged.

PR #163, opened 2020-04-02
Make sure NPM hooks are called

gitnik:
Continued from here since the original creator did not respond and only a small change was needed: https://github.com/awslabs/aws-lambda-builders/pull/161


gitnik:
Any news?

gitnik:
Glad to hear that :-)

PR #164, opened 2020-04-05
Add CGO_ENABLED=0 env in Windows

hottestchilipepper:
Following README.md

*Issue #, if available:*

*Description of changes:*

According to README.md of [aws-lambda-go](https://github.com/aws/aws-lambda-go) for developers on Windows, we should put like below
```
$env:GOOS = "linux"
$env:GOARCH = "amd64"
$env:CGO_ENABLED = "0"
go build -o main main.go
~\Go\Bin\build-lambda-zip.exe -o main.zip main
```
So I added $env:CGO_ENABLED = "0" code in this project


jfuss:
Is this actually needed? This locks customers into not using CGO. What happens when you actually want to set CGO_ENABLED=1?

I don't think we have enough understanding to merged/approve this. Why can't customers set this on their machine and control this themselves?

hottestchilipepper:
> Is this actually needed? This locks customers into not using CGO. What happens when you actually want to set CGO_ENABLED=1?
> 
> I don't think we have enough understanding to merged/approve this. Why can't customers set this on their machine and control this themselves?

According to 'https://github.com/aws/aws-lambda-go#for-developers-on-windows', we have to use CGO_ENABLED=0 on Windows.
Cgo lets Go packages call C code. So, I think it cannot be enabled on Windows because Lambda uses Linux machines. C library such as dll cannot run properly on Linux. Nobody may use CGO_ENABLED=1 on Windows. Because of this reason, The document which I show you says that we use CGO_ENABLED=0 on Windows.

jfuss:
Given how outdated this is, I am going to close.

I chatted with some of the owners of `aws-lambda-go` package. It sounds like some customers could still want `CGO_ENABLED=1` but that is unlikely or uncommon. We could cater to the common cases here but we still want a way to override the value. Maybe reading to see if the env var is set and not override. That is likely better so we aren't messing with customers setting this and then suddenly getting different behavior. 

@hottestchilipepper If this still impacts you and you are willing to submit a new PR that does the above, I would have happy to review.

PR #166, opened 2020-04-10
feat: Makefile based builder for provided runtimes

sriram-mv:
Why is this change necessary?

* provided runtimes are runtimes that are custom built with own
bootstraps etc, and its hard to determine which language it actually is.
A generic makefile based builder gives the control back to the user on
how the project needs to be built.

How does it address the issue?

* Provide a way to build for provided runtimes, which does not exist
today.


aws_lambda_builders/workflows/provided_make/DESIGN.md
```diff
@@ -0,0 +1,71 @@
...
+* Is `make` truly platform independent?
+  * Make comes with linux subsystem on windows and can also be installed with `choco`. 
+
+* Does this become a way to introduce plugins, makefile can have any commands in it?
+  * We only care about certain build targets. so essentially this is a pluggable builder, but nothing beyond that at this point in time.
+
+* Which environment variables are usable in this makefile?
+  * There are a series of whitelisted environment variables that need to be defined and not be overriden within the Makefile to work. Currently that is just `$ARTIFACTS_DIR`
+
+* Can this be used even for runtimes that have builders associated with it? eg: python3.8?
```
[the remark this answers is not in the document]
jfuss: @awood45 This is something that should be considered in SAM CLI not lambda builders. Lambda builders is a library of builders that someone can run, how SAM CLI picks which one is completely within SAM CLI.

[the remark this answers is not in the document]
sriram-mv: Yes, this would be choosing a workflow based on the manifest in SAM CLI. for eg: if a makefile is present and a build target is present, we choose the makefile workflow, else if we choose the bundled the workflow.

The logic to choose an appropriate workflow would need to be added there.

https://github.com/awslabs/aws-sam-cli/blob/develop/samcli/lib/build/workflow_config.py#L137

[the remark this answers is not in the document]
c2tarun: Even in sam-cli do we need to provide use of makefile for known workflows. I understand it is adding one more feature with some flexibility. But is it needed? What problem will this be solving? I think it'll add another case which then need lot of testing to make sure we didn't miss an edge case.

aws_lambda_builders/workflows/provided_make/actions.py
```diff
@@ -0,0 +1,74 @@
...
+        self.manifest_path = manifest_path
+        self.scratch_dir = scratch_dir
+        self.osutils = osutils
+        self.subprocess_make = subprocess_make
+        self.build_logical_id = build_logical_id
+
+    def _artifact_dir_path(self):
+        # This is required when running on windows to determine if we are running in linux
+        # subsystem or on native cmd or powershell.
+        return Path(self.artifacts_dir).as_posix() if self.osutils.which("sh") else self.artifacts_dir
```
c2tarun, line 52: Does this consider other types of shells? like zsh, csh etc? On that note, I think we should have special case for windows environment, which have rather low number of shells. Linux have lot of shells.

aws_lambda_builders/workflows/provided_make/make.py
```diff
@@ -0,0 +1,92 @@
...
+
+        if not args:
+            raise ValueError("requires at least one arg")
+
+        invoke_make = [self.make_exe] + args
+
+        LOG.debug("executing Make: %s", invoke_make)
+
+        p = self.osutils.popen(
+            invoke_make, stdout=self.osutils.pipe, stderr=self.osutils.pipe, cwd=cwd, env=env if env else {}
```
c2tarun, line 84: nit: why don't we default env to `{}` instead of `None`?

aws_lambda_builders/workflows/provided_make/actions.py
sriram-mv: this is a good point, we can check if we are on windows before we start looking for any shell.

aws_lambda_builders/workflows/provided_make/DESIGN.md
```diff
@@ -0,0 +1,71 @@
...
+Injected Environment Variables by Makefile Lambda Builder Workflow per Function to be built.
+
+``
+ARTIFACTS_DIR=/Users/noname/sam-app/.aws-sam/build/HelloWorldFunction
+``
+
+Sample Makefile:
+
+````
+build-HelloWorldFunction:
```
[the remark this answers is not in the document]
sriram-mv: this is how the design and implementation are designed at this point. Is there a reason you prefer it the other way around?

aws_lambda_builders/workflows/provided_make/DESIGN.md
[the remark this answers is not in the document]
sriram-mv: I would think it would be opting on a makefile workflow if necessary. Not neccesarily forced. if you are using a makefile based build approach, you would be responsible for the build instead.

aws_lambda_builders/workflows/provided_make/make.py
sriram-mv: sure,we can.

aws_lambda_builders/workflows/provided_make/make.py
```diff
@@ -0,0 +1,92 @@
...
+            raise ValueError("args must be a list")
+
+        if not args:
+            raise ValueError("requires at least one arg")
+
+        invoke_make = [self.make_exe] + args
+
+        LOG.debug("executing Make: %s", invoke_make)
+
+        p = self.osutils.popen(
```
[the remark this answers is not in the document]
sriram-mv: probably. but the osutils are desinged in a manner such that each is independent, so that each workflow can in essence be its own module. But open to considering it.

aws_lambda_builders/workflows/provided_make/DESIGN.md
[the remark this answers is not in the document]
c2tarun: Nope in line 21 it says `{Function_Logical_Id]-build` that is why I said.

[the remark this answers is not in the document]
sriram-mv: Good catch!

aws_lambda_builders/workflows/provided_make/make.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def run(self, args, env=None, cwd=None):
+
+        """
+        Runs the action.
+
+        :type args: list
+        :param args: Command line arguments to pass to Make
+
+        :type args: dict
```
jfuss, line 57: Should this be env?
```suggestion
        :type env: dict
```

aws_lambda_builders/workflows/provided_make/utils.py
```diff
@@ -0,0 +1,46 @@
...
+"""
+
+import os
+import platform
+import subprocess
+
+from aws_lambda_builders.utils import which
+
+
+class OSUtils(object):
```
jfuss, line 12: The famous Copy-paste OSUtils. We should make a generic one eventually

aws_lambda_builders/workflows/provided_make/utils.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def makedirs(self, path):
+        return os.makedirs(path)
+
+    def popen(self, command, stdout=None, stderr=None, env=None, cwd=None):
+        p = subprocess.Popen(command, stdout=stdout, stderr=stderr, env=env, cwd=cwd)
+        return p
+
+    def environ(self):
+        return os.environ
```
jfuss, line 30: Shouldn't this be a os.environ.copy()? That way we don't risk mutating customers env var by accident? Some workflows already do this (like Go Dep for example).

aws_lambda_builders/workflows/provided_make/workflow.py
```diff
@@ -0,0 +1,50 @@
...
+
+        super(ProvidedMakeWorkflow, self).__init__(
+            source_dir, artifacts_dir, scratch_dir, manifest_path, runtime=runtime, **kwargs
+        )
+
+        self.os_utils = OSUtils()
+
+        # Find the logical id of the function to be built.
+        options = kwargs.get("options") or {}
+        build_logical_id = options.get("build_logical_id", None)
```
jfuss, line 34: What happens when this is not provided?

Should we just error out here if it is a requirement of the builder?

tests/integration/workflows/provided_make/test_provided_make.py
```diff
@@ -0,0 +1,69 @@
...
+            "requests",
+            "requests-2.23.0.dist-info",
+        }
+
+        expected_files = self.test_data_files.union(dependencies_installed)
+        output_files = set(os.listdir(self.artifacts_dir))
+        self.assertEquals(expected_files, output_files)
+
+    def test_must_build_python_project_through_makefile_unknown_target(self):
+        with self.assertRaises(WorkflowFailedError):
```
jfuss, line 61: Should we have a way to make this more specific of an Error?

tests/unit/workflows/provided_make/test_actions.py
```diff
@@ -0,0 +1,67 @@
...
+from mock import patch, ANY
+
+from aws_lambda_builders.actions import ActionFailedError
+from aws_lambda_builders.workflows.provided_make.actions import ProvidedMakeAction
+from aws_lambda_builders.workflows.provided_make.make import MakeExecutionError
+
+
+class TestProvidedMakeAction(TestCase):
+    def setUp(self):
+        self.original_env = os.environ
```
jfuss, line 14: .copy()?

aws_lambda_builders/workflows/provided_make/DESIGN.md
```diff
@@ -0,0 +1,72 @@
...
+The Mechanism proposed here is that of `Makefile` Builder that has a particular make target available based on the logical id of the function to be built.
+The `Makefile` is the manifest and is present in the same path as the original source directory.
+
+
+### Implementation
+
+When the makefile is called through a lambda builder workflow, the appropriate target is triggered and artifacts should be copied to the exposed environmental variable `$ARTIFACTS_DIR`. This environment variable is seeded by the lambda builder with a value set to it. The targets are defined as 
+
+```
+build-{Function_Logical_Id}
```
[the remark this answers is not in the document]
sriram-mv: its easy to setup such a target by the user themselves and throw common stuff there. if there is a general target it becomes hard to do logical separation from building perspective, especially when multiple functions might share the same Makefile.

aws_lambda_builders/workflows/provided_make/actions.py
```diff
@@ -0,0 +1,79 @@
...
+import logging
+from pathlib import Path
+
+from aws_lambda_builders.actions import BaseAction, Purpose, ActionFailedError
+from .make import MakeExecutionError
+
+LOG = logging.getLogger(__name__)
+
+
+class ProvidedMakeAction(BaseAction):
```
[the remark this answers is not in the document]
sriram-mv: we can rename it to just custom. its more generic that way.

aws_lambda_builders/workflows/provided_make/utils.py
sriram-mv: yeah. we should.

aws_lambda_builders/workflows/provided_make/workflow.py
sriram-mv: yeah we would error out saying that the required make file target was not present.

PR #169, opened 2020-05-08
--ignore-failed-sources flag fix sam build error

fvilches17:
Please refer to this issue I created: https://github.com/awslabs/aws-sam-cli/issues/1970

Summary: when users have private feeds configured on their machines, the sam build command will fail because the dotnet cli tries to reference private feeds configured in the machine's global Nuget.Config

Adding the --ignore-failed-sources flag (e.g. dotnet tool install "Amazaon.Lambda.Tools" -g --ignore-failed-sources) fixes the issue.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/aws_lambda_builders/workflows/dotnet_clipackage/actions.py
+++ b/aws_lambda_builders/workflows/dotnet_clipackage/actions.py
@@ -30,11 +30,11 @@ class GlobalToolInstallAction(BaseAction):
     def execute(self):
         try:
             LOG.debug("Installing Amazon.Lambda.Tools Global Tool")
-            self.subprocess_dotnet.run(["tool", "install", "-g", "Amazon.Lambda.Tools"])
+            self.subprocess_dotnet.run(["tool", "install", "-g", "Amazon.Lambda.Tools", "--ignore-failed-sources"])
         except DotnetCLIExecutionError as ex:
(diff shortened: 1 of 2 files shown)


fvilches17:
hi @sriram-mv 

Based on the logs here I can't determine why the build failure for linux

https://ci.appveyor.com/project/AWSSAMCLI/aws-lambda-builders/builds/32804943

any more changes needed?

I believe the --ignore-failed-sources flags should work on Linux
https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-restore#global-packages-folder

sriram-mv:
you can use your own custom makefile builder for dotnet runtimes to add additional arguments to the binaries that are being used for building.

Here is an example for using makefile for custom runtimes, but a makefile builder can still be used for officially supported lambda runtimes.

sriram-mv:
However this is a good feature to have in the native builder as well, I will look into the build failures.

sriram-mv:
ah, looks like theres a formatting problem and the `black` binary is complaining. we use `black` for autoformatting of the codebase.

fvilches17:
Hi @sriram-mv 

Problem solved, just needed to apply formatting with black on one file.

Please re-check/re-approve/merge if all looking good : -)

fvilches17:
Hi @sriram-mv 

please see above

PR #178, opened 2020-06-17
feat(python): Support downloading manylinux2010 and manylinux2014

jfuss:
*Issue #, if available:*
N/A

*Description of changes:*
Some details are here: https://github.com/awslabs/aws-sam-cli/issues/2014#issuecomment-645508588

Previously for Python functions, we would only download manylinux1 whls. More details can be found in [PEP 513](https://www.python.org/dev/peps/pep-0513/), but manylinux1 whls support x86_64 linux platforms that have glibc 2.5 or greater. Over time, Python introduced 2 more manylinux versions (manylinux2010 (PEP 571)[https://www.python.org/dev/peps/pep-0571/] and manylinux2014 [PEP 599](https://www.python.org/dev/peps/pep-0599/). Looking at [manylinux2014 platform detection](https://www.python.org/dev/peps/pep-0599/#platform-detection-for-installers), a platform supports manylinux2014 if it is linux-x86_64 (among other but this is the one we care about) and has a glibc 2.17 or newer. As far as I know, AWS Lambda doesn't package a _manylinux module so I have skipped over that validation parts. Doing a quick check on AWS Lambda, all the glibc versions are 2.17 or greater and is an x86_64 platform. So AWS Lambda can support manylinux2014. 

Now looking at the [backwards compatible of manylinux2014](https://www.python.org/dev/peps/pep-0599/#backwards-compatibility-with-manylinux2010-wheels), PEP 599 states that "As a result, manylinux1 and manylinux2010 wheels are considered manylinux2014 wheels. A pip that recognizes the manylinux2014 platform tag will thus install manylinux2010 wheels for manylinux2014 platforms -- even when explicitly set -- when no manylinux2014 wheels are available." So if a platform is compatible with manylinux2014, it is also compatible with manylinux2010 and manylinux1. This allows us to also support manylinux2010 whls.


sriram-mv:
I read through the PEPs and the assumptions we make to include additional wheel formats which makes sense, looks like the CI is broken because of a version upgrade on the cerifi package. might require some tests being fixed.

PR #202, opened 2020-10-15
fix bug when using the default dependency pip builder

ItayGoren:
*Description of changes:*
When creating a default `PythonPipDependencyBuilder` for specific python version using the following:
`PythonPipDependencyBuilder('python3.8').build_dependencies('/tmp/foo/', '/path/python/requirements.txt', '/tmp/bar/')`
The following flow occurs:
```python
if dependency_builder is None:
    dependency_builder = DependencyBuilder(self.osutils, runtime)
```
This calls `DependencyBuilder.__init__(self, osutils, runtime, pip_runner=None)`. Notice that `pip_runner` is left `None`.
Then, inside the `__init__` method the following if is `True`:
```python
if pip_runner is None:
    pip_runner = PipRunner(python_exe=None, pip=SubprocessPip(osutils))
```
`SubprocessPip` constructor is defined by
```python
def __init__(self, osutils=None, python_exe=None, import_string=None):
```
And calls the following:
```python
self.python_exe = python_exe
if import_string is None:
    import_string = pip_import_string(python_exe=self.python_exe)
```
In this case, `python_exe` is `None` which pass an invalid parameter to `pip_import_string`.

The solution I submit pass the python's runtime value as the value to `SubprocessPip` inside `DependencyBuilder` constructor.


ohad-alike (approved):

hadas-alike (approved):

CoshUS:
@ItayGoren 
Thanks for the PR!
I couldn't find a code path in this repo that creates `PythonPipDependencyBuilder` without setting `python_exe` in `PipRunner`. Are you using `aws-lambda-builders` as a dependency?

ItayGoren:
The following code
```python
from aws_lambda_builders.workflows.python_pip.packager import PythonPipDependencyBuilder
import tempfile
d = tempfile.TemporaryDirectory()
PythonPipDependencyBuilder('python3.8').build_dependencies('.alike-build/lambda', d.name, 'python/requirements.txt')
```
Gets the following error:
```
  File "<string>", line 5, in <module>
  File "/opt/hostedtoolcache/Python/3.8.8/x64/lib/python3.8/site-packages/aws_lambda_builders/workflows/python_pip/packager.py", line 100, in __init__
    dependency_builder = DependencyBuilder(self.osutils, runtime)
  File "/opt/hostedtoolcache/Python/3.8.8/x64/lib/python3.8/site-packages/aws_lambda_builders/workflows/python_pip/packager.py", line 178, in __init__
    pip_runner = PipRunner(python_exe=None, pip=SubprocessPip(osutils))
  File "/opt/hostedtoolcache/Python/3.8.8/x64/lib/python3.8/site-packages/aws_lambda_builders/workflows/python_pip/packager.py", line 535, in __init__
    import_string = pip_import_string(python_exe=self.python_exe)
  File "/opt/hostedtoolcache/Python/3.8.8/x64/lib/python3.8/site-packages/aws_lambda_builders/workflows/python_pip/compat.py", line 10, in pip_import_string
    p = os_utils.popen(cmd, stdout=os_utils.pipe, stderr=os_utils.pipe, env=os_utils.original_environ())
  File "/opt/hostedtoolcache/Python/3.8.8/x64/lib/python3.8/site-packages/aws_lambda_builders/workflows/python_pip/utils.py", line 96, in popen
    p = subprocess.Popen(command, stdout=stdout, stderr=stderr, env=env)
  File "/opt/hostedtoolcache/Python/3.8.8/x64/lib/python3.8/subprocess.py", line 858, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/opt/hostedtoolcache/Python/3.8.8/x64/lib/python3.8/subprocess.py", line 1587, in _execute_child
    and os.path.dirname(executable)
  File "/opt/hostedtoolcache/Python/3.8.8/x64/lib/python3.8/posixpath.py", line 152, in dirname
    p = os.fspath(p)
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

sriram-mv:
This looks good, is there a unit test that could be added?

ItayGoren:
can this bug be solved? almost a year passed

jfuss:
@ItayGoren Is this still causing an issue? If so, can you please add tests? The change seems reasonable but without tests, it's hard for us to approve. We will want tests so we can guard against future regressions and make sure this solves the root issue. Integration and unit tests for this, would be the requirement.

mndeveci:
@ItayGoren I tried to extend unit tests in your branch to cover your implementation, but then I found out that passing `runtime` (string representation of lambda runtime enum) to `SubprocessPip` is not correct since it is expecting python executable path.

It will work if you have a python executable in your PATH environment variable with the same name as the lambda enum, but for other use cases, it will fail.

I would recommend changing the way you are initializing `PythonPipDependencyBuilder` by initalizing and passing `DependencyBuilder`.

Marking this PR as draft for now.

PR #206, opened 2020-10-28
Handle versioneer version strings to Python dependency from git repo

verdimrc:
*Issue #, if available:* N/A

*Description of changes:*

This PR allows `requirements.txt` to contain a dependency to a git repository of a Python package that uses `versioneer` to manage its version strings.

Currently, missing wheels are identified by comparing "package name & version strings" between `deps` and `compatible_wheels`, but `versioneer`-ed sdist will look like `packagename-x.y.z+0.<commit_id>.dirty(sdist)`.

The proposed change is to compare only the package names in `deps` and `compatible_wheels`.


verdimrc:
Closing this. Investigating a better approach.

Root cause is: `extract_zipfile()` didn't set executable flag, causing the local git repo to be in dirty state.

PR #207, opened 2020-10-28
Manage black version using requirement file

aahung:
*Issue #, if available:*

*Description of changes:*

Manage black in dev.txt, almost the same as https://github.com/aws/serverless-application-model/pull/1748

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -103,11 +103,6 @@ for:
       - sh: "sudo unzip -d /opt/gradle /tmp/gradle-*.zip"
       - sh: "PATH=/opt/gradle/gradle-5.5/bin:$PATH"
 
-      # Install black
-      - sh: "wget -O /tmp/black https://github.com/python/black/releases/download/19.10b0/black"
-      - sh: "chmod +x /tmp/black"
-      - sh: "/tmp/black --version"
-
     build_script:
       - "python -c \"import sys; print(sys.executable)\""
       - "LAMBDA_BUILDERS_DEV=1 pip install -e \".[dev]\""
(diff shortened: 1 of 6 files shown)


hoffa (approved):

.appveyor.yml
```diff
@@ -119,5 +114,5 @@ for:
       # Runs only in Linux
       - "LAMBDA_BUILDERS_DEV=1 pytest -vv tests/integration"
 
-      # Validate Code was formatted with Black
-      - "/tmp/black --check setup.py tests aws_lambda_builders"
+      # Validate Code was formatted with Black, black is only supported in Python 3
+      - if [[ $PYTHON_VERSION != "2"* ]]; then black --check setup.py tests aws_lambda_builders; fi
```
hawflau, line 118: is there any reason that you use `$PYTHON_VERSION != "2"*` but not `$PYTHON_VERSION == "3"*` or `$PYTHON_VERSION -gt "2"`?

aahung: Changed to `if [[ $PYTHON_VERSION > "3" ]]; then black --check setup.py tests aws_lambda_builders; fi`

hawflau (dismissed):

.appveyor.yml
hoffa: This might be pedantry, but the `>` operator [compares lexicographically](https://www.gnu.org/software/bash/manual/html_node/Bash-Conditional-Expressions.html); is that the intention?

aahung: either lexicographically or numerically should be able to specify "python 3 or higher" right?

aahung:
As discussed with @hoffa offline, with `> "3"` depends on the expected [lexicographic order](https://en.wikipedia.org/wiki/Lexicographic_order) to be used, and there might be a chance this is used instead, which makes `"2.7.14" > "3"`.

hawflau (approved):

hoffa (approved):

PR #210, opened 2020-11-12
Reduce generated binary size

cgarvis:
Adding the `ldflags` `-s -w` will reduce the size of the generated binary.

-s  disable symbol table
-w  disable DWARF generation

### DIFF
--- a/aws_lambda_builders/workflows/go_modules/builder.py
+++ b/aws_lambda_builders/workflows/go_modules/builder.py
@@ -50,6 +50,9 @@ class GoModulesBuilder(object):
         if self.mode and self.mode.lower() == BuildMode.DEBUG:
             LOG.debug("Debug build requested: Setting configuration to Debug")
             cmd += ["-gcflags='all=-N -l'"]
+        else:
+            cmd += ["-ldflags='-s -w'"]
+            
         cmd += ["-o", output_path, source_dir_path]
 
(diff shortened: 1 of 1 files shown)


aws_lambda_builders/workflows/go_modules/builder.py
```diff
@@ -50,6 +50,9 @@ def build(self, source_dir_path, output_path):
         if self.mode and self.mode.lower() == BuildMode.DEBUG:
             LOG.debug("Debug build requested: Setting configuration to Debug")
             cmd += ["-gcflags='all=-N -l'"]
+        else:
+            cmd += ["-ldflags='-s -w'"]
```
CoshUS, line 54: @cgarvis 
Thanks for the PR! 
Can you also update the related unit test `tests/unit/workflows/go_modules/test_builder.py::TestGoBuilder::test_run_executes_bundler_on_nixes`?

CoshUS (changes requested):

hoffa:
Doesn't this strip symbols from stack traces, making debugging much harder?

Also there's precedent to stripping Go binaries causing issues: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=717172

I'd vote against such a broad-brush approach unless the Go team explicitly recommends this as a supported approach.

jfuss:
Given https://github.com/aws/aws-lambda-builders/pull/210#issuecomment-800567284, I am going to close this as won't fix. If this is still something we should do, please open an issue so we can discuss this further.

PR #214, opened 2020-12-09
fix: run GlobalToolInstallAction in a lock block and only once

mndeveci:
https://github.com/aws/aws-lambda-builders/issues/211

*Description of changes:*
This change is adding a lock and an installation flag to GlobalToolInstallAction, to run it only once when it has run in a parallel build process.

### DIFF
--- a/aws_lambda_builders/workflows/dotnet_clipackage/actions.py
+++ b/aws_lambda_builders/workflows/dotnet_clipackage/actions.py
@@ -2,6 +2,7 @@
 Actions for .NET dependency resolution with CLI Package
 """
 
+import threading
 import os
 import logging
 
@@ -14,6 +15,8 @@ LOG = logging.getLogger(__name__)
 
 
 class GlobalToolInstallAction(BaseAction):
+    __lock = threading.Lock()
+    __tolls_installed = False
 
     """
     A Lambda Builder Action which installs the Amazon.Lambda.Tools .NET Core Global Tool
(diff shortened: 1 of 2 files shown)


tests/unit/workflows/dotnet_clipackage/test_actions.py
```diff
@@ -42,6 +45,24 @@ def test_global_tool_update_failed(self):
...
+            GlobalToolInstallAction(self.subprocess_dotnet),
+            GlobalToolInstallAction(self.subprocess_dotnet),
+        ]
+
+        for action in actions:
+            executor.submit(action.execute)
+
+        executor.shutdown()
+
+        self.subprocess_dotnet.assert_has_calls(
```
mgrandis, line 62: It seems we want to check that only 1 call to install was made.
You should maybe use this instead? 
https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called_once_with

tests/unit/workflows/dotnet_clipackage/test_actions.py
```diff
@@ -42,6 +45,24 @@ def test_global_tool_update_failed(self):
...
+            [call.run(["tool", "install", "-g", "Amazon.Lambda.Tools", "--ignore-failed-sources"])]
+        )
+
```
mgrandis, line 65: You can also maybe add a test where self.subprocess_dotnet raises a DotnetCLIExecutionError?

aws_lambda_builders/workflows/dotnet_clipackage/actions.py
```diff
@@ -28,15 +31,30 @@ def __init__(self, subprocess_dotnet):
...
             try:
-                self.subprocess_dotnet.run(["tool", "update", "-g", "Amazon.Lambda.Tools", "--ignore-failed-sources"])
+                LOG.debug("Installing Amazon.Lambda.Tools Global Tool")
+                self.subprocess_dotnet.run(["tool", "install", "-g", "Amazon.Lambda.Tools", "--ignore-failed-sources"])
+                GlobalToolInstallAction.__tools_installed = True
             except DotnetCLIExecutionError as ex:
-                raise ActionFailedError("Error configuring the Amazon.Lambda.Tools .NET Core Global Tool: " + str(ex))
+                LOG.debug("Error installing probably due to already installed. Attempt to update to latest version.")
+                try:
+                    self.subprocess_dotnet.run(
```
elbayaaa, line 50: Not sure whether it is a good practice to control the flow inside the exception handling or not. Is there away to check if it is already installed, then update otherwise, install?

tests/unit/workflows/dotnet_clipackage/test_actions.py
```diff
@@ -42,6 +45,24 @@ def test_global_tool_update_failed(self):
...
+        actions = [
+            GlobalToolInstallAction(self.subprocess_dotnet),
+            GlobalToolInstallAction(self.subprocess_dotnet),
+            GlobalToolInstallAction(self.subprocess_dotnet),
+        ]
+
+        for action in actions:
+            executor.submit(action.execute)
+
+        executor.shutdown()
```
elbayaaa, line 60: Will this wait till the spawned threads finishes execution? In java, the main thread need to "join" those threads first before releasing the pool.

mgrandis: https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor.shutdown  wait is True by default, so yes

tests/unit/workflows/dotnet_clipackage/test_actions.py
mndeveci: This is already been covered by other tests above.

mgrandis: Oops, missed those!

aws_lambda_builders/workflows/dotnet_clipackage/actions.py
mndeveci: You are right, but I also dotnet cli doesn't have an option to say install or update if it is already installed.

tests/unit/workflows/dotnet_clipackage/test_actions.py
mndeveci: Good catch, will update this

tests/unit/workflows/dotnet_clipackage/test_actions.py
mgrandis: You can avoid calling it though if you use `with ThreadPoolExecutor() as executor:`

mndeveci: Good one, will update to use context here

aws_lambda_builders/workflows/dotnet_clipackage/actions.py
mgrandis: Have tried calling `update` only? It seems that is uninstall then reinstalls the tool: https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet-tool-update

mndeveci: I tried to run `update` only after uninstalling the tool and it worked. But I want to be cautious about what we are changing here since this is been already used widely. Maybe older version (or in any other environments) of dotnet cli might not be working as it is working right now.

elbayaaa (approved):

mgrandis (approved):

PR #223, opened 2021-02-19
fix: Python Workflow with Incorrect Environ

CoshUS:
*Issue #, if available:*
https://github.com/aws/aws-lambda-builders/issues/224

*Description of changes:*
Patches the LB_LIBRARY_PATH envar and passes the original one to all popens in Python workflow.
This is required as PyInstaller changes LB_LIBRARY_PATH and affects its subprocesses.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -64,6 +64,7 @@ for:
     - "choco install bzr"
     - "choco install dep"
     - setx PATH "C:\go\bin;C:\gopath\bin;C:\Program Files (x86)\Bazaar\;C:\Program Files\Mercurial;%PATH%;"
+    - "refreshenv"
     - "go version"
     - "go env"
 
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -65,7 +65,6 @@ for:
     - "choco install dep"
     - setx PATH "C:\Program Files\Go;C:\gopath\bin;C:\Program Files (x86)\Bazaar\;C:\Program Files\Mercurial;%PATH%;"
     - setx GOROOT "C:\Program Files\Go"
-    - "refreshenv"
     - "go version"
     - "go env"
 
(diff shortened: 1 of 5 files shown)


.appveyor.yml
```diff
@@ -60,7 +60,7 @@ for:
 
     # setup go
     - rmdir c:\go /s /q
-    - "choco install golang"
+    - "choco install golang --version 1.15.7"
```
CoshUS, line 63: Pinning down golang version to 1.15 as 1.16 changes default Windows installation directory which breaks AppVeyor.

wchengru (dismissed):

aws_lambda_builders/workflows/python_pip/utils.py
```diff
@@ -16,6 +16,19 @@ class OSUtils(object):
     def environ(self):
         return os.environ
 
+    def original_environ(self):
+        # https://pyinstaller.readthedocs.io/en/stable/runtime-information.html#ld-library-path-libpath-considerations
+        env = dict(os.environ)
+        lp_key = "LD_LIBRARY_PATH"
+        original_lp = env.get(lp_key + "_ORIG")
```
sriram-mv, line 23: is this just specific to some systems where `LD_LIBRARY_PATH_ORIG` is populated? is there any way we can make this very specific to pyinstaller? does pyinstaller re-introduce any other env vars?

sriram-mv (changes requested):

aws_lambda_builders/workflows/python_pip/utils.py
CoshUS: LD_LIBRARY_PATH_ORIG is only populated if LD_LIBRARY_PATH is set. There are no other env vars that are generated by PyInstaller. I will add a check to run this only in Linux.

sriram-mv (dismissed):
Looks good, could you clean up the commit history?

CoshUS:
Cleaned up merge commits.

wchengru (approved):

sriram-mv (approved):

PR #245, opened 2021-06-15
feat: Allow Ruby runtime to build without requiring a manifest

mildaniel:
*Which issue(s) does this change fix?*
https://github.com/aws/aws-sam-cli/issues/784

*Why is this change necessary?*
Allow Ruby runtime to build without requiring a manifest file.

*How does it address the issue?*
Logs a warning and continues to build Ruby projects that are missing a Gemfile.

*What side effects does this change have?*

*Checklist:*

* [ ] Add input/output type hints to new functions/methods
* [ ] Write design document (Do I need to write a design document?) 
* [x] Write unit tests
* [x] Write/update functional tests
* [x] Write/update integration tests
* [ ] make pr passes
* [ ] make update-reproducible-reqs if dependencies were changed
* [ ] Write documentation

### DIFF
--- a/aws_lambda_builders/workflows/ruby_bundler/bundler.py
+++ b/aws_lambda_builders/workflows/ruby_bundler/bundler.py
@@ -51,7 +51,11 @@ class SubprocessBundler(object):
         out, _ = p.communicate()
 
         if p.returncode != 0:
-            # Bundler has relevant information in stdout, not stderr.
-            raise BundlerExecutionError(message=out.decode("utf8").strip())
+            if p.returncode == 10:
+                # Bundler error code 10 indicates `Gemfile not found`
+                LOG.warning("Gemfile not found. Continuing the build without dependencies.")
+            else:
+                # Bundler has relevant information in stdout, not stderr.
+                raise BundlerExecutionError(message=out.decode("utf8").strip())
 
         return out.decode("utf8").strip()
--- a/aws_lambda_builders/workflows/ruby_bundler/bundler.py
+++ b/aws_lambda_builders/workflows/ruby_bundler/bundler.py
@@ -51,12 +51,7 @@ class SubprocessBundler(object):
         out, _ = p.communicate()
 
         if p.returncode != 0:
-            # Log warning if attempting to build without a manifest.
-            if p.returncode == 10:
-                # Bundler error code 10 indicates `Gemfile not found`
-                LOG.warning("Gemfile not found. Continuing the build without dependencies.")
-            else:
-                # Bundler has relevant information in stdout, not stderr.
-                raise BundlerExecutionError(message=out.decode("utf8").strip())
+            # Bundler has relevant information in stdout, not stderr.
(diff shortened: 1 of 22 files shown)


wchengru (approved):

aws_lambda_builders/workflows/ruby_bundler/bundler.py
```diff
@@ -51,7 +51,16 @@ def run(self, args, cwd=None):
         out, _ = p.communicate()
 
         if p.returncode != 0:
-            # Bundler has relevant information in stdout, not stderr.
-            raise BundlerExecutionError(message=out.decode("utf8").strip())
+            if p.returncode == 10:
```
hoffa, line 54: Is this documented?

mildaniel: Although not explicitly documented by Bundler, the Bundler error codes can be seen [here](https://github.com/rubygems/bundler/blob/3f0638c6c8d340c2f2405ecb84eb3b39c433e36e/lib/bundler/errors.rb#L36).

hoffa: I see; please add a comment for reference to others as this is non-standard.

Tangentially, could this be handled in a more standard way (similar to https://github.com/aws/aws-lambda-builders/pull/243)? It seems Bundler always expects a `Gemfile` and isn't meaningful without it.

mgrandis: We could maybe put the `10` in a constant like `GEMFILE_NOT_FOUND` (this would also remove the need for the comment at line 55).

You could put your link to error codes on the line before the constant.

mildaniel: A discussion was had with the team regarding this change. This solution is non-standard because of the way Bundler behaves. Doing a similar approach to Python pip is dangerous because of Bundler's ability to search for a Gemfile in parent directories, meaning that it could break for users with non-standard project structures. Although unconventional, this approach was deemed safer.

I will go ahead and make the changes that Mathieu suggested to improve the clarity.

hoffa: Makes sense, thanks for the explanation. 👍

aws_lambda_builders/workflows/ruby_bundler/bundler.py
```diff
@@ -50,8 +50,22 @@ def run(self, args, cwd=None):
 
         out, _ = p.communicate()
 
+        """
+        Bundler error codes can be found here:
+        https://github.com/rubygems/bundler/blob/3f0638c6c8d340c2f2405ecb84eb3b39c433e36e/lib/bundler/errors.rb#L36
+        """
+        GEMFILE_NOT_FOUND = 10
```
mgrandis, line 57: You need to put this outside of the function :)

mgrandis (approved):

PR #248, opened 2021-06-21
fix integration test for mismatched python version

jpassaro:
Teach integration test to pass when the "wrong" version of python is
present but has no `pip` module.

By submitting this pull request, I confirm that my contribution is made
under the terms of the Apache 2.0 license.


CoshUS (approved):

mndeveci (approved):

tests/integration/workflows/python_pip/test_python_pip.py
```diff
@@ -1,5 +1,6 @@
 import os
 import shutil
+import six
```
hawflau, line 3: nit: Consider not using `six`

```suggestion
```

tests/integration/workflows/python_pip/test_python_pip.py
```diff
@@ -137,7 +138,15 @@ def test_mismatch_runtime_python_project(self):
...
-            self.assertIn("Binary validation failed", str(ex))
+            # handle both e.g. missing /usr/bin/python2.7 and situation where
+            # python2.7 does not have pip installed (as is the case in some
+            # Mac environments)
+            ex_s = str(ex)
+            if (
+                "Binary validation failed" not in ex_s
+                and "pip executable not found in your python environment" not in ex_s
+            ):
+                six.raise_from(AssertionError("Unexpected exception"), ex)
```
hawflau, line 149: nit: Consider not using `six`

 ```suggestion
                raise AssertionError("Unexpected exception") from ex
```

hawflau (approved):
nit: We should refrain from using `six` as we don't support Python2.7 anymore. `raise from` is natively supported in Python3. It's okay to still merge the PR if we can't apply the suggested changes. We can do it in another PR.

PR #249, opened 2021-06-21
Look for updated pip logging to detect directory links

jpassaro:
*Issue #, if available:*
Fixes #246 

*Background*
Per #246, the `pip` logging that aws-lambda-builders previously used to detect directory links (i.e. local packages) has been removed. It was replaced by another more detailed piece of logging that only appears when `pip download` is invoked with verbosity (`-v`).

Without this, the previous behavior of invoking `pip wheel` on the local links in question no longer happens. The upshot is that python projects with a `src/` setup would no longer be properly installed to the artifact directory.

*Description of changes:*
* Add an integration test ensuring that a python project with a `src/` setup is installed correctly.
* Update the regex used to scan pip output.
* Add verbosity to `pip download` arguments to ensure the output corresponding to the new regex will be logged.
* Enforce the regex's assumption of what the logging will look like by ensuring an up-to-date version of pip in the project `requirements`.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #250, opened 2021-06-24
fix: Pip not resolving local packages

mildaniel:
*Which issue(s) does this change fix?*
https://github.com/aws/aws-sam-cli/issues/2705

*Why is this change necessary?*
Fix local dependencies not being resolved by pip.

*How does it address the issue?*
Looks like [pip updated the output](https://github.com/pypa/pip/commit/b28e2c4928cc62d90b738a4613886fb1e2ad6a81#diff-5225c8e359020adb25dfc8c7a505950fd649c6c5775789c6f6517f7913f94542L529) that we try to string match, meaning we weren't catching local packages for newer versions of pip. 

This change adds multiple layers of regex matching to include both the old and new output format.

*What side effects does this change have?*

*Checklist:*

* [ ] Add input/output type hints to new functions/methods
* [ ] Write design document (Do I need to write a design document?) 
* [x] Write unit tests
* [ ] Write/update functional tests
* [x] Write/update integration tests
* [x] make pr passes
* [ ] make update-reproducible-reqs if dependencies were changed
* [ ] Write documentation

### DIFF
--- a/aws_lambda_builders/workflows/python_pip/packager.py
+++ b/aws_lambda_builders/workflows/python_pip/packager.py
@@ -596,7 +596,7 @@ class SubprocessPip(object):
 class PipRunner(object):
     """Wrapper around pip calls used by chalice."""
 
-    _LINK_IS_DIR_PATTERN = "Processing (.+?)\n  Link is a directory, ignoring download_dir"
+    _LINK_IS_DIR_PATTERNS = ["Processing (.+?)\n  Link is a directory, ignoring download_dir", "Processing (.+?)\n"]
 
     def __init__(self, python_exe, pip, osutils=None):
         if osutils is None:
@@ -644,10 +644,16 @@ class PipRunner(object):
                 package_name = match.group(1)
                 raise NoSuchPackageError(str(package_name))
             raise PackageDownloadError(error)
+
+        # Extract local packages from pip output.
+        # Iterate over possible pip outputs depending on pip version.
         stdout = out.decode()
-        matches = re.finditer(self._LINK_IS_DIR_PATTERN, stdout)
-        for match in matches:
-            wheel_package_path = str(match.group(1))
+        wheel_package_paths = set()
+        for pattern in self._LINK_IS_DIR_PATTERNS:
+            for match in re.finditer(pattern, stdout):
+                wheel_package_paths.add(str(match.group(1)))
+
+        for wheel_package_path in wheel_package_paths:
             # Looks odd we do not check on the error status of building the
             # wheel here. We can assume this is a valid package path since
             # we already passed the pip download stage. This stage would have
--- a/aws_lambda_builders/workflows/python_pip/packager.py
+++ b/aws_lambda_builders/workflows/python_pip/packager.py
@@ -596,6 +596,9 @@ class SubprocessPip(object):
 class PipRunner(object):
     """Wrapper around pip calls used by chalice."""
 
+    # Update regex pattern to correspond with the updated output from pip
+    # Specific commit:
+    # https://github.com/pypa/pip/commit/b28e2c4928cc62d90b738a4613886fb1e2ad6a81#diff-5225c8e359020adb25dfc8c7a505950fd649c6c5775789c6f6517f7913f94542L529
     _LINK_IS_DIR_PATTERNS = ["Processing (.+?)\n"]
 
     def __init__(self, python_exe, pip, osutils=None):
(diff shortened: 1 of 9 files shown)


aws_lambda_builders/workflows/python_pip/packager.py
```diff
@@ -596,7 +596,7 @@ def main(self, args, env_vars=None, shim=None):
 class PipRunner(object):
     """Wrapper around pip calls used by chalice."""
 
-    _LINK_IS_DIR_PATTERN = "Processing (.+?)\n  Link is a directory, ignoring download_dir"
+    _LINK_IS_DIR_PATTERNS = ["Processing (.+?)\n  Link is a directory, ignoring download_dir", "Processing (.+?)\n"]
```
wchengru, line 599: - You can add some docs on why the pattern is changed.
- Match the second one should always match the first. I believe you have already deduplicated this case, but you can simplify it.
- If we only match "Processing sth.", will we introduce non-dir links?

mildaniel: 1. I can definitely go ahead and add some docs.
2. It is true for this case, but in the future if the output is to change again, there's no guarantee that the new string will be a subset of the old one. Doing it this way should be more extensible and backwards compatible. It also prevents from having to re-write existing pip test stubs that explicitly write out ```Processing (.+?)\n  Link is a directory, ignoring download_dir```. However, if you think removing the ``` Link is a directory, ignoring download_dir``` pattern is more appropriate, I can certainly go ahead and do that.
3. This was my main concern. Searching through the pip code, though, I think this should be okay as it doesn't appear to log this type of pattern anywhere else. This could always change, but that's the nature of this type of implementation.

wchengru: For 2, You have done a good change on making it a list. You can just keep the second one since the first one is not necessary.

jfuss: Won’t that break customers using an older version of pip?

mildaniel: I don't think it would because the matched group will be the same in both cases, and anything matched by ```Processing (.+?)\n  Link is a directory, ignoring download_dir``` will also be matched by ```Processing (.+?)\n```. I just tested this and it seems to be fine. If there is backward compatibility concern, I can also include the old pattern.

jpassaro:
if this fix is preferred over #249, can I ask that you at least include the integration test case from that change?

mildaniel:
> if this fix is preferred over #249, can I ask that you at least include the integration test case from that change?

Hi @jpassaro. From a quick look, the integration test appears to be quite similar to the one in this pr. Is there something specific you are testing for in #249?

jpassaro:
yes, that setup tests specifically the scenario where requirements.txt has just one entry, `<sourcedir>`, which has a python project with a `src/` structure:

```
<sourcedir>/setup.py
<sourcedir>/setup.cfg
<sourcedir>/src/mycode/__init__.py
<sourcedir>/src/mycode/something.py
```

this was how I discovered the bug and I'd like to be sure that if a future change conflicts with this setup, the tests will catch it.

mildaniel:
I've updated the integration test to match file structure you defined. If there is a similar breaking change in the future, it will catch it.

jpassaro:
thanks, that meets my ask entirely. if it wasn't clear, I was suggesting to keep your test case (it seems useful to me) and add mine as an additional test; whether to do that or replace it as you've done is entirely up to y'all. Either way I really appreciate your attention.

tests/functional/workflows/python_pip/test_packager.py
```diff
@@ -519,7 +519,7 @@ def test_does_fail_on_invalid_local_package(self, tmpdir, osutils, pip_runner):
         pip, runner = pip_runner
         appdir, builder = self._make_appdir_and_dependency_builder(reqs, tmpdir, runner)
         requirements_file = os.path.join(appdir, "requirements.txt")
-        pip.set_return_tuple(0, (b"Processing ../foo\n" b"  Link is a directory," b" ignoring download_dir"), b"")
```
jfuss, line 522: Wonder if we should keep thins so we know we don't break back compat? Or make this output from pip parameterized to test both cases?

jfuss (changes requested):
Looks good to me. Just a question on one of the tests.

tests/functional/workflows/python_pip/test_packager.py
mildaniel: Reverted those tests to keep using the old output string. Updated the unit tests to test a mixture of reading both new and old.

wchengru (approved):

jfuss (approved):

jpassaro:
@wchengru @jfuss @mildaniel thanks for your work and attention on this. Any idea when this change will make it to a release of aws-sam-cli?

PR #259, opened 2021-08-01
fix(go version parts): remove alphabets from the version for validation

sriram-mv:
- Go versions like 1.12rc1 or 1.16beta1 are supported.
- Test added.

*Issue #, if available:*

https://github.com/aws/aws-lambda-builders/issues/189

*Description of changes:*


jfuss (approved):

aws_lambda_builders/workflows/go_modules/validator.py
```diff
@@ -47,7 +56,10 @@ def validate(self, runtime_path):
         if p.returncode == 0:
             out_parts = out.decode().split()
             if len(out_parts) >= 3:
-                version_parts = [int(x.replace("rc", "")) for x in out_parts[2].replace(self.LANGUAGE, "").split(".")]
+                version_parts = [
+                    int(GoRuntimeValidator.sanitize_go_version_part(version_part))
```
mgrandis, line 60: Could we use this regex `go(\d)\.(x|\d+)` to extract major/minor versions instead?
We could directly compare it to expected values.
We could also use it at line 50 if we put it into a `get_go_versions()` function, returning both values.

mgrandis (changes requested):
I think we can simplify this a little bit.

aws_lambda_builders/workflows/go_modules/validator.py
```diff
@@ -3,18 +3,20 @@
...
 from aws_lambda_builders.exceptions import MisMatchRuntimeError
 
 LOG = logging.getLogger(__name__)
 
 
 class GoRuntimeValidator(object):
-
     LANGUAGE = "go"
     SUPPORTED_RUNTIMES = {"go1.x"}
+    GO_VERSION_REGEX: Pattern[str] = re.compile("go(\\d)\\.(x|\\d+)")
```
mgrandis, line 19: I think you need to remove the type here, it is not supported on python2

mgrandis (approved):

Issue #229, opened 2021-04-06
Pythonpipbuilder: resolve dependencies - list index out of range


pabloiarriola:
**Description:**
I am trying to use sam build with the following requirements but its throwing the error: Pythonpipbuilder: resolve dependencies - list index out of range  

- pyotp
- ulid-py
- aws_encryption_sdk
- boto3
- requests
- attrs
- cryptography

**Steps to reproduce the issue:**
1. Create a virtual env.
1. Activate virtual env in a terminal
2. pip install -r requirements.txt
4. sam build 

**Observed result:**
Build Failed
Error: PythonPipBuilder:ResolveDependencies - list index out of range

**Expected result:**
Build Succeeded

**Additional environment details (Ex: Windows, Mac, Amazon Linux etc)**
Amazon Linux 2 Workspace 
Python3.8

Jasonca2:
I have a similar issue with different requirements. How are we supposed to use Sam with this underlying issue?

pabloiarriola:
@Jasonca2 thats pretty much the problem, that we cannot use SAM. Have you found any workarounds?

Jasonca2:
> @Jasonca2 thats pretty much the problem, that we cannot use SAM. Have you found any workarounds?

I have not. The issue is Sam, installed on Linux via there recommended installation method, brings in the aws-lambda-builders package as a requirement, not as part of the source code. The lambda-builders package would need to be patched and then aws-Sam updated to use the new version. I can't think of a workaround and would really like it if the Sam team could fix this.

sriram-mv:
@pabloiarriola @Jasonca2 what version of pip are you on? does sam build --use-container work?

Jasonca2:
> @pabloiarriola @Jasonca2 what version of pip are you on? does sam build --use-container work?

I'm using python3.7.10 and pip v20.1

I'm using the latest updated version of Sam as well.

jzavgok:
Any news on this? We are having the same issue :(

pabloiarriola:
> Any news on this? We are having the same issue :(

I got it working using sam build --use-container its pretty much the only way I found that works

pabloiarriola:
> > @pabloiarriola @Jasonca2 what version of pip are you on? does sam build --use-container work?
> 
> I'm using python3.7.10 and pip v20.1
> 
> I'm using the latest updated version of Sam as well.

the only difference with me is that im running python 3.8, the only workaround that worked was to use sam build --use-container

[...]

PR #265, opened 2021-08-24
fix: improve support for pep517 builds

sriram-mv:
* A pep517 build can declare build dependencies. Pip will then
know to install these dependencies before trying to build a wheel
file.

* When creating a build environment, it's only guaranteed to last
for the duration of the build process. It's not accessible once
a pip command finishes running.

* When we try to retrieve the version of a package we run a "modified"
form of "python setup.py egg_info".

* The problem with this is that we're not using the build environment
that has all the build dependencies installed (it's already gone),
so if setup.py imports a module (e.g. cython) because it expects
it to be there because it declared it as a build dependency
the egg_info command will fail.

* We don't check the RC or have a fallback case if we can't generate
egg info.

* We fail with an indecipherable IndexError.

We now have a fallback where if we can't import/run the setup.py file,
we assume the PKG-INFO file should be in the top level directory of the
sdist so we check if it's there, and if so we use that file.

*Issue #, if available:*

https://github.com/aws/aws-lambda-builders/issues/229

*Description of changes:*

This is a copy of https://github.com/aws/chalice/pull/1496 which ran into same issues as references in above issue.

### DIFF
--- a/aws_lambda_builders/workflows/python_pip/packager.py
+++ b/aws_lambda_builders/workflows/python_pip/packager.py
@@ -64,6 +64,14 @@ class PackageDownloadError(PackagerError):
     pass
 
 
+class UnsupportedPackageError(Exception):
+    """Unable to parse package metadata."""
+
+    def __init__(self, package_name):
+        # type: (str) -> None
+        super(UnsupportedPackageError, self).__init__("Unable to retrieve name/version for package: %s" % package_name)
+
+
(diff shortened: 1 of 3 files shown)


sriram-mv:
Created custom installer for linux and observed sam build worked for the same set of dependencies as mentioned in #229

wchengru (approved):

CoshUS (approved):

PR #267, opened 2021-08-26
Bugfix/awscli issue 2047 mvn copy correct dependencies

prower-turnitin:
#### Which issue(s) does this change fix?
https://github.com/aws/aws-sam-cli/issues/2047

#### Why is this change necessary?
SAM should package only dependencies required for runtime execution of the Lambda for Java.  SAM Templates which have  
'AWS::Serverless::LayerVersion' required by 'AWS::Serverless::Function' will need to compile layers first so those  
dependencies are available for compilation of the functions.  This updates SAM cli maven support to only package
the runtime objects required for execution. The maven workflow which creates both zip packages for Layers and
Functions need to be updated as the requirements for Layers packaging is different than the zip package allowed for 
Functions.  Example repo created by @gustahrodrigues who opened issue 2047 of samcli will now build and deploy both
layer and function.  Now the dependency provided by the layer won't be duplicated in the Lambda function deployment.
I made some updates to @gustahrodrigues example so some transitive dependencies wouldn't be included in the function
and only in the layer.  https://github.com/prower-turnitin/sam-app-with-layers  Build and deploy this SAM project with
this new version you can check the contents of zip files upload to s3 to see we now have minimal expected contexts,
when using layer and function.  

#### How does it address the issue?
Updates maven.py to only copy dependencies declared by the 'AWS::Serverless::Function' maven pom to be required for
runtime execution.  This is done selecting runtime as the included scope.  Updates order of building  
'AWS::Serverless::LayerVersion' and 'AWS::Serverless::Function' as the common dependency artifacts kept inside  
the layer artifact will be needed for compilation of the function which depends on artifacts in the layer.
It creates a Layers workflow which creates a zip package which follows the AWS Lambda Java Layers format. 
See https://docs.aws.amazon.com/lambda/latest/dg/configuration-layers.html.

#### What side effects does this change have?
I don't know of any.

#### Checklist

- [ ] Add input/output [type hints](https://docs.python.org/3/library/typing.html) to new functions/methods
- [ ] Write design document ([Do I need to write a design document?](https://github.com/aws/aws-sam-cli/blob/develop/DEVELOPMENT_GUIDE.md#design-document))
- [ ] Write unit tests
- [ ] Write/update functional tests
- [ ] Write/update integration tests
- [ ] `make pr` passes
- [ ] `make update-reproducible-reqs` if dependencies were changed
- [ ] Write documentation

By submitting this pull request, I confirm that my contribution is made under the terms of the [Apache 2.0 license](https://www.apache.org/licenses/LICENSE-2.0).


harreld (approved):

prower-turnitin:
@sanathkr @awood45 @jfuss @aahung @sriram-mv Hi, can one of you help me with getting this and related PR https://github.com/aws/aws-sam-cli/pull/3209 reviewed please?

aahung:
@prower-turnitin Thanks for creating the PR, we will prioritizing looking at it

prower-turnitin:
@mndeveci I updated this PR to fix conflicts and tests.  My changes are just dependency with sam-cli.  When I build sam-cli with or without my changes I get same test errors with `make pr`.  It seems develop has these test errors.  I haven't done much python development but would like to get this fix in for Java Maven development using sam-cli.  What else is needed for getting these changes accept?  I assume that somebody is working on the sam-cli develop test issues but I can merge in those changes so my sam-cli branch also runs successfully.

prower-turnitin:
@sanathkr @awood45 @jfuss @aahung @sriram-mv @mndeveci Any feedback on this PR?  Do you not understand problem this solves?  I can try to explain it better.

mndeveci:
@prower-turnitin thanks for all the work to put this changes together, we really appreciate that.

We discussed the proposed changes in this PR and we want to continue with following approach.

Dependencies which have `provided` scope should not be included in the final package. But since this is been used for a while, there might be some customers who is using this as default behavior. They might copy paste a dependency with `provided` scope (which should be a `runtime` or `compile` scope), and it is working for them now. If we change this behavior, there is a chance that these customers will be broken. Their built function will not be working anymore.

For not facing this problem, we are proposing to use this new behavior behind a flag, so that customers who wants to use this feature can start using it. We are also planning to print a warning message if a Java/Maven function contains `provided` dependency in the `pom.xml` file. We want these customers to run their build with this new flag, and validate that their function works as expected. We will also decide a future date to remove that flag, and set changes in this PR as default behavior.

We will be contacting with you for this PR and the one for SAM CLI to make the required changes in the following weeks.

Thanks!

emaayan:
> @prower-turnitin thanks for all the work to put this changes together, we really appreciate that.
> 
> We discussed the proposed changes in this PR and we want to continue with following approach.
> 
> Dependencies which have `provided` scope should not be included in the final package. But since this is been used for a while, there might be some customers who is using this as default behavior. They might copy paste a dependency with `provided` scope (which should be a `runtime` or `compile` scope), and it is working for them now. If we change this behavior, there is a chance that these customers will be broken. Their built function will not be working anymore.
> 
> For not facing this problem, we are proposing to use this new behavior behind a flag, so that customers who wants to use this feature can start using it. We are also planning to print a warning message if a Java/Maven function contains `provided` dependency in the `pom.xml` file. We want these customers to run their build with this new flag, and validate that their function works as expected. We will also decide a future date to remove that flag, and set changes in this PR as default behavior.
> 
> We will be contacting/working with you for this PR and the one for SAM CLI to make the required changes in the following weeks.
> 
> Thanks!

how are you going to handle local debug lambdas? when used with v2 sdk dependencies as they are , a local debug currently fails to run because it won't include http transitive dependencies so i'm forced to include those as well as compile, this will not happen for example when creating a deployment package using the assembly plugin because i've my maven dependencies plugin was configured correctly, but sam cli brakes my configuration by hardcoding the compile 
i think sam should respect the pom configuration in handling the deployment package consistently 

all this still btw still has the bigger issue if not handling multi-module maven projects which sam completely ignores which severely restricts the use of java 
for example , if i have a lambda function which shares a code module with another pom module in the same repo, sam will ignore that module as it only copies stuff under the current pom. 
if that pom also contains a reference to a parent pom the entire build will fail because sam would not be able to reference that parent.

prower-turnitin:
> @prower-turnitin thanks for all the work to put this changes together, we really appreciate that.
> 
> We discussed the proposed changes in this PR and we want to continue with following approach.
> 
> Dependencies which have `provided` scope should not be included in the final package. But since this is been used for a while, there might be some customers who is using this as default behavior. They might copy paste a dependency with `provided` scope (which should be a `runtime` or `compile` scope), and it is working for them now. If we change this behavior, there is a chance that these customers will be broken. Their built function will not be working anymore.
> 
> For not facing this problem, we are proposing to use this new behavior behind a flag, so that customers who wants to use this feature can start using it. We are also planning to print a warning message if a Java/Maven function contains `provided` dependency in the `pom.xml` file. We want these customers to run their build with this new flag, and validate that their function works as expected. We will also decide a future date to remove that flag, and set changes in this PR as default behavior.
> 
> We will be contacting/working with you for this PR and the one for SAM CLI to make the required changes in the following weeks.
> 
> Thanks!

@mndeveci I understand concern about this change impacting existing projects. I'd expect it to be pretty rare.  It would make sense to me to print a Warning for `provided` if the dependency doesn't match libs in a Layer being used by the Lambda.  If the runtime environment actually provides some dependencies then wouldn't want to give warning about those dependencies too.  That would be set of dependencies you should skip the warning on.  I'd expect it would be possible to know when to give a warning for `provided` in cases where all the Layers are in the same projects `template.yaml`.  Only in the case where the Layer was published outside a particular projects `template.yaml` would it be quite difficult to know if dependencies aren't actually present but not impossible.  One could download the layer and examine real artifacts with what is being declared as provided by project.  That's a bit next level and maybe too much to do.  In context of SAM it would only be the Layers and runtime dependencies which would use maven `provided`.  I'd think it would be prudent to skip the warning in cases where SAM cli can simply determine the dependency is being met by a Layer in the SAM project.

I do agree it makes sense to introduce this change in a flag at first to avoid problems.  It also maybe a good idea to give a warning for existing path too.  Let users know that the `provided` dependency which they should be expecting in Layer will be both in the Layer and in the Lambda. (In last example Lambda would have `provided` and Layer `compile`.) This could hint to the existence of this new flag so that they could avoid uploading dependencies redundantly.  This ultimately what I want to solve not uploading same dependency multiple times because it is declared as `provided` in the Lambda but SAM CLI due to this issue includes copies in both Lambda and Layer.

Happy to help.   

P.S. I haven't really looked into how Gradle behaves.  It has an equivalent `compileOnly` dependency which gradle users would expect to behave the same as `provided`.

prower-turnitin:
> > @prower-turnitin thanks for all the work to put this changes together, we really appreciate that.
> > We discussed the proposed changes in this PR and we want to continue with following approach.
> > Dependencies which have `provided` scope should not be included in the final package. But since this is been used for a while, there might be some customers who is using this as default behavior. They might copy paste a dependency with `provided` scope (which should be a `runtime` or `compile` scope), and it is working for them now. If we change this behavior, there is a chance that these customers will be broken. Their built function will not be working anymore.
> > For not facing this problem, we are proposing to use this new behavior behind a flag, so that customers who wants to use this feature can start using it. We are also planning to print a warning message if a Java/Maven function contains `provided` dependency in the `pom.xml` file. We want these customers to run their build with this new flag, and validate that their function works as expected. We will also decide a future date to remove that flag, and set changes in this PR as default behavior.
> > We will be contacting/working with you for this PR and the one for SAM CLI to make the required changes in the following weeks.
> > Thanks!
> 
> how are you going to handle local debug lambdas? when used with v2 sdk dependencies as they are , a local debug currently fails to run because it won't include http transitive dependencies so i'm forced to include those as well as compile, this will not happen for example when creating a deployment package using the assembly plugin because i've my maven dependencies plugin was configured correctly, but sam cli brakes my configuration by hardcoding the compile i think sam should respect the pom configuration in handling the deployment package consistently
> 
> all this still btw still has the bigger issue if not handling multi-module maven projects which sam completely ignores which severely restricts the use of java for example , if i have a lambda function which shares a code module with another pom module in the same repo, sam will ignore that module as it only copies stuff under the current pom. if that pom also contains a reference to a parent pom the entire build will fail because sam would not be able to reference that parent.


@emaayan I didn't have any trouble running `sam local start-api -d 5005`.  Try this example https://github.com/prower-turnitin/sam-app-with-layers/tree/useSdk2InExample. I added use of SDK 2 in both lambda and layer.

@emaayan I'm not trying to provide some kind of support for multi-module maven projects with SAM cli.  I'm not sure if those really make sense in the SAM context.  As one can mix runtimes in the projects `template.yaml`.  I've done this before mixing Java and Javascript.  From my understanding of SAM each Lambda `AWS::Serverless::Function` declares its runtime so SAM cli can build each based on the runtime type.  So each folder contains its own project to build Lambda or Layer artifact.  I get the typical expectation for Java maven project in one git repo would have just one pom.xml at the root, but that is not what SAM CLI is doing.  It builds each based on the runtime declared in the `template.yaml`.  I expect I could mix and match Maven and Gradle.  I could build a Layer with gradle and use that Layer with a Maven Lambda.  Not typical in one repo but very typical for both Gradle and Maven to Java projects to use dependencies built in separate project repositories.  So just keep in mind that SAM cli `sam build` is trying to do a `mvn package` or `gradle build` for each `AWS::Serverless::Function` and `AWS::Serverless:LayerVersion`with a java runtime.

You can partially use SAM CLI with Maven multi-module project if your content uri points to the exact jar being built.  In this case you have a `pom.xml` in root next to the `template.yaml`.  In this case you don't use `sam build` but `mvn package` at the root folder.  After that use `sam package --output-template-file packaged.yaml --s3-bucket <bucket> --s3-prefix <prefix>` and `sam deploy --template-file packaed.yaml --stack-name <stackname> --capabilities CAPABILITY_IAM`.  Local debug works too if you build artifacts before running.  This is kind of a hacky but it works.

This change is just to remove redundant dependencies in a Lamba when the dependency is scope `provided`.

emaayan:
probably because it uses shade plugin which i don't (as it may create
problems if you the same classes in several jars), i much prefer using the
assembly plugin which uses the zip deployment package including a lib
folder for dependencies, the shade package phase create an uber jar so only
that's get copied to .aws folder, but assembly plugin relies on copy
dependencies plugin which needs to bring runtime dependencies as well. but
because the compile flag is hardcoded into sam, those won't get copied.

when working with interlij  i cannot target content uri to a jar because it
always locks on to a directory

a mulit-maven project also makes sense if you have separate lambdas but you
wish all of them to inherit the same plugin versions, build process and
order (like unit test, test reports, package , signing etc..)
that's one of the main problems i have with sam (that and the fact that
it's tied to Cloud formation, where as in our team TF was decided to be
used)
for example we have a beanstalk spring boot project which uses lambda for
the sole reason to process a massive zip file that we don't want to
ovarload the memory of the beanstalk, that lamda could have been used as
submodule of the sptring-boot project, but it can't because the parent
module cannot be accessed

the other issue is that it overrides maven build process (which is why i
said it make sense of it to be a plugin ,like a beanstalk plugin i have for
maven,


another benefit is that they can all reference a shared module that can
contain things like abstract handlers, loggers, clients without using
nexus.

mndeveci:
Again, thanks a lot for your contribution @prower-turnitin 

Closing this in favor of https://github.com/aws/aws-lambda-builders/pull/318

In that PR, we are using same recommended fix for Java Maven build but we are putting it behind a feature flag so that all users can test it before switching to this new behavior. We also included fix for Java Gradle build for layers since it has the same issue (Java runtimes  does not support loading `.class` files from layers. Because of that all layer code [including business logic and the dependencies] should be bundled as `jar` files).

Let's continue discussion in that PR. And please let us know if you see any issues or concerns regarding to that implementation.

Thanks!

PR #272, opened 2021-09-01
feat/Node and Ruby Dependency Dir and Download Dependency

mingkun2020:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/aws_lambda_builders/workflows/nodejs_npm/workflow.py
+++ b/aws_lambda_builders/workflows/nodejs_npm/workflow.py
@@ -66,8 +66,8 @@ class NodejsNpmWorkflow(BaseWorkflow):
                 self.actions.append(CopySourceAction(self.dependencies_dir, artifacts_dir))
             else:
                 LOG.info(
-                    "download_dependencies is false and the dependencies_dir is not exists, just copying the source "
-                    "files into the artifacts directory "
+                    "download_dependencies is False and dependencies_dir is None. Copying the source files into the "
+                    "artifacts directory. "
                 )
 
         self.actions.append(NodejsNpmrcCleanUpAction(artifacts_dir, osutils=osutils))
--- a/aws_lambda_builders/workflows/nodejs_npm/workflow.py
+++ b/aws_lambda_builders/workflows/nodejs_npm/workflow.py
@@ -3,7 +3,7 @@ NodeJS NPM Workflow
 """
 from aws_lambda_builders.path_resolver import PathResolver
 from aws_lambda_builders.workflow import BaseWorkflow, Capability
-from aws_lambda_builders.actions import CopySourceAction
+from aws_lambda_builders.actions import CopySourceAction, CopyDependenciesAction
 from .actions import NodejsNpmPackAction, NodejsNpmInstallAction, NodejsNpmrcCopyAction, NodejsNpmrcCleanUpAction
 from .utils import OSUtils
 from .npm import SubprocessNpm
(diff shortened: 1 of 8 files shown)


aws_lambda_builders/workflows/nodejs_npm/workflow.py
```diff
@@ -40,18 +40,30 @@ def __init__(self, source_dir, artifacts_dir, scratch_dir, manifest_path, runtim
...
             npm_pack,
             npm_copy_npmrc,
             CopySourceAction(tar_package_dir, artifacts_dir, excludes=self.EXCLUDED_FILES),
-            npm_install,
-            NodejsNpmrcCleanUpAction(artifacts_dir, osutils=osutils),
         ]
 
+        if self.download_dependencies:
+            # installed the dependencies into artifact folder
+            npm_install = NodejsNpmInstallAction(artifacts_dir, subprocess_npm=subprocess_npm)
```
mndeveci, line 53: nit: `npm_install` variable not needed

aws_lambda_builders/workflows/nodejs_npm/workflow.py
```diff
@@ -40,18 +40,30 @@ def __init__(self, source_dir, artifacts_dir, scratch_dir, manifest_path, runtim
...
+            self.actions.append(npm_install)
+
+            # if dependencies folder exists, copy dependencies into dependencies into dependencies folder
+            if self.dependencies_dir:
+                self.actions.append(CopyDependenciesAction(source_dir, artifacts_dir, self.dependencies_dir))
+        else:
+            # if dependencies folder exists and not download dependencies, simply copy the dependencies from the
+            # dependencies folder to artifact folder
+            if self.dependencies_dir:
+                self.actions.append(CopySourceAction(self.dependencies_dir, artifacts_dir))
```
mndeveci, line 63: What happens if `self.dependencies_dir` is undefined? I think we need to flip the if statement and use;
`if not self.download_dependencies and self.dependencies_dir`

Because we should run regular build if `download_dependencies` is `False` or `dependencies_dir` is not defined.

aws_lambda_builders/workflows/ruby_bundler/workflow.py
```diff
@@ -31,12 +31,21 @@ def __init__(self, source_dir, artifacts_dir, scratch_dir, manifest_path, runtim
...
+            self.actions.append(bundle_deployment)
+
+            # if dependencies folder exists, copy dependencies into dependencies into dependencies folder
+            if self.dependencies_dir:
+                self.actions.append(CopyDependenciesAction(source_dir, artifacts_dir, self.dependencies_dir))
+        else:
+            # if dependencies folder exists and not download dependencies, simply copy the dependencies from the
+            # dependencies folder to artifact folder
+            if self.dependencies_dir:
+                self.actions.append(CopySourceAction(self.dependencies_dir, artifacts_dir))
```
mndeveci, line 51: What happens if `self.dependencies_dir` is undefined? See comment above

mndeveci (changes requested):
It is looking good in general, may need couple of small changes.

aws_lambda_builders/workflows/nodejs_npm/workflow.py
mndeveci: @CoshUS in your table for `False - None` row, shouldn't we fallback to `True - None` scenario or throw an exception? 

Because if SAM CLI sends `download_dependencies=False` with no `dependencies_dir`, that looks like an invalid call. We are basically saying no need to download dependencies (and the only reason could be that we already downloaded it and asking lambda builders to re-use that folder again) but we don't pass a dependencies folder.

CoshUS: For SAM CLI, `False - None` is certainly an invalid use case, but Lambda builders can be used as a library which makes sense for it to have logical actions based on each of the combinations.

Falling back to `True - None` seems to go against what the `download_dependencies` parameter mean.
I think throwing an exception is also fine as well if supporting `False - None` is non-trivial.

mingkun2020: Do we ever pass the `False - None` from the sam cli? The regular one should be `True - None` as @CoshUS mentioned.

CoshUS: No, `False - None` does not get used by SAM CLI.

mndeveci: My worry is, if we (or someone who is using this library) pass `False - None` by mistake and lambda-builders will say "the build is completed successfully", it feels like we are sweeping an invalid use case under the rug.

I agree that this could be another valid use case, where another cli or library wants to just build the source but don't want to pull any dependencies because there is no dependency definition. But if that is the case, then I would suggest to emit a log message saying we are skipping copying dependencies since dependencies folder are not defined.

CoshUS: I'm in favor of throwing an exception for this case.

mingkun2020: I think `download_dependencies=False` and `dependencies_dir=None` still make sense. Since don’t download dependencies and don’t copy files from `dependencies_dir` would means just keep the source files. So currently I emit a log message said `download_dependencies is false and the dependencies_dir is not exists, just copying the source files into the artifacts directory `. If customer has strong opinion to change it to throwing exception, we can do a quick change later.

mndeveci (approved):
LGTM. 

About our discussion, I am OK for both options (logging or throwing an exception). So feel free to change to raising an exception if you think that would be better.

aws_lambda_builders/workflows/nodejs_npm/workflow.py
```diff
@@ -40,18 +44,34 @@ def __init__(self, source_dir, artifacts_dir, scratch_dir, manifest_path, runtim
...
+            else:
+                LOG.info(
+                    "download_dependencies is false and the dependencies_dir is not exists, just copying the source "
```
CoshUS, line 69: nit: "download_dependencies is False and dependencies_dir is None. Copying the source files into the artifacts directory."

aws_lambda_builders/workflows/ruby_bundler/workflow.py
```diff
@@ -31,12 +34,26 @@ def __init__(self, source_dir, artifacts_dir, scratch_dir, manifest_path, runtim
...
+            else:
+                LOG.info(
+                    "download_dependencies is false and the dependencies_dir is not exists, just copying the source "
```
CoshUS, line 57: Same as above.

CoshUS (approved):
LGTM. 1 small comment on changing the wording a bit for the warning.

PR #301, opened 2021-10-07
[Auto Dependencies Layer] clean up dependencies action

mingkun2020:
add action for clean up the dependencies folder before installing/moving/copying

### DIFF
--- a/aws_lambda_builders/actions.py
+++ b/aws_lambda_builders/actions.py
@@ -40,6 +40,9 @@ class Purpose(object):
     # Action is compiling source code
     COMPILE_SOURCE = "COMPILE_SOURCE"
 
+    # Action is cleaning up the target folder
+    CLEAN_UP = "CLEAN_UP"
+
     @staticmethod
     def has_value(item):
         return item in Purpose.__dict__.values()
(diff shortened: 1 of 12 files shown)


aws_lambda_builders/actions.py
```diff
@@ -159,3 +162,26 @@ def execute(self):
             new_destination = os.path.join(self.dest_dir, name)
 
             shutil.move(dependencies_source, new_destination)
+
+
+class CleanUpAction(BaseAction):
```
CoshUS, line 167: I assume this will remove all the files in a directory but doesn't delete the directory itself right?
If so, can we add some comments to describe this behavior.

aws_lambda_builders/workflows/python_pip/workflow.py
```diff
@@ -83,6 +83,9 @@ def __init__(self, source_dir, artifacts_dir, scratch_dir, manifest_path, runtim
         if osutils.file_exists(manifest_path):
```
CoshUS, line 83: Let's use an early return to simplify this part.
```
if not osutils.file_exists(manifest_path):
  LOG.warning("requirements.txt file not found. Continuing the build without dependencies.")
  return [CopySourceAction(source_dir, artifacts_dir, excludes=self.EXCLUDED_FILES)]

...
```

aws_lambda_builders/actions.py
```diff
@@ -159,3 +162,26 @@ def execute(self):
...
+    PURPOSE = Purpose.CLEAN_UP
+
+    def __init__(self, target_dir):
+        self.target_dir = target_dir
+
+    def execute(self):
+        targets = os.listdir(self.target_dir)
+
+        for name in targets:
+            target_path = os.path.join(self.target_dir, name)
```
mndeveci, line 182: Maybe a little bit debug statement here that we are cleaning up this file/folder

aws_lambda_builders/workflows/java_gradle/workflow.py
```diff
@@ -11,6 +11,7 @@
 from .gradle import SubprocessGradle
 from .gradle_resolver import GradleResolver
 from .gradle_validator import GradleValidator
+from ...actions import CleanUpAction
```
mndeveci, line 14: nit: please use absolute path for imports

PR #318, opened 2022-01-21
feat: java build changes

mndeveci:
Created from original PR: https://github.com/aws/aws-lambda-builders/pull/267 (author: @prower-turnitin)

#### Which issue(s) does this change fix?

There are 2 issues with current way of building Java functions or layers;
- First one is related to dependency scope for Maven projects. Lambda Builders is currently using `compile` as included scope during build, which bundles dependencies with `provided` scope. This PR changes that to use `runtime` scope which will exclude `provided` in the final package.
- Second one is related to Layers. For Java runtimes, AWS Lambda Layer does only support using JAR files in the packaged artifact. Lambda Builders is currently bundles compiled `.class` files for the `.java` files in the project. This change uses generated JAR files for the final artifact so that it can be used in the layers.

aws/aws-sam-cli#2047

#### Why is this change necessary?

- Using `provided` scope for dependencies in Java Maven project is not working as expected, which blocks use cases such as using a dependency from referenced layer.
- Java files in layers are bundled incorrectly, which makes them unavailable during execution.

#### How does it address the issue?

- For Java Maven projects, we will be using `-DincludeScope=runtime` which will exclude dependencies with `provided` scope.
- For Layers, we will be bundling JAR files instead of individual CLASS files.

#### What side effects does this change have?

Any customer who accidentally uses `provided` scope instead of `compile` or `runtime` for a Maven project will be affected by this change. For that reason these changes are planned to be shipped behind a feature flag.


aws_lambda_builders/utils.py
```diff
@@ -29,17 +29,25 @@ def copytree(source, destination, ignore=None):
     :param ignore:
         A function that returns a set of file names to ignore, given a list of available file names. Similar to the
         ``ignore`` property of ``shutils.copytree`` method
+
+    :type include: function
```
CoshUS, line 33: nit: Callable type instead of function for easy of read
```suggestion
    :type include: Callable[[str], bool]
```

.pylintrc
```diff
@@ -317,7 +317,7 @@ max-args=5
 ignored-argument-names=_.*
 
 # Maximum number of locals for function / method body
-max-locals=15
+max-locals=20
```
CoshUS, line 320: Why do we need to change this rule globally?
Are we able to disable the rule as an special case for the offending block instead?

aws_lambda_builders/workflows/java/utils.py
```diff
@@ -37,17 +40,8 @@ def exists(self, p):
...
-            self.makedirs(destination)
-        names = self.listdir(source)
-        for name in names:
-            new_source = os.path.join(source, name)
-            new_destination = os.path.join(destination, name)
-            if os.path.isdir(new_source):
-                self.copytree(new_source, new_destination)
-            else:
-                self.copy(new_source, new_destination)
+    def copytree(self, source, destination, ignore=None, include=None):
```
CoshUS, line 43: Why do we need this copytree instead of using aws_lambda_builders.utils.copytree in the callers?

CoshUS (changes requested):
Thanks for the PR! A few comments.

aws_lambda_builders/workflows/java/utils.py
mndeveci: We used to have a custom method here to copy the files and folders. I've changed here to use the one in `aws_lambda_builders.utils.copytree`. I didn't want to change every place which uses this copy in order to keep the change in a meaningful scope. 
But I agree that we can do some code cleanup or refactoring some parts of lambda builders in the future. There are lots of duplication and multiple implementation of the same logic.

.pylintrc
mndeveci: Good callout, let me change this for individual functions rather than the whole project

CoshUS (approved):
LGTM! Thanks for addressing the comments.

moelasmar (approved):

PR #321, opened 2022-02-01
Fix python ARM build issue

eldritchideen:
*Issue #, if available:*
https://github.com/aws/aws-lambda-builders/issues/302

*Description of changes:*
This PR fixes the issue where SAM fails to package up a dependency on ARM where `pip` needs to build the wheel from source that contains a C extension. The wheel when build is named for `wrapt-1.13.3-cp39-cp39-linux_aarch64.whl` and this doesn't match the current list of compatible platforms for ARM. This has previously been reported with the `wrapt` library used by the XRay SDK. 

@heitorlessa @jfuss 


jfuss:
So we don't run arm tests :( but we need to do that. Appveyor doesn't support arm. I think we may need to move to codebuild (at least for arm) but that is out of scope for this. I am ok with this change knowing the arm tests are not running :(

@eldritchideen Did you validate this on arm locally some how?

eldritchideen:
> @eldritchideen Did you validate this on arm locally some how?

Yes, tested on an AL2 t4g instance.

jfuss (approved):

torresxb1:
@eldritchideen  Thank you for the contribution!

PR #322, opened 2022-02-03
feat: Esbuild Implementation Changes

mildaniel:
*Description of changes:*
- Change esbuild workflow logic to be done in a scratch dir
- Get the esbuild build properties from the options passed down to lambda builders instead of a package.json
- Add logic for handling simplicity defined entry points. e.g. sam-cli doesn't know if it's a JS or TS file, it would just pass down the handler
- Update DESIGN.md to reflect new changes and move it from the node workflow to the esbuild workflow


aws_lambda_builders/workflows/nodejs_npm_esbuild/actions.py
```diff
@@ -49,26 +53,26 @@ def execute(self):
...
-        entry_paths = [self.osutils.joinpath(self.source_dir, entry_point) for entry_point in entry_points]
+        entry_paths = [self.osutils.joinpath(self.scratch_dir, entry_point) for entry_point in entry_points]
 
         LOG.debug("NODEJS building %s using esbuild to %s", entry_paths, self.artifacts_dir)
 
-        for entry_point in entry_paths:
-            if not self.osutils.file_exists(entry_point):
-                raise ActionFailedError("entry point {} does not exist".format(entry_point))
+        explicit_entry_points = []
+        for entry_path, entry_point in list(zip(entry_paths, entry_points)):
```
hawflau, line 72: Isn't `zip(entry_paths, entry_points)` already iterable?

aws_lambda_builders/workflows/nodejs_npm_esbuild/actions.py
```diff
@@ -79,6 +83,20 @@ def execute(self):
         args.append("--target={}".format(target))
         args.append("--outdir={}".format(self.artifacts_dir))
         try:
-            self.subprocess_esbuild.run(args, cwd=self.source_dir)
+            self.subprocess_esbuild.run(args, cwd=self.scratch_dir)
         except EsbuildExecutionError as ex:
             raise ActionFailedError(str(ex))
+
+    def _get_explicit_file_type(self, entry_point, entry_path):
```
hawflau, line 90: Please add some docstring about what this method does

aws_lambda_builders/workflows/nodejs_npm_esbuild/actions.py
```diff
@@ -79,6 +83,20 @@ def execute(self):
...
+        if Path(entry_point).suffix:
+            if self.osutils.file_exists(entry_path):
+                return entry_point
+            raise ActionFailedError("entry point {} does not exist".format(entry_path))
+
+        if self.osutils.file_exists(get_typescript_file_type(entry_path)):
+            return get_typescript_file_type(entry_point)
+
+        if self.osutils.file_exists(get_javascript_file_type(entry_path)):
+            return get_javascript_file_type(entry_point)
```
hawflau, line 100: Do we need to do this refactoring and create `get_typescript_file_type` and `get_javascript_file_type`? I think this would be simpler if we don't need to use `get_typescript_file_type` and `get_javascript_file_type` elsewhere:
```suggestion
        for ext in [".ts", ".js"]:
            entry_path_with_ext = entry_path + ext
            if self.osutils.file_exists(entry_path_with_ext):
                return entry_path_with_ext
```

aws_lambda_builders/workflows/nodejs_npm_esbuild/workflow.py
```diff
@@ -48,27 +47,35 @@ def __init__(self, source_dir, artifacts_dir, scratch_dir, manifest_path, runtim
         if osutils is None:
             osutils = OSUtils()
 
+        subprocess_npm = SubprocessNpm(osutils)
+        subprocess_esbuild = self._get_esbuild_subprocess(subprocess_npm, scratch_dir, osutils)
```
hawflau, line 51: Is there a chance that `subprocess_esbuild` won't be used if the function exits early on L56?

aws_lambda_builders/workflows/nodejs_npm_esbuild/workflow.py
```diff
@@ -81,50 +88,46 @@ def actions_with_bundler(self, source_dir, artifacts_dir, bundler_config, osutil
...
-            install_action = NodejsNpmInstallAction(source_dir, subprocess_npm=subprocess_npm, is_production=False)
+            install_action = NodejsNpmInstallAction(scratch_dir, subprocess_npm=subprocess_npm, is_production=False)
 
-        esbuild_action = EsbuildBundleAction(source_dir, artifacts_dir, bundler_config, osutils, subprocess_esbuild)
-        return [install_action, esbuild_action]
+        esbuild_action = EsbuildBundleAction(scratch_dir, artifacts_dir, bundler_config, osutils, subprocess_esbuild)
+        return [copy_action, install_action, esbuild_action]
 
-    def get_manifest_config(self, osutils, manifest_path):
+    def get_manifest_config(self):
```
hawflau, line 110: Should this be renamed?

hawflau (changes requested):

aws_lambda_builders/workflows/nodejs_npm_esbuild/workflow.py
mildaniel: Yup, I moved this up intending to change the copy source action to a bundle action since we can now do that without a package.json. Will fix this up.

moelasmar (approved):

hawflau (approved):

PR #343, opened 2022-03-15
Esbuild --loader option support

chris-armstrong:
Adds support for esbuild `--loader` option

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #340, opened 2022-03-13
Add support for --external to esbuild configuration

chris-armstrong:
This PR adds support to the esbuild `BuildProperties` to specify a set of externals to use when building a source code bundle. This is useful when the package you require is on a layer and cannot be bundled with esbuild. It converts an array of `Externals` to a set of `esbuild` `--external:<module>` parameters.

For example, if your `BuildProperties` contains:

```yaml
  MyFunction:
    Metadata:
      BuildProperties:
        Externals: ['aws-sdk', 'puppeteer']
```

Then `--external:aws-sdk --external:puppeteer` will be added to the esbuild command line.

(Please forgive any non-idiomatic Python code - this is the first time writing Python for a number of years so I'm rusty on latest best practice).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #344, opened 2022-03-15
feat: (WIP) Cleanup and Generalize esbuild Builder

mildaniel:
Please do not merge, this is still a work in progress.
*Description of changes:*
- Use the esbuild API spec to create a generalized command builder that will make it easy to include new flags
- Remove the external node.js action, replace with adding an external flags with all dependencies in the package.json. This will make accelerate a few seconds faster, and remove Node.js as a dependency.
- Support for some new API flags such as external
- Cleans up some bloated functions


aws_lambda_builders/workflows/nodejs_npm_esbuild/esbuild.py
```diff
@@ -102,3 +93,141 @@ def run(self, args, cwd=None):
...
+    "minify",
+    "sourcemap",
+]
+
+# single value types (--target=es2020)
+SUPPORTED_ESBUILD_APIS_SINGLE_VALUE = [
+    "target",
+]
+
+# Multi-value types (--external:axios --external:aws-sdk)
```
chris-armstrong, line 110: There is a fourth possibility to consider here - key-value types, e.g. the `--loader:x=y` option:

```
  --loader:X=L          Use loader L to load file extension X, where L is
                        one of: js | jsx | ts | tsx | css | json | text |
                        base64 | file | dataurl | binary
```

which I've tried to support in #343 as dictionary in the bundler_config

aws_lambda_builders/workflows/nodejs_npm_esbuild/actions.py
```diff
@@ -62,138 +55,35 @@ def __init__(
...
-
-        entry_paths = [self.osutils.joinpath(self.scratch_dir, entry_point) for entry_point in entry_points]
-
-        LOG.debug("NODEJS building %s using esbuild to %s", entry_paths, self.artifacts_dir)
-
-        explicit_entry_points = []
-        for entry_path, entry_point in zip(entry_paths, entry_points):
-            explicit_entry_points.append(self._get_explicit_file_type(entry_point, entry_path))
+        # Handle the event of marking all dependencies as external
+        if self.skip_deps or "./node_modules/*" in self.bundler_config.get("external", []):
```
chris-armstrong, line 73: Does this cover cases where node_modules may be hoisted further up (such as in a yarn workspaces repository) e.g. `../../node_modules/*`?

chris-armstrong (commented):
I'm guessing these changes are to supercede what I've attempted to do in #343 and #340, but I've left a couple of comments for consideration

mildaniel:
Hey @chris-armstrong, really appreciate you taking the time to review and contribute. I started doing some cleanup work in this area to try and make the supporting new esbuild flags easier and more maintainable. I'd still like to accept your contributions since I'm not sure when I'll get around to finishing these changes but we definitely want support for these flags. 

Thanks again!

aws_lambda_builders/workflows/nodejs_npm_esbuild/esbuild.py
mildaniel: I would imagine that this format still falls under the multi-value types as described in the esbuild API. Users will be able to configure loaders in their template by adding it in the metadata BuildProperties like
```
BuildProperties:
   Loader:
    - x=y
```
If you think this is can be improved, please let me know.

chris-armstrong:
Appreciate it, thanks!

> Hey @chris-armstrong, really appreciate you taking the time to review and contribute. I started doing some cleanup work in this area to try and make the supporting new esbuild flags easier and more maintainable. I'd still like to accept your contributions since I'm not sure when I'll get around to finishing these changes but we definitely want support for these flags.
> 
> Thanks again!

aws_lambda_builders/workflows/nodejs_npm_esbuild/actions.py
mildaniel: Unfortunately it wouldn't, but right now lambda builders only installs the dependencies it finds in the manifest at the CodeURI root, so I don't imagine this would break anything. I image there's some area for improvement here though.

aws_lambda_builders/workflows/nodejs_npm_esbuild/esbuild.py
chris-armstrong: I tried to make it more idiomatic by structuring it as a dictionary, as from what I understand, this option wouldn't permit multiple loaders for the same file extension e.g.:

```yaml
  BuildProperties:
    Loader:
      .ts: ts
      .grpc: text
```

mildaniel: I completely agree that structuring it as a dict is likely more idiomatic. My concern is more around consistency of configuration to make it a better user experience. The more configuration methods we support, the more confusing it can get. I'd also like to mirror the esbuild API as closely as possible since I think that will make it easier to consume. I don't think we need to worry too much about validation as esbuild can validate its own APIs and return an appropriate error message.

chris-armstrong: Fair enough - It will probably make it less confusing and easier to document the different types of options (as opposed to each option itself) i.e. negating the need to mirror the documentation of esbuild too closely.

PR #351, opened 2022-04-06
fix: log results of make build

calabrla:
*Issue #, if available:*
N/A

*Description of changes:*
The Make builder does not log anything to stdout when it executes the make command. This change logs the output of the make command so developers can see the results of the execution. This follows the same pattern that exists in the `dotnetcli` builder.

### DIFF
--- a/aws_lambda_builders/workflows/custom_make/make.py
+++ b/aws_lambda_builders/workflows/custom_make/make.py
@@ -84,6 +84,9 @@ class SubProcessMake(object):
 
         out, err = p.communicate()
 
+        # Typically this type of information is logged to DEBUG, however, since the Make builder
+        # can be different per customer's use case, it is helpful to always log the output so
+        # developers can diagnose any issues.
         LOG.info(out.decode("utf8").strip())
 
         if p.returncode != 0:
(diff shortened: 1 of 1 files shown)


aws_lambda_builders/workflows/custom_make/make.py
```diff
@@ -84,6 +84,8 @@ def run(self, args, env=None, cwd=None):
 
         out, err = p.communicate()
 
+        LOG.info(out.decode("utf8").strip())
```
mndeveci, line 87: Should we also think about printing `err` output as well?

mndeveci (commented):
Thanks for the PR @calabrla! Added single comment there

aws_lambda_builders/workflows/custom_make/make.py
calabrla: Isn't it sufficient that the `err` is raised by the `MakeExecutionError` if the return code is not `0` in lines 89-90?

mildaniel (approved):

mildaniel:
nit: We normally log the outputs here at DEBUG level, where there are some exceptions like dotnet. I think for Makefile it makes sense to have as INFO due to the nature of it potentially being different for each customer's use case. Can we just leave a small comment noting why this is logged at INFO level to make it clear that this is the exception and not the rule.

calabrla:
@mildaniel I've added a comment explaining why we're logging to INFO.

mndeveci (approved):

Issue #341, opened 2022-03-14
[Announcement] Deprecating Go Dep Builder


jfuss:
As of now, AWS Lambda Builders supports building Go Functions with Dep: https://github.com/golang/dep. However, this project has been deprecated as of 2020.

### Am I impacted by this deprecation?

Following use cases will be impacted by this change;

If you are using aws-lambda-builders as a library and using the Dep Builder.

### What is deprecation schedule?

AWS Lambda Builders will mark the Dep builder as deprecated by March 14, 2022 and remove completely from AWS Lambda Builders module by April 11. 2022.

### Will this impact AWS SAM CLI?

No, AWS SAM CLI does not support Dep builders for Go Functions.

jfuss:
Released in [v1.15](https://github.com/aws/aws-lambda-builders/releases/tag/v1.15.0).

Closing

PR #354, opened 2022-04-20
chore(dep): Remove Dep builder

jfuss:
*Issue #, if available:*
#341

*Description of changes:*
Remove Dep Builder


mndeveci (approved):
+0 -907!

.appveyor.yml
```diff
@@ -100,14 +95,9 @@ for:
 
       # Install latest gradle
       - sh: "sudo apt-get -y remove gradle"
-      - sh: "sudo apt-get install go-dep"
```
mndeveci, line 103: With the version bump in SAM CLI, we might need to clean these as well in case we have some integration tests there.

jfuss: SAM CLI does not have Go Dep support, so there is no updates on SAM CLI at all.

mildaniel (approved):

PR #362, opened 2022-05-12
chore: remove redundant six usage

mndeveci:
Address issues from earlier PR: https://github.com/aws/aws-lambda-builders/pull/248#pullrequestreview-970151649


jfuss:
Can we remove six all together?

Would require removing the dep: https://github.com/aws/aws-lambda-builders/blob/develop/requirements/base.txt and from these spots as well: https://github.com/aws/aws-lambda-builders/search?q=six. Looks like the only usage is `six.string_types` which can be replaced by `str`.

mndeveci:
> Can we remove six all together?
> 
> Would require removing the dep: https://github.com/aws/aws-lambda-builders/blob/develop/requirements/base.txt and from these spots as well: https://github.com/aws/aws-lambda-builders/search?q=six. Looks like the only usage is `six.string_types` which can be replaced by `str`.

I followed this website to remove all `six` usage: https://python-future.org/compatible_idioms.html#metaclasses
I am not sure if we need to keep `object` as base class here;
`class BaseAction(object, metaclass=_ActionMetaClass)`

Let me know if that is redundant.

jfuss (approved):

jfuss:
> I am not sure if we need to keep object as base class here;
class BaseAction(object, metaclass=_ActionMetaClass)

I am not sure either and would rely on our tests for validation here. Without reading more deeply, this seems fine.

hawflau (approved):

PR #372, opened 2022-07-05
feat: Esbuild Generic Command Builder

mildaniel:
*Description of changes:*
These changes create a generic command builder to handle the three types of esbuild flags that are part of the esbuild CLI API. This PR also removes the workaround of calling a separate Node.js script to mark all dependencies as external, instead we go through the `package.json`  and mark all dependencies there as external individually.


aws_lambda_builders/workflows/nodejs_npm_esbuild/actions.py
```diff
@@ -49,168 +49,58 @@ def __init__(
...
-        if self.skip_deps:
-            LOG.info("Running custom esbuild using Node.js")
-            # Don't pass externals because the esbuild.js template makes everything external
-            script = EsbuildBundleAction._get_node_esbuild_template(
-                explicit_entry_points, target, self.artifacts_dir, minify, sourcemap
-            )
-            self._run_external_esbuild_in_nodejs(script)
-            return
+        if self._should_bundle_deps_externally():
+            esbuild_command.build_with_no_dependencies()
```
mildaniel, line 81: Instead of calling the external Node.js plugin to handle this, which would require us to update the Node.js template each time, we just make a list of all dependencies in the `package.json` file and set those to `external`.

aws_lambda_builders/workflows/nodejs_npm_esbuild/esbuild.py
```diff
@@ -102,3 +95,198 @@ def run(self, args, cwd=None):
...
+
+    def __init__(
+        self, scratch_dir: str, artifacts_dir: str, bundler_config: Dict[Any, Any], osutils: OSUtils, manifest: str
+    ):
+        self._scratch_dir = scratch_dir
+        self._artifacts_dir = artifacts_dir
+        self._bundler_config = bundler_config
+        self._osutils = osutils
+        self._manifest = manifest
+        self._command = []
```
mndeveci, line 130: nit: I assume this is `List[str]` should we add typing to this since it is empty list when we initialize this?

aws_lambda_builders/workflows/nodejs_npm_esbuild/esbuild.py
```diff
@@ -102,3 +95,198 @@ def run(self, args, cwd=None):
...
+        return self
+
+    def build_default_values(self) -> "EsbuildCommandBuilder":
+        """
+        Build the default values that each call to esbuild should contain
+
+        :rtype: EsbuildCommandBuilder
+        :return: An instance of the command builder
+        """
+        args = ["--bundle", "--platform=node", "--format=cjs", "--outdir={}".format(self._artifacts_dir)]
```
mndeveci, line 193: Are we planning to parameterize the `platform` and `format` flags as well?

aws_lambda_builders/workflows/nodejs_npm_esbuild/esbuild.py
```diff
@@ -102,3 +95,198 @@ def run(self, args, cwd=None):
...
+        """
+        Get a list of all the multi-value flag types (e.g. --external:aws-sdk)
+
+        :rtype: List[str]
+        :return: Arguments to be appended to the command list
+        """
+        args = []
+        for param in SUPPORTED_ESBUILD_APIS_MULTI_VALUE:
+            if param in self._bundler_config:
+                vals = self._bundler_config.get(param)
```
mndeveci, line 260: nit: let's name this as `values`

aws_lambda_builders/workflows/nodejs_npm_esbuild/esbuild.py
mildaniel: I've been debating this myself. I'm not sure when/if platform would change, but since Lambda now allows for ES modules so format could be configurable.

mildaniel: I parameterized `format` as well.

mndeveci (approved):

aws_lambda_builders/workflows/nodejs_npm_esbuild/actions.py
```diff
@@ -49,168 +49,58 @@ def __init__(
...
+            if EXTERNAL_KEY in self._bundler_config:
+                # Already marking everything as external,
+                # shouldn't attempt to do it again when building args from config
+                self._bundler_config.pop(EXTERNAL_KEY)
+
+        args = (
```
mildaniel, line 87: Do we need to check for duplicate keys in the command before running?

mildaniel: Confirmed this shouldn't occur.

PR #375, opened 2022-07-14
feat: Changed Sourcemap default value to false

lucashuy:
Related to AWS SAM CLI change: https://github.com/aws/aws-sam-cli/pull/4062

*Description of changes:*
This change changes source map's default value from `true` to `false`.

### DIFF
--- a/aws_lambda_builders/workflows/nodejs_npm_esbuild/DESIGN.md
+++ b/aws_lambda_builders/workflows/nodejs_npm_esbuild/DESIGN.md
@@ -128,7 +128,7 @@ Users might want to tweak some of these runtime arguments for a specific project
 * `minify`: boolean, defaulting to `true`
 * `sourcemap`: boolean, defaulting to `false`
 
-Here is an example that deactivates minification and source maps, and supports JavaScript features compatible with Node.js version 10.
+Here is an example that deactivates minification, enables source maps, and supports JavaScript features compatible with Node.js version 10.
 
 ```json
 {
--- a/aws_lambda_builders/workflows/nodejs_npm_esbuild/DESIGN.md
+++ b/aws_lambda_builders/workflows/nodejs_npm_esbuild/DESIGN.md
@@ -117,17 +117,16 @@ testing flow before invoking `sam build`. For additional typescript caveats with
 
 #### Configuring the bundler
 
-The Lambda builder invokes `esbuild` with sensible defaults that will work for the majority of cases. Importantly, the following three parameters are set by default
+The Lambda builder invokes `esbuild` with sensible defaults that will work for the majority of cases. Importantly, the following parameters are set by default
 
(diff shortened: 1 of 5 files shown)


mildaniel (approved):
LGTM! But we also have integration tests inside of SAM-CLI. Make sure you change those accordingly as well when making these types of changes.

aws_lambda_builders/workflows/nodejs_npm_esbuild/DESIGN.md
```diff
@@ -117,24 +117,23 @@ testing flow before invoking `sam build`. For additional typescript caveats with
 
 #### Configuring the bundler
 
-The Lambda builder invokes `esbuild` with sensible defaults that will work for the majority of cases. Importantly, the following three parameters are set by default
+The Lambda builder invokes `esbuild` with sensible defaults that will work for the majority of cases. Importantly, the following parameters are set by default
 
 * `--minify`, as it [produces a smaller runtime package](https://esbuild.github.io/api/#minify)
-* `--sourcemap`, as it generates a [source map that allows for correct stack trace reporting](https://esbuild.github.io/api/#sourcemap) in case of errors (see the [Error reporting](#error-reporting) section above)
```
sriram-mv, line 123: does removing a default setting cause any breaking changes? what happens to an app which did not bother about setting any defaults, do we see a difference in the artifacts being built?

https://github.com/aws/aws-sam-cli/pull/4062 indicates that enabling source map becomes an either or instead of both.

lucashuy: On it's own, setting the source map option (or previously with it being `True` by default) would only generate the source map file. To actually use it, customers would have to also provide `NODE_OPTIONS: --enable-source-maps`. With this change to default `False`, it would technically cause a breaking change without the PR in SAM CLI. Lambda builders would no longer include source map file by default.

You're right on with the PR in SAM CLI, that PR is the other side to this one where the customer only has to provide either or. Once both these changes are released and leave the develop branch, customers shouldn't see any changes to their existing setups.

sriram-mv: Ok thanks, so these are closely related and need to be released together.

sriram-mv: Since this could be a considering a breaking change with just aws lambda builders in mind (if someone were to use it directly as a binary instead of via aws-sam-cli), we should think of upping the protocol version? though the protocol version depends on the arguments passed in to the workflow.

sriram-mv (approved):
I'm approving with the caveat that this is experimental, we should have a mechanism in place such that we do not make such kind of breaking changes as there could be other consumers of aws lambda builders apart from aws sam cli.

PR #378, opened 2022-08-04
feat: add custom working directory for custom build workflow

moelasmar:
*Description of changes:*
Update the Custom builder workflow to accept a working directory within the options directory, to allow the custom builder to execute Makefile that exist outside the Lambda resources source code, or that run some building logic on a project level instead of lambda resource code path level.

### DIFF
--- a/aws_lambda_builders/workflows/custom_make/actions.py
+++ b/aws_lambda_builders/workflows/custom_make/actions.py
@@ -52,7 +52,8 @@ class CustomMakeAction(BaseAction):
         :param build_logical_id: the lambda resource logical id that will be built by the custom action.
 
         :type working_directory: str
-        :param working_directory: path to the working directory where the Makefile will be executed.
+        :param working_directory: path to the working directory where the Makefile will be executed. Use the scratch_dir
+        as the working directory if the input working_directory is None
         """
         super(CustomMakeAction, self).__init__()
         self.artifacts_dir = artifacts_dir
--- a/aws_lambda_builders/workflows/custom_make/actions.py
+++ b/aws_lambda_builders/workflows/custom_make/actions.py
@@ -22,7 +22,16 @@ class CustomMakeAction(BaseAction):
     DESCRIPTION = "Running build target on Makefile"
     PURPOSE = Purpose.COMPILE_SOURCE
 
-    def __init__(self, artifacts_dir, scratch_dir, manifest_path, osutils, subprocess_make, build_logical_id):
+    def __init__(
+        self,
+        artifacts_dir,
+        scratch_dir,
+        manifest_path,
+        osutils,
+        subprocess_make,
+        build_logical_id,
+        working_directory=None,
+    ):
         """
         :type artifacts_dir: str
         :param artifacts_dir: directory where artifacts needs to be stored.
@@ -38,6 +47,12 @@ class CustomMakeAction(BaseAction):
 
         :type subprocess_make aws_lambda_builders.workflows.custom_make.make.SubprocessMake
         :param subprocess_make: An instance of the Make process wrapper
+
+        :type build_logical_id: str
+        :param build_logical_id: the lambda resource logical id that will be built by the custom action.
+
+        :type working_directory: str
+        :param working_directory: path to the working directory where the Makefile will be executed.
         """
         super(CustomMakeAction, self).__init__()
         self.artifacts_dir = artifacts_dir
@@ -46,6 +61,7 @@ class CustomMakeAction(BaseAction):
         self.osutils = osutils
         self.subprocess_make = subprocess_make
         self.build_logical_id = build_logical_id
+        self.working_directory = working_directory if working_directory else scratch_dir
 
     @property
     def artifact_dir_path(self):
@@ -91,7 +107,7 @@ class CustomMakeAction(BaseAction):
                     "build-{logical_id}".format(logical_id=self.build_logical_id),
                 ],
                 env=current_env,
-                cwd=self.scratch_dir,
+                cwd=self.working_directory,
             )
         except MakeExecutionError as ex:
             raise ActionFailedError(str(ex))
--- a/aws_lambda_builders/workflows/custom_make/utils.py
+++ b/aws_lambda_builders/workflows/custom_make/utils.py
@@ -35,15 +35,6 @@ class OSUtils(object):
     def abspath(self, path):
         return os.path.abspath(path)
 
-    def isabspath(self, path):
-        return os.path.isabs(path)
-
-    def relpath(self, path, start):
-        return os.path.relpath(path, start)
-
-    def join_relpath(self, path, start):
-        return os.path.join(start, path)
-
     @property
     def pipe(self):
         return subprocess.PIPE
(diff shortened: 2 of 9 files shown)


aws_lambda_builders/workflows/custom_make/actions.py
```diff
@@ -46,6 +61,7 @@ def __init__(self, artifacts_dir, scratch_dir, manifest_path, osutils, subproces
         self.osutils = osutils
         self.subprocess_make = subprocess_make
         self.build_logical_id = build_logical_id
+        self.working_directory = working_directory if working_directory else scratch_dir
```
sriram-mv, line 64: Why this implied behavior? we should document that in the docstring as well.

aws_lambda_builders/workflows/custom_make/utils.py
```diff
@@ -35,6 +35,15 @@ def normpath(self, path):
     def abspath(self, path):
         return os.path.abspath(path)
 
+    def isabspath(self, path):
```
sriram-mv, line 38: we really need to start bringing together all the utils on path, there are quite a few floating around in this repository.

aws_lambda_builders/workflows/custom_make/workflow.py
```diff
@@ -34,6 +34,21 @@ def __init__(self, source_dir, artifacts_dir, scratch_dir, manifest_path, runtim
...
+        else:
+            if self.os_utils.isabspath(working_directory):
+                # Working directory should be relative to the source_dir, so when we copy the source_dir to scratch_dir
+                # the working directory value is still valid
+                working_directory = self.os_utils.relpath(working_directory, source_dir)
+
+            # join the working directory to the scratch_dir
+            working_directory = self.os_utils.normpath(self.os_utils.join_relpath(working_directory, scratch_dir))
+
+        if working_directory is None:
```
sriram-mv, line 50: why the double if?

aws_lambda_builders/workflows/custom_make/utils.py
```diff
@@ -35,6 +35,15 @@ def normpath(self, path):
...
+        return os.path.isabs(path)
+
+    def relpath(self, path, start):
+        return os.path.relpath(path, start)
+
+    def join_relpath(self, path, start):
```
sriram-mv, line 44: do we need this join_relpath? we seem to be just joining paths and not relative paths.

aws_lambda_builders/workflows/custom_make/workflow.py
```diff
@@ -34,6 +34,21 @@ def __init__(self, source_dir, artifacts_dir, scratch_dir, manifest_path, runtim
...
+            working_directory = scratch_dir
+        else:
+            if self.os_utils.isabspath(working_directory):
+                # Working directory should be relative to the source_dir, so when we copy the source_dir to scratch_dir
+                # the working directory value is still valid
+                working_directory = self.os_utils.relpath(working_directory, source_dir)
+
+            # join the working directory to the scratch_dir
+            working_directory = self.os_utils.normpath(self.os_utils.join_relpath(working_directory, scratch_dir))
```
sriram-mv, line 48: is working directory going to be a sub dir within scratch dir all the time? then is this more of a naming? its not really a directory that can exist by itself.

sriram-mv (changes requested):

aws_lambda_builders/workflows/custom_make/utils.py
moelasmar: no it can be just join_path

aws_lambda_builders/workflows/custom_make/workflow.py
moelasmar: it should be always a sub directory within source_code path, to be copied to the scratch_dir, and then we can run parallel build.

aws_lambda_builders/workflows/custom_make/workflow.py
moelasmar: I forgot to remove it

moelasmar:
I updated the PR to remove the working_directory path calculation and to make it relative to scratch_dir, as I think this information is important for SAM CLI, but not to the builders. Customers can use the builders to run against any working directory.

But in SAM CLI we need to make sure that the working directory is a sub-directory of the source code to make sure that when we copy the source code to the scratch directory, we do not cause any changes to the customer's source code.

aws_lambda_builders/workflows/custom_make/actions.py
moelasmar: This is the current default behaviour. Custom builder run against the scratch_dir, so if there is no working_directory passed by the customers, we should use the current default behaviour.  I updated the docstring.

aws_lambda_builders/workflows/custom_make/utils.py
moelasmar: removed these utilities as it will not be used as per our discussion related to use the passed working_directory as it is, and to move the path calculation to sam cli.

aws_lambda_builders/workflows/custom_make/workflow.py
moelasmar: it was by mistake, any way I removed all this logic, we do not need to do the path calculation here, and will do it in SAM CLI.

aws_lambda_builders/workflows/custom_make/utils.py
moelasmar: removed.

aws_lambda_builders/workflows/custom_make/workflow.py
moelasmar: removed

tests/integration/workflows/custom_make/test_custom_make.py
```diff
@@ -57,6 +57,48 @@ def test_must_build_python_project_through_makefile(self):
         output_files = set(os.listdir(self.artifacts_dir))
         self.assertEqual(expected_files, output_files)
 
+    def test_build_python_project_failed_through_makefile_without_working_directory(self):
+        source_code = os.path.join(os.path.dirname(__file__), "testdata", "makefile-in-different-working-directory")
+        manifest_path_valid = os.path.join(source_code, "Makefile")
+        with self.assertRaises(WorkflowFailedError):
```
sriram-mv, line 63: why does this test fail and raise a WorkflowFailedError? working_directory is an optional param right?

sriram-mv (changes requested):

sriram-mv (approved):

tests/integration/workflows/custom_make/test_custom_make.py
torresxb1: +1

torresxb1: ah I think it's because the requirements file is inside `source_code`, not in `makefile-in-different-working-directory`

torresxb1 (approved):

tests/integration/workflows/custom_make/test_custom_make.py
moelasmar: Yes, I had a call with Sriram to explain that but forgot to reply to this comment

PR #379, opened 2022-08-09
feat: Install esbuild deps with production flag

mildaniel:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/aws_lambda_builders/workflows/nodejs_npm/workflow.py
+++ b/aws_lambda_builders/workflows/nodejs_npm/workflow.py
@@ -145,7 +145,7 @@ class NodejsNpmWorkflow(BaseWorkflow):
         return [PathResolver(runtime=self.runtime, binary="npm")]
 
     @staticmethod
-    def get_install_action(source_dir, artifacts_dir, subprocess_npm, osutils, build_options, is_production=True):
+    def get_install_action(source_dir, artifacts_dir, subprocess_npm, osutils, build_options):
         """
         Get the install action used to install dependencies at artifacts_dir
 
@@ -180,4 +180,4 @@ class NodejsNpmWorkflow(BaseWorkflow):
         if (osutils.file_exists(lockfile_path) or osutils.file_exists(shrinkwrap_path)) and npm_ci_option:
             return NodejsNpmCIAction(artifacts_dir, subprocess_npm=subprocess_npm)
 
-        return NodejsNpmInstallAction(artifacts_dir, subprocess_npm=subprocess_npm, is_production=is_production)
+        return NodejsNpmInstallAction(artifacts_dir, subprocess_npm=subprocess_npm, is_production=True)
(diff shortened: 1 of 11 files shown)


aws_lambda_builders/workflows/nodejs_npm_esbuild/workflow.py
```diff
@@ -188,7 +192,7 @@ def _accelerate_workflow_actions(
                     actions += esbuild_no_deps
             else:
                 # Invalid workflow, can't have no dependency dir and no installation
-                raise EsbuildExecutionError(message="Lambda Builders encountered and invalid workflow")
+                raise EsbuildExecutionError(message="Lambda Builders encountered an invalid workflow")
```
jfuss, line 195: Should this error be more specific? The comment has more context which could be useful to help customers understand why the build failed. As a customer "invalid workflow" doesn't really tell me much.

aws_lambda_builders/workflows/nodejs_npm_esbuild/workflow.py
```diff
@@ -127,7 +127,11 @@ def actions_with_bundler(
         )
 
         install_action = NodejsNpmWorkflow.get_install_action(
-            source_dir, scratch_dir, subprocess_npm, osutils, self.options, is_production=False
+            source_dir,
+            scratch_dir,
+            subprocess_npm,
+            osutils,
+            self.options,
```
jfuss, line 134: I assume `is_production` defaults to `True`?

Is there any reason a customer may want `False`? If so, how would a customer do that?

jfuss (changes requested):

aws_lambda_builders/workflows/nodejs_npm_esbuild/workflow.py
mildaniel: Yes, it defaults to true. We don't support any way for customers to set it to False, and we actually don't use it as False anywhere except here so I'm just going to remove that parameter. Setting `production=False` means that development dependencies will also be installed which adds a ton of bloat and install time that customers likely won't want packaged with their Lambda. Before, we expected customers to include `esbuild` as a dev dependency. Now, we expect it to be installed on their PATH or as a normal dependency. This is the behaviour we currently have with our Node.js workflow and are making esbuild consistent.

jfuss (approved):

aws_lambda_builders/workflows/nodejs_npm/workflow.py
```diff
@@ -180,4 +180,4 @@ def get_install_action(source_dir, artifacts_dir, subprocess_npm, osutils, build
         if (osutils.file_exists(lockfile_path) or osutils.file_exists(shrinkwrap_path)) and npm_ci_option:
             return NodejsNpmCIAction(artifacts_dir, subprocess_npm=subprocess_npm)
 
-        return NodejsNpmInstallAction(artifacts_dir, subprocess_npm=subprocess_npm, is_production=is_production)
+        return NodejsNpmInstallAction(artifacts_dir, subprocess_npm=subprocess_npm, is_production=True)
```
mildaniel, line 183: We can just remove this parameter.

lucashuy (approved):

PR #389, opened 2022-10-06
Add `-trimpath` to GoModulesBuilder

scrofungulus:
This PR addresses the issue described [here](https://github.com/aws/aws-sam-cli/issues/3258) in the AWS SAM CLI [repository](https://github.com/aws/aws-sam-cli). 

*Description of changes:*
Adds a `trim_go_path` option to the `GoModulesBuilder`. Directly from the `go help build` command from the Go CLI: 
```
-trimpath
                remove all file system paths from the resulting executable.
		Instead of absolute file system paths, the recorded file names
		will begin either a module path@version (when using modules),
		or a plain import path (when using the standard library, or GOPATH).
```

Ultimately, this addresses an issue that occurs when using `sam build`. Without `-trimpath` the compiled binaries computed hashes are different each time forcing CloudFormation to redeploy a Lambda function without code changes. 

### DIFF
--- a/tests/integration/workflows/go_modules/test_go.py
+++ b/tests/integration/workflows/go_modules/test_go.py
@@ -112,6 +112,19 @@ class TestGoWorkflow(TestCase):
             options={"artifact_executable_name": "no-deps-main-arm64"},
             architecture="arm64",
         )
-
         pathname = Path(self.artifacts_dir, "no-deps-main-arm64")
         self.assertEqual(get_executable_arch(pathname), "AArch64")
+
+    def test_builds_with_trimpath(self):
+        source_dir = os.path.join(self.TEST_DATA_FOLDER, "no-deps")
+        built_trimpath = self.builder.build(
+            source_dir,
+            self.artifacts_dir,
+            self.scratch_dir,
+            os.path.join(source_dir, "go.mod"),
+            runtime=self.runtime,
+            options={"artifact_executable_name": "no-deps-main-trimpath", "trim_go_path": True},
+            architecture="x86_64",
+        )
+        pathname = Path(self.artifacts_dir, "no-deps-main-trimpath")
+        self.assertEqual(get_executable_arch(pathname), "x64")
--- a/aws_lambda_builders/workflows/go_modules/builder.py
+++ b/aws_lambda_builders/workflows/go_modules/builder.py
@@ -21,7 +21,7 @@ class GoModulesBuilder(object):
 
     LANGUAGE = "go"
 
-    def __init__(self, osutils, binaries, mode=BuildMode.RELEASE, architecture=X86_64):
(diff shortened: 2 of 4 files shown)


bcochranemax (approved):

afischerMAX (approved):

sriram-mv:
Overall, this looks good. These options will need to be passed in from AWS SAM CLI as well here: https://github.com/aws/aws-sam-cli/blob/develop/samcli/lib/build/app_builder.py#L715

tests/integration/workflows/go_modules/test_go.py
```diff
@@ -112,6 +112,19 @@ def test_builds_arm64_architecture(self):
             options={"artifact_executable_name": "no-deps-main-arm64"},
             architecture="arm64",
         )
-
         pathname = Path(self.artifacts_dir, "no-deps-main-arm64")
         self.assertEqual(get_executable_arch(pathname), "AArch64")
+
+    def test_builds_with_trimpath(self):
```
sriram-mv, line 118: Can we add a test to show the difference between when trim path is enabled vs not enabled?

sriram-mv (changes requested):
I am happy to approve with a small test change.

mndeveci (changes requested):
Thanks for your contribution! I think changes looks good to me. Can we add an intergration test to validate this case? Integration test file can be found here; https://github.com/aws/aws-lambda-builders/blob/develop/tests/integration/workflows/go_modules/test_go.py#L18

tests/integration/workflows/go_modules/test_go.py
scrofungulus: @sriram-mv  @mndeveci I have added this test case. I wrote a utility that will return the md5 hexdigest of the built binaries. I created a copy of the `no-deps` go project in `testdata`. By building an exact copy of that project with `-trimpath` enabled we can assert that their md5 hex digests are the same. Similarly, we can assert that they will not be equal without the `-trimpath` flag. This was the most straightforward way that I could think of. Hopefully it is sufficient.

sriram-mv (approved):
Thanks for making the changes!

mndeveci (approved):
Thanks for your contribution, looks good to me as well!

sriram-mv:
@scrofungulus to use this effectively, the next step is to make changes in AWS SAM CLI as well. since it passes in the options based on the builder that is chosen. Those changes are showcased here: https://github.com/aws/aws-sam-cli/blob/develop/samcli/lib/build/app_builder.py#L714

Would you be open to contributing those changes there as well?

collen-maxex:
@sriram-mv I am absolutely open to updating it.

scrofungulus:
@sriram-mv I'm definitely open to contributing! Looks straightforward.

scrofungulus:
@sriram-mv pull number [4300](https://github.com/aws/aws-sam-cli/pull/4300) addresses your above comment! 😃

PR #401, opened 2022-11-10
Restore file permissions after unzip

Beau-Gosse-dev:
*Description of changes:* 

Because of [this issue](https://bugs.python.org/issue15795), files coming back from Amazon.Lambda.Tools were not getting the same permissions after unzipping. In the case where we want to `sam local invoke` on a custom runtime, this means we'll get a permission error since the unzipped files don't have execute permissions. This uses the [code from SAM CLI](https://github.com/aws/aws-sam-cli/blob/develop/samcli/local/lambdafn/zip.py#L81) to set permissions correctly.

### DIFF
--- a/aws_lambda_builders/workflows/dotnet_clipackage/utils.py
+++ b/aws_lambda_builders/workflows/dotnet_clipackage/utils.py
@@ -11,98 +11,6 @@ from aws_lambda_builders.utils import which
 LOG = logging.getLogger(__name__)
 
 
-def _is_symlink(file_info):
-    """
-    Check the upper 4 bits of the external attribute for a symlink.
-    See: https://unix.stackexchange.com/questions/14705/the-zip-formats-external-file-attribute
-    Parameters
-    ----------
-    file_info : zipfile.ZipInfo
-        The ZipInfo for a ZipFile
-    Returns
-    -------
-    bool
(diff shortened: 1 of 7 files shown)


aws_lambda_builders/workflows/dotnet_clipackage/utils.py
```diff
@@ -1,14 +1,107 @@
...
 import platform
-import shutil
 import subprocess
 import zipfile
 from aws_lambda_builders.utils import which
 
+LOG = logging.getLogger(__name__)
+
+
+def _is_symlink(file_info):
```
jfuss, line 14: Can you move this into the OSUtils instead of global scope?

It would be useful to have a comment and link to the code in SAM CLI. Just helps future people know we copied it from one place to the other.

jfuss (changes requested):
Overall, this looks good. Just a small comment to address.

aws_lambda_builders/workflows/dotnet_clipackage/utils.py
Beau-Gosse-dev: Yeah, I assume you want all the new methods inside OSUtils then? Is that right? I'll add the comment and a link to the code in SAM CLI

jfuss (approved):

mndeveci (approved):

PR #407, opened 2022-11-17
fix: Change npm bin to npm root

mildaniel:
*Description of changes:*
npm 9  has removed the `npm bin` command which was used to find the local installation of esbuild. 
This change switches the `npm bin` command to use `npm root` which should be backwards compatible.


hnnasit (approved):
LGTM

mndeveci:
> _Description of changes:_ npm 9 has removed the `npm bin` command which was used to find the local installation of esbuild. This change switches the `npm bin` command to use `npm root` which should be backwards compatible.
> 
> By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Just to confirm, `npm bin` doesn't work with npm v9, but `npm root` can work with earlier versions? So it won't break any existing customers, right?

mildaniel:
> > _Description of changes:_ npm 9 has removed the `npm bin` command which was used to find the local installation of esbuild. This change switches the `npm bin` command to use `npm root` which should be backwards compatible.
> > By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
> 
> Just to confirm, `npm bin` doesn't work with npm v9, but `npm root` can work with earlier versions? So it won't break any existing customers, right?

That's correct. [npm reference](https://docs.npmjs.com/cli/v8/commands/npm-root)

tests/unit/workflows/nodejs_npm_esbuild/test_workflow.py
```diff
@@ -76,15 +77,15 @@ def test_sets_up_esbuild_search_path_from_npm_bin(self):
             experimental_flags=[],
         )
 
-        self.osutils.popen.assert_called_with(["npm", "bin"], stdout="PIPE", stderr="PIPE", cwd="scratch_dir")
+        self.osutils.popen.assert_called_with(["npm", "root"], stdout="PIPE", stderr="PIPE", cwd="scratch_dir")
         esbuild = workflow.actions[2]._subprocess_esbuild
 
         self.assertIsInstance(esbuild, SubprocessEsbuild)
-        self.assertEqual(esbuild.executable_search_paths, ["project/bin"])
+        self.assertEqual(esbuild.executable_search_paths, [str(Path("project/.bin"))])
```
moelasmar, line 84: I tested running `npm root` command, and actually the output is not `project/.bin` as `npm bin` command output. my npm version is `8.1.1`

I think it will not make a difference in our case, but I am sharing that with you so may be I am missing any thing

aws_lambda_builders/workflows/nodejs_npm_esbuild/workflow.py
```diff
@@ -215,7 +216,8 @@ def get_resolvers(self):
 
     def _get_esbuild_subprocess(self, subprocess_npm, scratch_dir, osutils) -> SubprocessEsbuild:
         try:
-            npm_bin_path = subprocess_npm.run(["bin"], cwd=scratch_dir)
+            npm_bin_path_root = subprocess_npm.run(["root"], cwd=scratch_dir)
+            npm_bin_path = str(Path(npm_bin_path_root, ".bin"))
```
moelasmar, line 220: how can we make sure the `npm root` command is available in all supported npm versions?

moelasmar (commented):
should we update our Appveyor jobs to run the nodejs build test cases in different versions

tests/unit/workflows/nodejs_npm_esbuild/test_workflow.py
mndeveci: Given that there are some discrepancies, do we have integration tests that will catch any of these possible errors? Or these differences are just fine since all integration tests have passed.

mildaniel: In this case it shouldn't make a difference. `npm bin` would return the path to the binary directory itself which we mock and then later assert. `npm root` returns the path to the active `node_modules`, which should have a `.bin` directory within it. In this case we have to append that directory ourselves so that tests should essentially be equivalent.

Our integration test cases should already cover this case, we just need to run them using npm 9 in our test jobs.

aws_lambda_builders/workflows/nodejs_npm_esbuild/workflow.py
mildaniel: I checked in the documentation which versions this command is supported in https://docs.npmjs.com/cli/v8/commands/npm-root. Should be all the same versions as `npm bin` + npm 9.

mildaniel:
> should we update our Appveyor jobs to run the nodejs build test cases in different versions

I was planning on doing this as a follow-up item.

mndeveci (approved):

moelasmar (approved):

PR #408, opened 2022-11-20
CVE-2007-4559 Patch

TrellixVulnTeam:
# Patching CVE-2007-4559

Hi, we are security researchers from the Advanced Research Center at [Trellix](https://www.trellix.com). We have began a campaign to patch a widespread bug named CVE-2007-4559. CVE-2007-4559 is a 15 year old bug in the Python tarfile package. By using extract() or extractall() on a tarfile object without sanitizing input, a maliciously crafted .tar file could perform a directory path traversal attack. We found at least one unsantized extractall() in your codebase and are providing a patch for you via pull request. The patch essentially checks to see if all tarfile members will be extracted safely and throws an exception otherwise. We encourage you to use this patch or your own solution to secure against CVE-2007-4559. Further technical information about the vulnerability can be found in this [blog](https://www.trellix.com/en-us/about/newsroom/stories/research/tarfile-exploiting-the-world.html).

If you have further questions you may contact us through this projects lead researcher [Kasimir Schulz](mailto:kasimir.schulz@trellix.com).


hawflau (dismissed):
Thanks @TrellixVulnTeam for the security patch!

I believe we should look to move `is_within_directory` and `safe_extract` to a  common place to reduce repetition. Also should we add some tests?

hnnasit (approved):
LGTM!

aws_lambda_builders/utils.py
```diff
@@ -196,3 +196,28 @@ def create_symlink_or_copy(source: str, destination: str) -> None:
             exc_info=ex if LOG.isEnabledFor(logging.DEBUG) else None,
         )
         copytree(source, destination)
+
+
+def _is_within_directory(directory, target):
```
lucashuy, line 201: nit: can we add type hinting for the params and return value for these new methods?

tests/functional/test_utils.py
```diff
@@ -63,6 +63,22 @@ def test_must_return_valid_go_architecture(self):
         self.assertEqual(get_goarch(""), "amd64")
 
 
+class TestExtractTarFile(TestCase):
+    def test_extract_tarfile_unpacks_a_tar(self):
```
lucashuy, line 67: question: is testing the exception (ie actually performing the attack) something that should be done here?

hawflau: good catch. I intended but forgot to add the path traversal case. Will add it.

aws_lambda_builders/utils.py
hawflau: sure, can do

lucashuy (approved):

PR #410, opened 2022-11-25
test: Add GH Actions

mildaniel:
Migrating CI jobs from AppVeyor to GitHub actions.

This PR parallelizes runtime integration tests so that they can complete faster (from ~22 minutes to ~9 minutes).
It also parameterizes the runtime tests so that we can test multiple dependency versions per runtime (e.g. we're testing both npm 8.x and 9.x for Node.js-based runtimes).


.github/workflows/build.yml
```diff
@@ -0,0 +1,251 @@
...
+          - "3.8"
+          - "3.7"
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python }}
+      - uses: actions/setup-ruby@v1
+        with:
+          ruby-version: '^3.1'
```
mndeveci, line 228: Lambda supports ruby2.7, so should we setup that version for the tests as well?

mndeveci:
This is exciting! 🥳

.github/workflows/build.yml
mildaniel: Updated!

mndeveci (approved):

.github/workflows/build.yml
```diff
@@ -0,0 +1,255 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - ubuntu-latest
+          - windows-latest
+        python:
+          - "3.7"
+          - "3.8"
+          - "3.9"
```
hawflau, line 19: nit: shall we add "3.10" as well?

hawflau (approved):
approved with a small comment

.github/workflows/build.yml
mildaniel: I think this might result in exceeding the limit of jobs/pr... Might need a separate job that does a smoke test with all versions.

PR #411, opened 2022-11-28
fix: Copy over shrinkwrap if it exists

mildaniel:
*Description of changes:*
`npm 9.x` no longer packs the `npm-shrinkwrap.json` so we need to do it ourselves to respect this file during installation, similarly to how to copy over the `package-lock.json`.

### DIFF
--- a/tests/unit/workflows/nodejs_npm/test_actions.py
+++ b/tests/unit/workflows/nodejs_npm/test_actions.py
@@ -113,24 +113,31 @@ class TestNodejsNpmCIAction(TestCase):
 class TestNodejsNpmrcAndLockfileCopyAction(TestCase):
     @parameterized.expand(
         [
-            [False, False],
-            [True, False],
-            [False, True],
-            [True, True],
+            [False, False, False],
+            [True, False, False],
+            [False, True, False],
+            [True, True, False],
+            [False, False, True],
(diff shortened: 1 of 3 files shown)


tests/unit/workflows/nodejs_npm/test_actions.py
```diff
@@ -113,24 +113,31 @@ def test_raises_action_failed_when_npm_fails(self, SubprocessNpmMock):
...
-            [False, True],
-            [True, True],
+            [False, False, False],
+            [True, False, False],
+            [False, True, False],
+            [True, True, False],
+            [False, False, True],
+            [True, False, True],
+            [False, True, True],
+            [True, True, True],
```
mndeveci, line 123: nit: if this is all the combination of these 3 params, we can also write it like this;
```
@parameterized.expand(
  itertools.product([True, False], [True, False], [True, False])
)
```

mndeveci (approved):

torresxb1 (approved):

tests/unit/workflows/nodejs_npm/test_actions.py
mildaniel: Oh cool, let me try that, thanks!

PR #418, opened 2022-12-20
feat: handle default and supported values for building in source per workflow

torresxb1:
*Issue #, if available:*

*Description of changes:*
Building in source
- handles default values for each workflow
- handles supported values for each workflow
- ensures each workflow should define its own defaults and supported values
- error handling if the passed-in value is not supported by the workflow

### DIFF
--- a/aws_lambda_builders/workflow.py
+++ b/aws_lambda_builders/workflow.py
@@ -140,10 +140,13 @@ class _WorkflowMetaClass(type):
         if not isinstance(cls.CAPABILITY, Capability):
             raise ValueError("Workflow '{}' must register valid capabilities".format(cls.NAME))
 
-        # All workflows must define default value and supported values for build in source
-        if cls.BUILD_IN_SOURCE_BY_DEFAULT is None or cls.BUILD_IN_SOURCE_SUPPORT is None:
+        # All workflows must define valid default and supported values for build in source
+        if (
+            not isinstance(cls.BUILD_IN_SOURCE_SUPPORT, BuildInSourceSupport)
+            or cls.BUILD_IN_SOURCE_BY_DEFAULT not in cls.BUILD_IN_SOURCE_SUPPORT.value
+        ):
             raise ValueError(
-                "Workflow '{}' must define default value and supported values for build in source".format(cls.NAME)
+                "Workflow '{}' must define valid default and supported values for build in source".format(cls.NAME)
             )
 
         LOG.debug("Registering workflow '%s' with capability '%s'", cls.NAME, cls.CAPABILITY)
--- a/aws_lambda_builders/workflow.py
+++ b/aws_lambda_builders/workflow.py
@@ -261,26 +261,18 @@ class BaseWorkflow(object, metaclass=_WorkflowMetaClass):
         self.build_in_source = build_in_source
         if build_in_source is None:
             self.build_in_source = self.BUILD_IN_SOURCE_BY_DEFAULT
-        self._validate_build_in_source()
+        elif build_in_source not in self.BUILD_IN_SOURCE_SUPPORT.value:
+            LOG.warning(
+                'Workflow %s does not support value "%s" for building in source. Using default value: %s.',
+                self.NAME,
+                build_in_source,
+                self.BUILD_IN_SOURCE_BY_DEFAULT,
+            )
+            self.build_in_source = self.BUILD_IN_SOURCE_BY_DEFAULT
 
         # Actions are registered by the subclasses as they seem fit
         self.actions = []
         self._binaries = {}
-
-    def _validate_build_in_source(self):
-        """
-        Validates that the value of build_in_source is supported by the chosen workflow.
-        """
-        if self.build_in_source not in self.BUILD_IN_SOURCE_SUPPORT.value:
-            error_reason = ""
-            if self.BUILD_IN_SOURCE_SUPPORT == BuildInSourceSupport.NOT_SUPPORTED:
-                error_reason = "This workflow does not support building in source"
-            elif self.BUILD_IN_SOURCE_SUPPORT == BuildInSourceSupport.EXCLUSIVELY_SUPPORTED:
-                error_reason = "This workflow only supports building in source"
-            else:
-                error_reason = "Unsupported value for build_in_source"
-
-            raise WorkflowFailedError(workflow_name=self.NAME, action_name=None, reason=error_reason)
 
     def is_supported(self):
         """
--- a/aws_lambda_builders/workflow.py
+++ b/aws_lambda_builders/workflow.py
@@ -6,6 +6,7 @@ import os
 import logging
 
 from collections import namedtuple
+from enum import Enum
 
 from aws_lambda_builders.binary_path import BinaryPath
 from aws_lambda_builders.path_resolver import PathResolver
@@ -35,6 +36,21 @@ class BuildMode(object):
 
     DEBUG = "debug"
     RELEASE = "release"
+
+
+class BuildInSourceSupport(Enum):
+    """
+    Enum to define a workflow's support for building in source.
+    """
+
+    # can't build in source directory (e.g. only able to build in temporary or artifacts directories)
+    NOT_SUPPORTED = [False]
+
+    # can build in source directory but not required to
+    OPTIONALLY_SUPPORTED = [False, True]
+
+    # only able to build in source directory and not somewhere else
+    EXCLUSIVELY_SUPPORTED = [True]
 
 
 # TODO: Move sanitize out to its own class.
@@ -124,6 +140,12 @@ class _WorkflowMetaClass(type):
         if not isinstance(cls.CAPABILITY, Capability):
             raise ValueError("Workflow '{}' must register valid capabilities".format(cls.NAME))
 
+        # All workflows must define default value and supported values for build in source
+        if cls.BUILD_IN_SOURCE_BY_DEFAULT is None or cls.BUILD_IN_SOURCE_SUPPORT is None:
+            raise ValueError(
+                "Workflow '{}' must define default value and supported values for build in source".format(cls.NAME)
+            )
+
         LOG.debug("Registering workflow '%s' with capability '%s'", cls.NAME, cls.CAPABILITY)
         DEFAULT_REGISTRY[cls.CAPABILITY] = cls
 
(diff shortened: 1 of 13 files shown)


aws_lambda_builders/workflows/custom_make/workflow.py
```diff
@@ -44,10 +47,6 @@ def __init__(self, source_dir, artifacts_dir, scratch_dir, manifest_path, runtim
 
         subprocess_make = SubProcessMake(make_exe=self.binaries["make"].binary_path, osutils=self.os_utils)
 
-        # Don't build in source by default (backwards compatibility)
-        if build_in_source is None:
-            build_in_source = False
-
```
torresxb1, line 50: I recently added this in a previous PR to handle build-in-source defaults for custom makefile, but it's not needed anymore with the new class constant logic (`BUILD_IN_SOURCE_BY_DEFAULT`)

aws_lambda_builders/workflow.py
```diff
@@ -229,12 +257,31 @@ def __init__(
...
+        if self.build_in_source not in self.BUILD_IN_SOURCE_SUPPORT.value:
+            error_reason = ""
+            if self.BUILD_IN_SOURCE_SUPPORT == BuildInSourceSupport.NOT_SUPPORTED:
+                error_reason = "This workflow does not support building in source"
+            elif self.BUILD_IN_SOURCE_SUPPORT == BuildInSourceSupport.EXCLUSIVELY_SUPPORTED:
+                error_reason = "This workflow only supports building in source"
+            else:
+                error_reason = "Unsupported value for build_in_source"
+
+            raise WorkflowFailedError(workflow_name=self.NAME, action_name=None, reason=error_reason)
```
torresxb1, line 283: hmmm actually we might not want to fail. I might speak with UX about this but I think I'll change this to show a warning/log and use the default value for the workflow.

torresxb1: For context: one of the reasons we might not want this to fail is that the `--in-source` flag is applied to all functions and layers of a project for now. If there's one function/layer that doesn't support building in source, this would fail the whole build. If it just raised a warning and used the default value, users would still be able to build (although there might be less visibility into the warning message vs. an error message)

aws_lambda_builders/workflow.py
```diff
@@ -124,6 +140,12 @@ def __new__(mcs, name, bases, class_dict):
         if not isinstance(cls.CAPABILITY, Capability):
             raise ValueError("Workflow '{}' must register valid capabilities".format(cls.NAME))
 
+        # All workflows must define default value and supported values for build in source
+        if cls.BUILD_IN_SOURCE_BY_DEFAULT is None or cls.BUILD_IN_SOURCE_SUPPORT is None:
```
mndeveci, line 144: When this case should happen? I thought we will always fallback to default behaviour if something is wrong with the parameters (for also keeping backward compatibility)

aws_lambda_builders/workflow.py
```diff
@@ -229,12 +257,31 @@ def __init__(
...
+        Validates that the value of build_in_source is supported by the chosen workflow.
+        """
+        if self.build_in_source not in self.BUILD_IN_SOURCE_SUPPORT.value:
+            error_reason = ""
+            if self.BUILD_IN_SOURCE_SUPPORT == BuildInSourceSupport.NOT_SUPPORTED:
+                error_reason = "This workflow does not support building in source"
+            elif self.BUILD_IN_SOURCE_SUPPORT == BuildInSourceSupport.EXCLUSIVELY_SUPPORTED:
+                error_reason = "This workflow only supports building in source"
+            else:
+                error_reason = "Unsupported value for build_in_source"
```
mndeveci, line 281: Can we simplify this logic?

aws_lambda_builders/workflow.py
torresxb1: This is to ensure each workflow class implementation defines these two class constants/attributes. It makes sure that someone doesn't add a new workflow without specifying whether it supports building in source and whether it build in source by default or not.

mndeveci: Got it, then can we add unit tests to catch these errors when we introduce a new build workflow? That test can basically try to initialize all workflows and confirm none of them throws this exception.

aws_lambda_builders/workflow.py
torresxb1: Hmm it could be something like this:

```suggestion
    def _validate_build_in_source(self):
        """
        Validates that the value of build_in_source is supported by the chosen workflow.
        """
        if self.build_in_source not in self.BUILD_IN_SOURCE_SUPPORT.value:
            error_reason = f"Unsupported value for build_in_source. This workflow supports values {self.BUILD_IN_SOURCE_SUPPORT.value}"
```

I just thought the other messages above had better UX/messaging. Plus, I don't really like to include `build_in_source` as the name of this lambda builders parameter could be different than the user-exposed one on SAM CLI.

aws_lambda_builders/workflow.py
torresxb1: Ah forgot to include a screenshot, but tests already catch this (this logic is called when you initialize the class, I believe):
Screen Shot 2022-12-21 at 10 45 21 AM

torresxb1: ah I just realized you meant a test that initializes all workflows. 🤔  not sure if it's strictly necessary, as long as the workflow is part of __any__ test (even manual testing), this should be caught.

aws_lambda_builders/workflow.py
torresxb1: Maybe it could be: "Unsupported value for building in source. This workflow supports values {self.BUILD_IN_SOURCE_SUPPORT.value}"

aws_lambda_builders/workflow.py
torresxb1: update: confirmed this with PM and changed logic to show warning LOG instead of erroring out.

This is what it would look like if custom make didn't support building in source:
Screen Shot 2022-12-21 at 1 55 11 PM

aws_lambda_builders/workflow.py
torresxb1: updated this

aws_lambda_builders/workflow.py
mndeveci: I was referring to a unit test, which will find all workflow definitions and try to initialize them. Any workflow without these values will fail (I wonder if we can enforce this by typing) 🤔 

Can you confirm this part will be validated with integration testing, if so then it shouldn't be a problem for the future?

aws_lambda_builders/workflow.py
```diff
@@ -229,7 +257,18 @@ def __init__(
         self.architecture = architecture
         self.is_building_layer = is_building_layer
         self.experimental_flags = experimental_flags if experimental_flags else []
+
         self.build_in_source = build_in_source
+        if build_in_source is None:
+            self.build_in_source = self.BUILD_IN_SOURCE_BY_DEFAULT
```
sriram-mv, line 263: confused here, we set the same logic both if and in elif?

`self.build_in_source = self.BUILD_IN_SOURCE_BY_DEFAULT`

aws_lambda_builders/workflow.py
mndeveci: So we won't fail but will print the warning message? (I couldn't find this code any more 😂)

aws_lambda_builders/workflow.py
sriram-mv: I vote we need the test, we don't explicit check for the ValueError being raised.

sriram-mv (changes requested):

aws_lambda_builders/workflow.py
torresxb1: Basically, we should use the default value if the passed in value either <is none> or <is not supported> (but also we only show a warning message when it's not supported). Otherwise it should use the passed in value.

aws_lambda_builders/workflow.py
torresxb1: correct + I generalized/simplified the validation messaging stuff

aws_lambda_builders/workflow.py
torresxb1: Since I believe this validation logic runs when the classes are loaded, I think this will be caught by any test. I tested it by adding a new workflow (`MyWorkflow`) without defining the build-in-source-related constants, and I didn't add any new tests. Then I ran `make pr` and it was caught by the first unit test found:
Screen Shot 2023-01-03 at 1 31 51 PM

torresxb1: Overall, I don't think a test for this is necessary. Otherwise, not really sure what this test would look like. As I can just think of including `import aws_lambda_builders.workflows` in some test and not asserting anything.

aws_lambda_builders/workflow.py
sriram-mv:
```
if build_in_source not in self.BUILD_IN_SOURCE_SUPPORT.value:
   self.build_in_source = self.BUILD_IN_SOURCE_BY_DEFAULT
```

you could warnings for None too, since technically they do not need to be defined at the workflow level.

Does this suffice?

torresxb1: @sriram-mv Not sure if I understood your comment.

So, most of the time `build_in_source` will likely be `None` as this will be the default from the SAM CLI (can't really set a True/False default since each workflow currently has a different value as default). On the lambda builders side, each workflow will decide what the default value is when it receives `None` (what the couple lines above accomplish).

I don't think we should warn for `None` because that means by default, if the customer doesn't provide any specific True/False value from the sam cli side, they'll get this warning message. I don't feel like it's a good experience to get a new warning message for a valid use case that they were already doing before. I think ideally we only show this warning message if they explicitly provide an unsupported value (which is what I currently have here).

sriram-mv: Having the same assignment in both if and elif makes me question if we could do a different logic here, which is why I proposed that change. Agree on not adding warning logs for something that has not changed.

aws_lambda_builders/workflow.py
```diff
@@ -148,6 +173,12 @@ class BaseWorkflow(object, metaclass=_WorkflowMetaClass):
     # Optional list of manifests file/folder names supported by this workflow.
     SUPPORTED_MANIFESTS = []
 
+    # Whether the workflow builds in source by default, each workflow should define this.
+    # (some workflows build in temporary or artifact directories by default)
+    BUILD_IN_SOURCE_BY_DEFAULT = None
+    # Support for building in source, each workflow should define this.
+    BUILD_IN_SOURCE_SUPPORT = None
+
```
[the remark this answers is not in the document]
torresxb1: @mndeveci it won't be None by default. Each workflow is required to set their own default, and will raise an error if the workflow doesn't specify it (see lines 143-151 above)

mndeveci (approved):

sriram-mv (approved):
Not blocking, but will leave my comments in there on same code path being executed on if and elif. There could be ways in which we make that cleaner.

aws_lambda_builders/workflow.py
torresxb1: refactored a bit

PR #422, opened 2023-01-09
Add CodeQL GH Action

hawflau:

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -23,6 +23,7 @@ on:
 jobs:
   analyze:
     name: Analyze
+    if: github.repository_owner == 'aws'
     runs-on: ubuntu-latest
     permissions:
       actions: read
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -17,6 +17,8 @@ on:
   pull_request:
     # The branches below must be a subset of the branches above
     branches: [ "develop" ]
+  schedule:
+    - cron: '44 21 * * 2'
 
 jobs:
   analyze:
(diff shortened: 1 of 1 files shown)


hnnasit (approved):
LGTM!

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,77 @@
...
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
```
mndeveci, line 14: Sorry if this is a dummy question but just trying to understand our intention here. Why do we need to run this for push to branches as well as on a schedule? Wouldn't be enough if we only have it for pull requests?

hawflau: As stated in the comment in the yml:
```
https://github.com/aws/aws-sam-cli/pull/4552/checks?check_run_id=10533852983
```

I'll keep the on push as well, but limit the branches to only develop and main

mndeveci (approved):

.github/workflows/codeql.yml
```diff
@@ -0,0 +1,75 @@
+# For most projects, this workflow file will not need changing; you simply need
```
mndeveci, line 1: Can we also configure to run it only for `aws` repository? Like this one here: https://github.com/aws/aws-lambda-builders/blob/develop/.github/workflows/build.yml#L43

mndeveci (approved):

Issue #382, opened 2022-08-13
Bug: recursively copies artifact dir when it's in source dir


israelg99:
Recursively copies artifact dir when it's in source dir until it errors out:

JSON-RPC input:
```json
{
...
"source_dir":    "path/to/source",
"artifacts_dir":  "path/to/source/.build/artifact",
...
}
```

Then running `lambda-builders <JSON INPUT>`

Results in
```bash
PythonPipBuilder:CopySource - [Errno 63] File name too long: '/path/to/source/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/.build/artifact/python_dateutil-2.8.2.dist-info/top_level.txt'
```

Setting `artifact_dir` outside the `source_dir` solves this issue.

----

**Is this a bug, or is this expected? Perhaps we should add `ignore: {}` interface? I can contribute a PR for this if you guys can point me in the right direction.**

----

#### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: Mac
2. If using SAM CLI, `sam --version`:
3. AWS region: us-east-1 (shouldn't matter)

`Add --debug flag to any SAM CLI commands you are running`

jfuss:
@israelg99 `artifact_dir` and `source_dir` should be different directories unless you wanted to embed (and copy them) for some reason. To my knowledge, we don't want to restrict the behavior here so we don't want to fail the run. Maybe we should? But not sure what that could break now, if we do.

[...]

Issue #185, opened 2020-07-20
Python: Extend and/or override EXCLUDED_FILES


milesgranger:
### Describe your idea/feature/enhancement

It appears [the EXCLUDED_FILES is hardcoded.](https://github.com/awslabs/aws-lambda-builders/blob/develop/aws_lambda_builders/workflows/python_pip/workflow.py#L21)

It would be delightful if this could be extended/modified. I have many files and directories which are ignored by my own `.gitignore` but these continue to be included in the build directory by SAM. (`build/`, `.eggs/`, `.ipython/`, `.ipynb_checkpoints/`, `dist/`, `package.json`, `package-lock.json`, etc, etc)

Then there are files that shouldn't be ignored by git but should be ignored during the build such as a `tests` directory.

Furthermore, I don't actually want to ignore `.git/` b/c I use [setuptools-scm](https://pypi.org/project/setuptools-scm/) to control versioning with git, and during the build that fails because it has no git context.

### Proposal

Add configuration ability to override or extend with existing `.gitignore` or with own list, ie a `.samignore` file

carpnick:
This needs to be exposed in AWS SAM to support other tools like - the AWS cfn tool set. - IE - https://github.com/aws-cloudformation/cloudformation-cli-python-plugin

Directly related to this issue - https://github.com/aws-cloudformation/cloudformation-cli-python-plugin/issues/247

HWiese1980:
Has this feature been implemented? I'm running into the same problem with a python project that contains big virtual environments that I'd like to exclude.

festusyuma:
> at contains bi

Hi, Have you had any luck with this issue, I have some symlinks in my project node_modules/.bin and I am unable to run sam build --use-container cos it keeps trying to copy the symlinks and end up with errors like this

FileNotFoundError: [Errno 2] No such file or directory: '***/backend/node_modules/.bin/ts-node'

mildaniel:
Created a separate issue to consolidate the discussion for include/exclude functionality https://github.com/aws/aws-lambda-builders/issues/516.

PR #424, opened 2023-01-15
add option to ignore files/dirs

pkit:
adds a new option to all workflow builders: ignore specific files and directories

ex:
```
{
  options: { "ignore": [".build", ".tmp*"] }
}
```

fixes: #382 
fixes: #185 

And other numerous issues asking for a simple ignore feature

Added a test for python because I don't care for anything else right now.
But the feature is fully backwards compatible anyway.

### DIFF
--- a/aws_lambda_builders/actions.py
+++ b/aws_lambda_builders/actions.py
@@ -107,13 +107,16 @@ class CopySourceAction(BaseAction):
 
     PURPOSE = Purpose.COPY_SOURCE
 
-    def __init__(self, source_dir, dest_dir, excludes=None):
+    def __init__(self, source_dir, dest_dir, excludes=None, ignore=None):
         self.source_dir = source_dir
         self.dest_dir = dest_dir
         self.excludes = excludes or []
+        self.ignore = list(self.excludes)
+        if ignore and isinstance(ignore, (list, tuple)):
+            self.ignore += ignore
 
     def execute(self):
-        copytree(self.source_dir, self.dest_dir, ignore=shutil.ignore_patterns(*self.excludes))
+        copytree(self.source_dir, self.dest_dir, ignore=shutil.ignore_patterns(*self.ignore))
 
 
 class LinkSourceAction(BaseAction):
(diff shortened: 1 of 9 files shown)


carpnick:
Does it solve #185?  - The fact Excluded files is hard coded is the problem.  As I see this, isnt this still using the hard coded `EXCLUDED_FILES` var which means it still wont fix the use case as far as I am aware.

pkit:
Hmm, I wanted it to be backwards compatible. But it's not that hard to improve. For example, I can add `exclude` option, if set to `[]` it will remove all the hardcoded excludes.

carpnick:
> Hmm, I wanted it to be backwards compatible. But it's not that hard to improve. For example, I can add `exclude` option, if set to `[]` it will remove all the hardcoded excludes.

I think the easier route would be allow exclude to be set.  If unset, set to default `EXCLUDED_FILES`, then this allows the SAM repo to expose a parameter to send in your own ignore patterns.

pkit:
I'm not sure I want to force users to set the existing hundred excludes for python, just because they need to add only one.

carpnick:
> I'm not sure I want to force users to set the existing hundred excludes for python, just because they need to add only one.

Good point.  In #185 use case I need to remove some :).  So the only way I know how to support both is you override or dont.

torresxb1:
Thanks for bringing this to our attention and thanks for the discussion. Since it seems like we have some use cases where we want to ignore more things than `EXCLUDED_FILES` and some use cases where we want to ignore less things (like @carpnick 's case), then I do think the best solution might be some way of exposing control of this `EXCLUDED_FILES`. Not sure if the best way to do would be an option, an environment variable, or something else. 

But I do see @pkit 's concern:
> I'm not sure I want to force users to set the existing hundred excludes for python, just because they need to add only one.

Maybe the best way would be allowing three possibilities of controlling `EXCLUDED_FILES`: setting it (to manually specify/control this list), adding to it (to specify more files, on top of the default ones, to ignore), and removing from it (to specify which files not to ignore from the default list)

Anyway, will discuss with the team, and put this PR in draft in the meantime. Let us know if you have more thoughts!

pkit:
First I wanted to create a .gitignore-like file that can be fed into the json-rpc options.
This way we should solve all the problems even if we merge it with the existing EXCLUDES (as it has `!` for disabling paths)
But it looks like there is no good first-class support for gitignore in python, all packages I've checked lack some features or just don't even work. Writing a real gitignore support wasn't my goal, but maybe I'll dedicate some more time to it.

sriram-mv:
Related to https://github.com/aws/aws-lambda-builders/issues/430 atleast with respect to having a common functionality on includes/excludes: https://github.com/aws/aws-lambda-builders/issues/430

PR #428, opened 2023-01-18
refactor: assign each workflow a default build directory

torresxb1:
*Issue #, if available:*

*Description of changes:*

A recent change (https://github.com/aws/aws-lambda-builders/pull/418) added `BUILD_IN_SOURCE_BY_DEFAULT` to each workflow. This PR replaces that with `DEFAULT_BUILD_DIR`, which helps us generalize some logic more. We can now set and use a workflow's `build_dir` property using a combination of:
- the value of the `build_in_source` parameter passed in to lambda builders
  - if user wants to build in source, the build directory should be set to the source directory
- the workflow's default build directory (`DEFAULT_BUILD_DIR`)
  - in case `build_in_source` is not set, or it's set to `False`, or an unsupported value is passed in, we should use the workflow's existing default build directory

Note that we're still able to find out if a workflow builds in source by default just by checking if their `DEFAULT_BUILD_DIR` is the source directory. 

### DIFF
--- a/aws_lambda_builders/workflow.py
+++ b/aws_lambda_builders/workflow.py
@@ -38,10 +38,12 @@ class BuildMode(object):
     DEBUG = "debug"
     RELEASE = "release"
 
+
 class BuildDirectory(Enum):
     SCRATCH = "scratch"
     ARTIFACTS = "artifacts"
     SOURCE = "source"
+
 
 class BuildInSourceSupport(Enum):
     """
@@ -147,15 +149,11 @@ class _WorkflowMetaClass(type):
 
         # All workflows must define supported values for build in source
         if not isinstance(cls.BUILD_IN_SOURCE_SUPPORT, BuildInSourceSupport):
-            raise ValueError(
-                "Workflow '{}' must define supported values for build in source".format(cls.NAME)
-            )
+            raise ValueError("Workflow '{}' must define supported values for build in source".format(cls.NAME))
 
         # All workflows must define default build directory
         if not isinstance(cls.DEFAULT_BUILD_DIR, BuildDirectory):
-            raise ValueError(
-                "Workflow '{}' must define default build directory".format(cls.NAME)
-            )
+            raise ValueError("Workflow '{}' must define default build directory".format(cls.NAME))
 
         LOG.debug("Registering workflow '%s' with capability '%s'", cls.NAME, cls.CAPABILITY)
         DEFAULT_REGISTRY[cls.CAPABILITY] = cls
@@ -274,7 +272,6 @@ class BaseWorkflow(object, metaclass=_WorkflowMetaClass):
         self.actions = []
         self._binaries = {}
 
-
     def _select_build_dir(self, build_in_source: Optional[bool]) -> str:
         """
         Returns the build directory for the workflow.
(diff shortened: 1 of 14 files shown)


aws_lambda_builders/workflows/custom_make/workflow.py
```diff
@@ -23,7 +23,7 @@ class CustomMakeWorkflow(BaseWorkflow):
 
     EXCLUDED_FILES = (".aws-sam", ".git")
 
-    BUILD_IN_SOURCE_BY_DEFAULT = False
+    DEFAULT_BUILD_DIR = BuildDirectory.SCRATCH
```
torresxb1, line 26: by making this change, we can remove the logic of choosing a working/build directory based on the `build_in_source` parameter (see `_select_working_directory` below). Similarly, in other workflows we won't have to define this kind of logic either as the parent Workflow class now handles setting the `build_dir`.

aws_lambda_builders/workflows/custom_make/workflow.py
```diff
@@ -63,18 +60,12 @@ def __init__(self, source_dir, artifacts_dir, scratch_dir, manifest_path, runtim
...
-        if not self.build_in_source:
-            # if we're building on scratch_dir, we have to first copy the source there
-            self.actions.append(CopySourceAction(source_dir, scratch_dir, excludes=self.EXCLUDED_FILES))
+        if working_directory != source_dir:
+            # if we're not building in the source directory, we have to first copy the source
+            self.actions.append(CopySourceAction(source_dir, working_directory, excludes=self.EXCLUDED_FILES))
 
         self.actions.append(make_action)
 
-    def _select_working_directory(self, source_dir: str, scratch_dir: str, build_in_source: bool):
```
torresxb1, line 72: Due to the changes in this PR, we can remove the logic of choosing a working/build directory based on the `build_in_source` parameter. Similarly, in other workflows we won't have to define this kind of logic either because the parent Workflow class now handles setting the `build_dir`.

aws_lambda_builders/workflow.py
```diff
@@ -260,22 +266,38 @@ def __init__(
         self.architecture = architecture
         self.is_building_layer = is_building_layer
         self.experimental_flags = experimental_flags if experimental_flags else []
+        self.build_dir = self._select_build_dir(build_in_source)
+
+        # Actions are registered by the subclasses as they seem fit
+        self.actions = []
+        self._binaries = {}
```
torresxb1, line 273: just moved these up, didn't add these

lucashuy (approved):

sriram-mv:
We set a different build_dir per workflow is that any change in behavior from what is present today?

torresxb1:
>  We set a different build_dir per workflow is that any change in behavior from what is present today?

No change in customer-facing behaviour. Today each workflow has a different default build/install dir (some already build in the source directory, some in scratch, some in artifacts). 

This just explicitly (and clearly) writes what that default is for each workflow, and makes it a class attribute so that the parent class (`BaseWorkflow`) can generalize some logic: the build directory should be the source directory if we want to build in source, otherwise it should be the default.

sriram-mv (approved):
Approving since its a backward compatible change.

PR #431, opened 2023-01-25
Add Include resource functionality to npm and npm/esbuild

jasonterando:
Issue #: [430](https://github.com/aws/aws-lambda-builders/issues/430)

Description of changes:

* Addition of support for Metadata `Include` property that can be set to a file mask, or list of file masks, of files in the source directory, and copy those to the artifact (output) directory after build/packaging
* An action called `CopyResourceAction` and utilities `glob_copy` and `get_option_from_args` have been added to support the functionality

Note:  esbuild will work with the SAM CLI as-is, but for npm (without esbuild), SAM CLI will need to be updated to support sending include (currently, it will only send use_npm_ci and no other Metadata options).


qingchm:
Thanks for opening this pull request! We will look at your feature request and see if your changes can be accepted!

jasonterando:
I'll take a look and see what is causing unit tests to fail in Windows

aws_lambda_builders/utils.py
```diff
@@ -222,3 +223,32 @@ def extract_tarfile(tarfile_path: Union[str, os.PathLike], unpack_dir: Union[str
...
         tar.extractall(unpack_dir)
+
+
+def glob_copy(source: Union[str, list], destination: str) -> None:
+    items = source if isinstance(source, list) else [source]
+    dest_path = Path(destination)
+    known_paths = []
+    for item in items:
+        if Path(item).is_absolute():
+            raise ValueError('"{item}" is not a relative path'.format(item=item))
```
mildaniel, line 234: Let's use custom exceptions instead of raising `ValueError`.

aws_lambda_builders/utils.py
```diff
@@ -222,3 +223,32 @@ def extract_tarfile(tarfile_path: Union[str, os.PathLike], unpack_dir: Union[str
...
+            if save_to_dir not in known_paths:
+                os.makedirs(save_to_dir, exist_ok=True)
+                known_paths.append(save_to_dir)
+            shutil.copyfile(file, save_to)
+
+
+def get_option_from_args(args: Union[None, Dict[str, any]], option_name: str) -> any:
+    if args is not None:
+        if "options" in args:
+            if args["options"] is not None:
```
mildaniel, line 251: Let's do 
```
if args.get("options", None) is not None: 
```
instead so that if options isn't there, it doesn't raise a key error. And the same thing for wherever else we do dictionary accesses like this

mildaniel (changes requested):
Thanks for taking the time to raise this PR! I've left a few comments but overall it looks good. I do wonder if this should be applied to other runtimes since it seems it's applicable to more than just Node.js (maybe compiled languages). 

Would you also be able to raise the corresponding SAM-CLI PR for sending this parameter down?

aws_lambda_builders/utils.py
jasonterando: Any specific one you want me to use, or should I roll one?

aws_lambda_builders/utils.py
jasonterando: Roger that.  Is this the correct way - should I be doing something more "pythonic"?

```python
def get_option_from_args(args: Union[None, Dict[str, any]], option_name: str) -> any:
    if args is not None:
        options = args.get("options", None)
        if options is not None:
            return options.get(option_name, None)
    return None
```

jasonterando:
> Would you also be able to raise the corresponding SAM-CLI PR for sending this parameter down?

Assuming we can get this shepherded through, sure!

jasonterando:
I've updated the following:
1. Created a FileOperationError to raise when there is a problem copying
2. Refactored get_option_from_args to implement the suggested .get syntax
3. Added function robust_rmtree to facilitate retries for shutil.rmtree to workaround a Python on Windows issue (Python issue 40143)

aws_lambda_builders/utils.py
```diff
@@ -222,3 +225,53 @@ def extract_tarfile(tarfile_path: Union[str, os.PathLike], unpack_dir: Union[str
                 raise tarfile.ExtractError("Attempted Path Traversal in Tar File")
 
         tar.extractall(unpack_dir)
+
+
+def glob_copy(source: Union[str, list], destination: str) -> None:
```
mildaniel, line 230: Let's add type hints and docstrings to all functions.

aws_lambda_builders/utils.py
```diff
@@ -222,3 +225,53 @@ def extract_tarfile(tarfile_path: Union[str, os.PathLike], unpack_dir: Union[str
...
+    items = source if isinstance(source, list) else [source]
+    dest_path = Path(destination)
+    known_paths = []
+    for item in items:
+        if os.path.isabs(item.replace('\\', '/')):
```
mildaniel, line 235: Do we need the `item.replace()`? I think `os.path.isabs()` should already handle this, I could be wrong though.

aws_lambda_builders/workflows/nodejs_npm/workflow.py
```diff
@@ -65,6 +67,12 @@ def __init__(self, source_dir, artifacts_dir, scratch_dir, manifest_path, runtim
             source_dir, artifacts_dir, scratch_dir, manifest_path, osutils, subprocess_npm
         )
 
+        include = get_option_from_args(kwargs, "include")
+        if isinstance(include, (list, str)):
+            self.actions.append(CopyResourceAction(source_dir, include, artifacts_dir))
+        elif include is not None:
+            raise ValueError("Resource include items must be strings or lists of strings")
```
mildaniel, line 74: Let's use a custom exception here too.

aws_lambda_builders/utils.py
```diff
@@ -222,3 +225,53 @@ def extract_tarfile(tarfile_path: Union[str, os.PathLike], unpack_dir: Union[str
...
+            save_to = str(dest_path.joinpath(file))
+            LOG.debug("Copying resource file from source (%s) to destination (%s)", file, save_to)
+            save_to_dir = os.path.dirname(save_to)
+            if save_to_dir not in known_paths:
+                os.makedirs(save_to_dir, exist_ok=True)
+                known_paths.append(save_to_dir)
+            shutil.copyfile(file, save_to)
+
+
+def robust_rmtree(path, timeout=1):
```
mildaniel, line 250: If this is only used in tests, we should move it there.

mildaniel (changes requested):
Another thing to consider, and I'll take ownership of this, is that this will require a documentation update to publicize this new option.

aws_lambda_builders/utils.py
jasonterando: I had to have the replace in there to get the "absolute" detection to work properly under Windows.  "\foo" is considered a relative path, but "/foo" is absolute (go figure).  The purposes of this call is to make sure that the requested asset 's path is relative to the project, so either "\foo" or "/foo" should be treated as non-relative.

aws_lambda_builders/workflows/nodejs_npm/workflow.py
jasonterando: Yes sir

aws_lambda_builders/exceptions.py
```diff
@@ -73,3 +73,11 @@ class WorkflowUnknownError(LambdaBuilderError):
...
 
     MESSAGE = "{workflow_name}:{action_name} - {reason}"
+
+
+class FileOperationError(LambdaBuilderError):
+    """
+    Raised when a file operation fails
+    """
+
+    MESSAGE = "{operation_name} - {reason}"
```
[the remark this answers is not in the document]
jasonterando: Sorry, missed that

jasonterando:
If I take a stab at working on this stuff over the weekend, can you enable me to have Build & Test checks run.  I promise not to submit 100 times.

mildaniel:
> If I take a stab at working on this stuff over the weekend, can you enable me to have Build & Test checks run. I promise not to submit 100 times.

Already enabled 👍

aws_lambda_builders/utils.py
jasonterando: Moved to test/functional/test_utils.py

jasonterando:
Hi, latest changes committed - `make pr`  passed and I got all but one Gradle integration test to run (is there a Dockerfile available to set up Java, Go, Ruby, .NET, NodeJS?).

With regard to documentation, happy to submit it if you tell me where it should go and in what format.  Here's some Markdown I've put together as a first pass.

````
## **Include** Resources (optional)

This optional build property for nodejs_npm and nodejs_npm_esbuild workflows will copy files to the output (artifacts) directory, after bundling, for inclusion in deployment.

Valid value types include string, dictionary, or a list containing either strings and/or dictionaries.

### String Parameter

Use this to deep-copy files matching the specfied glob-style parameter to the artifact directory.

Example:

```yaml
Resources:
  MyLambda:
    Type: AWS::Serverless::Function
    Metadata:
       BuildMethod: esbuild
       BulidProperties:
         Include: resources/*.json
```

*Project (Source) Directory*
```
index.ts
  /resources
    file1.json
    file2.json
      /data
        file3.json
```

*Artifact (Destination) Directory*

```
index.js
file1.json
file2.json
/data
  file3.json
```

### Dictionary (Object) Parameter

Use this to control where files are copied to in the artifact directory and whether they are deep-copied.

Example:

```yaml
Resources:
  MyLambda:
    Type: AWS::Serverless::Function
    Metadata:
      BuildMethod: esbuild
      BuildProperties:
        Include:
          Source: resources/*.json
          Destination: resources
          Recursive: False
```

*Project (Source) Directory*
```
index.ts
  /resources
    file1.json
    file2.json
      /data
        file3.json
```

*Artifact (Destination) Directory*

```
index.js
  /resources
    file1.json
    file2.json
```

## List Parameter

A list of strings and dictionaries as described above can be used to trigger multiple copy operations.
````

moelasmar (commented):
I have a general comment on this PR, I feel that this action is not related to a specific Runtime, and instead it can be part of `sam build` command or may be `sam package` but not part of `aws-lambda-builders` library. I will raise this PR with the team to discuss what is the better approach to implement this feature.

jasonterando:
> I have a general comment on this PR, I feel that this action is not related to a specific Runtime, and instead it can be part of `sam build` command or may be `sam package` but not part of `aws-lambda-builders` library. I will raise this PR with the team to discuss what is the better approach to implement this feature.
> 
> I will make this PR as draft till we discuss it, and will update it later.

I don't disagree that, ideally, this would be a Runtime-independent feature.  However, even within NodeJS, there are differences in _when_ you implement it, based upon whether esbuild is being used or not. You could go the route of creating some kind of event-driven mechanism so that the individual Runtime workflows can take actions (before build, after build, copy resources after build, etc.), but that isn't there today.

Anyways, it's months later and I had to end up going a different route (orchestrating esbuild external to SAM).  If you want to nuke this PR, that's fine, we have a workaround, although it would have been nicer to use the integrated esbuild workflow.

mildaniel:
Closing until we figure out the correct path forward. Consolidating include/exclude functionality design and discussion into a single issue #516.

PR #433, opened 2023-01-28
refactor: esbuild refactor for readability

torresxb1:
*Issue #, if available:*

*Description of changes:*
I originally intended to add build-in-source support for the esbuild workflow but I found it hard with the way it was currently set up. I'm hoping my changes make it more readable and easier to work with in general, but let me know if you disagree with my changes or you think there could be further improvements.

Main changes done:
- Changed the bundle action:
  - parameter names were kind of coupled to scratch and artifacts directory, so I changed them to accept a working directory and an output directory
  - it now performs the esbuild version checking as part of this action (when needed)
- workflow conditionals/IFs cleanup: my main issue with the workflow was the accelerate workflow stuff was a bit complicated, with a bunch of conditions and nested ifs that went a few levels deep. In my proposed changes for the workflow, IFs would only go a single level deep.


aws_lambda_builders/workflows/nodejs_npm_esbuild/actions.py
```diff
@@ -74,10 +75,15 @@ def execute(self) -> None:
...
-            self._scratch_dir, self._artifacts_dir, self._bundler_config, self._osutils, self._manifest
+            self._working_directory, self._output_directory, self._bundler_config, self._osutils, self._manifest
         )
 
         if self._should_bundle_deps_externally():
+            check_minimum_esbuild_version(
+                minimum_version_required=MINIMUM_VERSION_FOR_EXTERNAL,
+                working_directory=self._working_directory,
+                subprocess_esbuild=self._subprocess_esbuild,
+            )
```
torresxb1, line 86: This used to be a separate action. Since we need to check the version every time that we want to mark things as `--external` then I think we should do the check when/before marking things as external (here), instead of having a separate action that would always have to be appended before this one. This cleans things up a bit, but it's also safer as we don't have to remember that we need a check action before this bundle action.

aws_lambda_builders/workflows/nodejs_npm_esbuild/actions.py
```diff
@@ -103,65 +109,62 @@ def _should_bundle_deps_externally(self) -> bool:
         return self._skip_deps or "./node_modules/*" in self._bundler_config.get(EXTERNAL_KEY, [])
 
 
-class EsbuildCheckVersionAction(BaseAction):
-    """
-    A Lambda Builder Action that verifies that esbuild is a version supported by sam accelerate
+def check_minimum_esbuild_version(
```
torresxb1, line 112: the logic of the version checking did not change

aws_lambda_builders/workflows/nodejs_npm_esbuild/workflow.py
```diff
@@ -40,7 +37,7 @@ class NodejsNpmEsbuildWorkflow(BaseWorkflow):
 
     CAPABILITY = Capability(language="nodejs", dependency_manager="npm-esbuild", application_framework=None)
 
-    EXCLUDED_FILES = (".aws-sam", ".git")
+    EXCLUDED_FILES = (".aws-sam", ".git", "node_modules")
```
torresxb1, line 40: we used to concatenate node_modules when excluding, so I thought it might make more sense to just add it here if we always exclude it.

aws_lambda_builders/workflows/nodejs_npm_esbuild/workflow.py
```diff
@@ -5,20 +5,17 @@
 import logging
```
torresxb1, line 5: This file changed quite a bit. Might be easier to open the new and the old file in different windows (without the file diffs), and comparing.

aws_lambda_builders/workflows/nodejs_npm_esbuild/workflow.py
```diff
@@ -53,150 +50,75 @@ def __init__(self, source_dir, artifacts_dir, scratch_dir, manifest_path, runtim
...
-            ),
-        ]
-        esbuild_with_deps = EsbuildBundleAction(
-            scratch_dir, artifacts_dir, bundler_config, osutils, subprocess_esbuild, self.manifest_path
-        )
+        if not self.download_dependencies and not self.dependencies_dir:
+            # Invalid workflow, can't have no dependency dir and no installation
+            raise EsbuildExecutionError(
+                message="Lambda Builders encountered an invalid workflow. A workflow can't "
+                "include a dependencies directory without installing dependencies."
```
torresxb1, line 77: I just moved this around, didn't write the message myself, but I wanted to point out that it seems a bit unclear/misleading (the code comment above would be more accurate). 
The thing is that we need to either install dependencies or you need to provide us with a directory of where your dependencies are stored.

aws_lambda_builders/workflows/nodejs_npm_esbuild/workflow.py
```diff
@@ -53,150 +50,75 @@ def __init__(self, source_dir, artifacts_dir, scratch_dir, manifest_path, runtim
...
+            )
 
-        install_action = NodejsNpmWorkflow.get_install_action(
-            source_dir,
-            scratch_dir,
-            subprocess_npm,
-            osutils,
-            self.options,
+        self.actions.append(
+            CopySourceAction(source_dir=self.source_dir, dest_dir=self.scratch_dir, excludes=self.EXCLUDED_FILES)
```
torresxb1, line 81: I think using named keywords can be more readable in cases where it's not immediately obvious what the parameter(s) would be for. E.g. here we would immediately know which folder we copy from and which one we copy to, without having to check the CopySourceAction definition.

aws_lambda_builders/workflows/nodejs_npm_esbuild/workflow.py
```diff
@@ -53,150 +50,75 @@ def __init__(self, source_dir, artifacts_dir, scratch_dir, manifest_path, runtim
...
-        return self._accelerate_workflow_actions(
-            source_dir, scratch_dir, actions, install_action, esbuild_with_deps, esbuild_no_deps
+        bundle_action = EsbuildBundleAction(
+            working_directory=self.scratch_dir,
+            output_directory=self.artifacts_dir,
+            bundler_config=bundler_config,
+            osutils=self.osutils,
+            subprocess_esbuild=self.subprocess_esbuild,
+            manifest=self.manifest_path,
+            skip_deps=not self.combine_dependencies,
```
torresxb1, line 91: `combine dependencies` determines whether we should bundle dependencies or not

aws_lambda_builders/workflows/nodejs_npm_esbuild/workflow.py
```diff
@@ -53,150 +50,75 @@ def __init__(self, source_dir, artifacts_dir, scratch_dir, manifest_path, runtim
...
-        :type install_action: BaseAction
-        :param install_action: Installation action for npm
-
-        :type esbuild_with_deps: BaseAction
-        :param esbuild_with_deps: Standard esbuild action bundling source with deps
+        if self.download_dependencies:
+            self.actions += [
+                NodejsNpmWorkflow.get_install_action(
+                    source_dir=source_dir,
+                    artifacts_dir=self.scratch_dir,
```
torresxb1, line 98: it would be better if this argument was named install_dir, as it actually controls where dependencies are installed. Will rename in another PR, when I introduce building-in-source to esbuild

tests/unit/workflows/nodejs_npm_esbuild/test_workflow.py
```diff
@@ -244,11 +244,10 @@ def test_workflow_sets_up_esbuild_actions_without_download_dependencies_with_dep
             experimental_flags=[],
         )
 
-        self.assertEqual(len(workflow.actions), 4)
+        self.assertEqual(len(workflow.actions), 3)
```
torresxb1, line 247: a few tests will have one action less because I moved the action to check esbuild version into the bundling action itself.

tests/unit/workflows/nodejs_npm_esbuild/test_workflow.py
```diff
@@ -269,8 +268,8 @@ def test_workflow_sets_up_esbuild_actions_with_download_dependencies_and_depende
 
         self.assertIsInstance(workflow.actions[0], CopySourceAction)
         self.assertIsInstance(workflow.actions[1], NodejsNpmInstallAction)
-        self.assertIsInstance(workflow.actions[2], CleanUpAction)
-        self.assertIsInstance(workflow.actions[3], EsbuildBundleAction)
+        self.assertIsInstance(workflow.actions[2], EsbuildBundleAction)
+        self.assertIsInstance(workflow.actions[3], CleanUpAction)
```
torresxb1, line 272: the order did change, to simplify the code logic. There should be no impact in the end-result, however. We just need to clean up the dependencies dir before we move the updated dependencies in there, but the bundling itself happens on another folder.

aws_lambda_builders/workflows/nodejs_npm_esbuild/actions.py
mndeveci: I don't understand the benefit of converting this from an `Action` to its own method. Lambda Builders written in this way where each step in build process is defined by actions, and actions are defined during workflow creation step. Once they are created, they can be executed in order. I feel like this is somewhat against the general design of this library.

aws_lambda_builders/workflows/nodejs_npm_esbuild/workflow.py
```diff
@@ -53,150 +50,75 @@ def __init__(self, source_dir, artifacts_dir, scratch_dir, manifest_path, runtim
...
+                    osutils=self.osutils,
+                    build_options=self.options,
+                ),
+                bundle_action,
+            ]
 
-        :type esbuild_no_deps: List[BaseAction]
-        :param esbuild_no_deps: esbuild action not including dependencies in the bundled artifacts
+        if not self.dependencies_dir:
+            # we need a dependencies dir for the logic below
```
mndeveci, line 107: This error is moved above, so do we still need to keep this `if` statement?

mndeveci (changes requested):
Thanks for the effort! Left some comments but will definitely do another round.

aws_lambda_builders/workflows/nodejs_npm_esbuild/actions.py
torresxb1: if two actions are such that they **always** need to be appended together (and in a certain order), they probably don't need to be separate actions. There is no case when the bundle action with the `--external` flag should be used without first having to check that their esbuild version supports it. In my opinion it is safer (we can't miss it), and cleaner, to have this check happen on the bundle action rather than as a separate action. 
I hope this doesn't go against the design of the library. I view bundling as **the** action, while the check is just something that needs to happen as part of the bundling.

aws_lambda_builders/workflows/nodejs_npm_esbuild/workflow.py
torresxb1: I didn't find any in the esbuild workflow. The excludes was used in one place, and it appended the node modules every time.

aws_lambda_builders/workflows/nodejs_npm_esbuild/actions.py
```diff
@@ -12,6 +12,7 @@
 LOG = logging.getLogger(__name__)
 
 EXTERNAL_KEY = "external"
+MINIMUM_VERSION_FOR_EXTERNAL = "0.14.13"
```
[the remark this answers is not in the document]
torresxb1: It's the esbuild version needed to support the `--external` flag to skip bundling some dependencies. It was already in the code, but I moved it to a constant.

aws_lambda_builders/workflows/nodejs_npm_esbuild/workflow.py
```diff
@@ -53,150 +50,75 @@ def __init__(self, source_dir, artifacts_dir, scratch_dir, manifest_path, runtim
...
+        self.subprocess_esbuild = self._get_esbuild_subprocess()
 
         bundler_config = self.get_build_properties()
 
-        if not osutils.file_exists(manifest_path):
-            LOG.warning("package.json file not found. Bundling source without dependencies.")
-            self.actions = [
+        if not self.osutils.file_exists(manifest_path):
+            LOG.warning("package.json file not found. Bundling source without installing dependencies.")
+            self.actions.append(
```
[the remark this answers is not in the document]
torresxb1: There shouldn't be any changes, and the successful tests hopefully reflect that. But I can change it back as I don't have a preference here.

aws_lambda_builders/workflows/nodejs_npm_esbuild/workflow.py
torresxb1: Yep, we still need to check because `self.download_dependencies` could be true to avoid the error above (but we still need to check that we have a dependencies dir before going to the logic below)

aws_lambda_builders/workflows/nodejs_npm_esbuild/actions.py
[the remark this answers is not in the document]
torresxb1: done

aws_lambda_builders/workflows/nodejs_npm_esbuild/workflow.py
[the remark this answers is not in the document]
torresxb1: done

aws_lambda_builders/workflows/nodejs_npm_esbuild/workflow.py
```diff
@@ -53,150 +50,75 @@ def __init__(self, source_dir, artifacts_dir, scratch_dir, manifest_path, runtim
...
-                raise EsbuildExecutionError(
-                    message="Lambda Builders encountered an invalid workflow. A workflow can't "
-                    "include a dependencies directory without installing dependencies."
-                )
-
-        return actions
+            # if we're reusing dependencies, then we need to symlink them before bundling
+            self.actions += [
+                LinkSourceAction(self.dependencies_dir, self.scratch_dir),
+                bundle_action,
```
mildaniel, line 120: Seems that the bundle action is already included in line 103, so this would run the bundler twice.

aws_lambda_builders/workflows/nodejs_npm_esbuild/workflow.py
mildaniel: That's right, we need to find the dependencies somewhere in this case. Suggested re-wording:
`Lambda Builders was unable to find the location of the dependencies since a dependencies directory was not provided and download dependencies is disabled.`

mildaniel (commented):
Thanks Rup, this is a great QoL change. Left a couple small comments but generally agree with the direction you took combining the bundle action and version check action since they are always bundled together.

aws_lambda_builders/workflows/nodejs_npm_esbuild/workflow.py
torresxb1: This is one part that can lead to confusion and wasn't able to figure out how to make it better. It would never bundle twice because this bundle on this line happens when `self.download_dependencies` is `False`, while the one in line 103 happens when `self.download_dependencies` is `True`

torresxb1: in other words, if we're reusing dependencies, then we didn't download them and we should first symlink and then bundle. If we downloaded dependencies, then we should bundle after downloading, (and we won't symlink or bundle again)

mildaniel: I'm seeing it if it's true for both cases:
```
            self.actions += [
                NodejsNpmWorkflow.get_install_action(
                    source_dir=source_dir,
                    artifacts_dir=self.scratch_dir,
                    subprocess_npm=self.subprocess_npm,
                    osutils=self.osutils,
                    build_options=self.options,
                ),
                bundle_action,
            ]

        if not self.dependencies_dir:
            # we need a dependencies dir for the logic below
            return

            # if we downloaded dependencies, we have to update dependencies_dir
            self.actions += [
                CleanUpAction(self.dependencies_dir),
                MoveDependenciesAction(self.source_dir, self.scratch_dir, self.dependencies_dir),
            ]
        else:
            # if we're reusing dependencies, then we need to symlink them before bundling
            self.actions += [
                LinkSourceAction(self.dependencies_dir, self.scratch_dir),
                bundle_action,
            ]
```

torresxb1: I think you're indentation might be off? 😅 . The first if in the code block you're posting should be in the same level as the other two if's below

torresxb1: or otherwise not sure what you mean

torresxb1: I just made an additional change related to this, hopefully it's clearer now.

aws_lambda_builders/workflows/nodejs_npm_esbuild/actions.py
```diff
@@ -27,20 +29,20 @@ class EsbuildBundleAction(BaseAction):
...
         bundler_config: Dict[str, Any],
         osutils: OSUtils,
         subprocess_esbuild: SubprocessEsbuild,
         manifest: str,
         skip_deps=False,
     ):
         """
-        :type scratch_dir: str
-        :param scratch_dir: an existing (writable) directory for temporary files
+        :type working_directory: str
```
mndeveci, line 41: nit: can we also update doc formats to numpy format that you added above (https://www.dataquest.io/blog/documenting-in-python-with-docstrings/)

aws_lambda_builders/workflows/nodejs_npm_esbuild/workflow.py
torresxb1: done

aws_lambda_builders/workflows/nodejs_npm_esbuild/actions.py
mndeveci: > if two actions are such that they always need to be appended together (and in a certain order)

I agree but then they should be combined as single action, if we won't need them individually.

Regarding the rest of your comment; this library can also be designed in a way that each action can be its own method, and workflows can call those methods depending on the parameters it has (like if it needs to download dependencies, or need to combine them etc.). But it will be hard to test the behavior with certain workflows, since we may need to manually check if a function is called or not. In current way, we can easily confirm behaviors in the tests by just looking at the order of the actions.  And confirming the behavior of the actions is easy, because they are isolated into their own generic classes.

This one change might be OK, but I want us to be careful about making this as an habit and move away from the 'Actions'.

torresxb1:
@mndeveci added tests to increase coverage and changed to numpy docstrings

aws_lambda_builders/workflows/nodejs_npm_esbuild/actions.py
torresxb1: @mndeveci and I discussed this offline and it seems like there was a misunderstanding, and we're good with this change. The two actions were indeed combined into a single action.

mndeveci (approved):
Thanks for the effort that you put here @torresxb1! Approving the changes since we confirm that expected actions are been setup with possible combinations through tests.

mildaniel (approved):

PR #435, opened 2023-01-30
chore: Move to ruff from pylint

jfuss:
When we started the project we defaulted to use pylint. Pylint has served it's purpose but it pretty slow. Ruff is a newer linter in the python ecosystem but is written in Rust. This makes Ruff was faster than pylint. On my machine (while testing in SAM CLI), pylint took about 70s to lint the repo but with ruff it took .04s.

*Issue #, if available:*
N/A

*Description of changes:*
See https://github.com/charliermarsh/ruff for more details.


moelasmar (approved):

pyproject.toml
```diff
@@ -1,3 +1,10 @@
+[tool.ruff]
+line-length = 120
+
+[tool.ruff.per-file-ignores]
+"__init__.py" = ["F401", "E501"]
+"aws_lambda_builders/workflow.py" = ["E501"]
```
torresxb1, line 6: why ignore E501 on these files?

torresxb1 (approved):

pyproject.toml
sriram-mv: I believe we have higher line lengths allowed by default. This is for an older rule on 79 chars per line.

sriram-mv (approved):

pyproject.toml
torresxb1: 🤔 I thought specifying the line length above would make this rule accommodate to that. If this rule still listens to 79 chars, why would we only ignore it for these files instead of everything?

jfuss: The workflow.py had two lines that I couldn’t find a good way to resolve. So elected to just ignore it. When black runs, everything is fixed anyways.

For ‘__init__.py’, they just do imports so it wasn’t required but was how some other repos did it. Black statement still holds true. 

Our line lengths are 120 as well.

PR #436, opened 2023-01-31
chore: Enable pylint within ruff

jfuss:
*Issue #, if available:*
N/A

*Description of changes:*
In #435, I added `ruff` to replace `pylint` linting. However, I missed that `ruff` does not add `pylint` in by default (only pycodestyle and Pyflakes. This PR targets `pylint` rules ruff has.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,6 +6,7 @@ select = [
     "F",  # Pyflakes
     "PL", # pylint
 ]
+ignore = ["PLR0913"]
 
 [tool.ruff.per-file-ignores]
 "__init__.py" = ["F401", "E501"]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,12 @@
 [tool.ruff]
 line-length = 120
 
+select = [
+    "E",  # Pycodestyle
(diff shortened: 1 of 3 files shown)


moelasmar (approved):

pyproject.toml
```diff
@@ -1,6 +1,13 @@
 [tool.ruff]
 line-length = 120
 
+select = [
+    "E",  # Pycodestyle
+    "F",  # Pyflakes
+    "PL", # pylint
+]
+ignore = ["PLR0913"]
```
mndeveci, line 9: If this is rule is broken in 1-2 places, should we add this inline rather than for the whole project here?

mndeveci (approved):
Left a non-blocking comment, overall it looks good.

pyproject.toml
jfuss: We explicitly disabled this in pylint: https://github.com/aws/aws-lambda-builders/blob/02b03aa91bc19a83f215e48c0b8a99c7a154afe5/.pylintrc#L64

We can but not sure if that there is a ton of value here. This is for the max args into a function.

PR #437, opened 2023-02-02
feat: use build_dir in esbuild workflow to support building in source

torresxb1:
*Issue #, if available:*

*Description of changes:*
Now that `build_dir` will point to either the source directory, if building in source, or the workflow's default build directory, we can refactor the workflow to use `build_dir` and it will enable build-in-source support.

### DIFF
--- a/aws_lambda_builders/workflows/nodejs_npm/actions.py
+++ b/aws_lambda_builders/workflows/nodejs_npm/actions.py
@@ -81,11 +81,10 @@ class NodejsNpmInstallAction(BaseAction):
     DESCRIPTION = "Installing dependencies from NPM"
     PURPOSE = Purpose.RESOLVE_DEPENDENCIES
 
-    def __init__(self, artifacts_dir, subprocess_npm):
+    def __init__(self, install_dir, subprocess_npm):
         """
-        :type artifacts_dir: str
-        :param artifacts_dir: an existing (writable) directory with project source files.
-            Dependencies will be installed in this directory.
+        :type install_dir: str
+        :param install_dir: Dependencies will be installed in this directory.
 
         :type subprocess_npm: aws_lambda_builders.workflows.nodejs_npm.npm.SubprocessNpm
         :param subprocess_npm: An instance of the NPM process wrapper
@@ -95,7 +94,7 @@ class NodejsNpmInstallAction(BaseAction):
         """
 
         super(NodejsNpmInstallAction, self).__init__()
-        self.artifacts_dir = artifacts_dir
+        self.install_dir = install_dir
         self.subprocess_npm = subprocess_npm
 
     def execute(self):
@@ -105,10 +104,10 @@ class NodejsNpmInstallAction(BaseAction):
         :raises lambda_builders.actions.ActionFailedError: when NPM execution fails
         """
         try:
-            LOG.debug("NODEJS installing in: %s", self.artifacts_dir)
+            LOG.debug("NODEJS installing in: %s", self.install_dir)
 
             self.subprocess_npm.run(
-                ["install", "-q", "--no-audit", "--no-save", "--unsafe-perm", "--production"], cwd=self.artifacts_dir
+                ["install", "-q", "--no-audit", "--no-save", "--unsafe-perm", "--production"], cwd=self.install_dir
             )
 
         except NpmExecutionError as ex:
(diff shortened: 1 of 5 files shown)


aws_lambda_builders/workflows/nodejs_npm/actions.py
```diff
@@ -81,11 +81,10 @@ class NodejsNpmInstallAction(BaseAction):
     DESCRIPTION = "Installing dependencies from NPM"
     PURPOSE = Purpose.RESOLVE_DEPENDENCIES
 
-    def __init__(self, artifacts_dir, subprocess_npm):
+    def __init__(self, install_dir, subprocess_npm):
```
torresxb1, line 84: renamed to generalize (e.g. now we can install in the source directory as well, so this should not be specific to the artifacts dir)

aws_lambda_builders/workflows/nodejs_npm_esbuild/workflow.py
```diff
@@ -77,23 +77,26 @@ def __init__(self, source_dir, artifacts_dir, scratch_dir, manifest_path, runtim
...
+            if self.build_dir == self.source_dir
+            else [CopySourceAction(source_dir=self.source_dir, dest_dir=self.build_dir, excludes=self.EXCLUDED_FILES)]
+        )
 
         if self.download_dependencies:
             self.actions.append(
                 NodejsNpmWorkflow.get_install_action(
                     source_dir=source_dir,
-                    artifacts_dir=self.scratch_dir,
+                    install_dir=self.build_dir,
```
moelasmar, line 91: I am not sure if I miss something or not, but I am not able to find the build_dir change in this class

moelasmar: never mind, I found it in the `BaseWorkflow` class.

moelasmar (approved):
It looks good to me, thanks Rup. I just have a question for the integration test case, my assumption is building in source should not be different than build in the scratch directory. If this assumption is correct, can we update our integration test cases to run in both source and scratch and make sure that it is working fine in both instead of having a new test case for build in source

lucashuy (approved):

sriram-mv:
as @moelasmar suggested, we could parameterize, but I'm not blocking on that.

sriram-mv (approved):

aws_lambda_builders/workflows/nodejs_npm/actions.py
sriram-mv: Something to consider when changing the arguments, let's make a callout to not changing the RPC protocol version because we did not touch the workflow args. https://github.com/aws/aws-lambda-builders/blob/develop/aws_lambda_builders/__init__.py#L8

If not considered, we should think of adding that into PR template so that we ask ourselves that question.

torresxb1: actually, _when_ should we change the RPC protocol version? Is it when we change any of the params passed in the request [here](https://github.com/aws/aws-lambda-builders/blob/develop/aws_lambda_builders/__main__.py#L88)? Or is it only on breaking changes for the params there?

torresxb1:
Hmm I see what you guys are saying, but the behaviour is slightly different when building in source (no copy operation, no dealing with `dependencies_dir` for accelerate's incremental build). Moreover, IMO it seems a bit overkill to parameterize all the integration tests whenever we add a new parameter for the workflow 🤔

moelasmar:
> Hmm I see what you guys are saying, but the behaviour is slightly different when building in source (no copy operation, no dealing with `dependencies_dir` for accelerate's incremental build) so I don't think we could just parameterize. Moreover, IMO it seems a bit overkill to parameterize all the integration tests.

my point is we need to make sure that all the cases we support in `build in scratch` case should work on the `build in source` case, that is why I asked to extend the current integration test cases to be built in source as well. If not possible to parameterize the current integration test cases, may be we need to duplicate them to test the build in source.

torresxb1:
@moelasmar yep, I get what you're saying. It just seems weird to me that because we added a new parameter, we'll have to test all combinations of it with all of the existing tests. It seems a bit much to me 🤔

torresxb1:
Discussed with the team, merging for now to close the task. If we feel like we need more tests, or need a refactoring of our integration tests, we can create another task/PR for this.

PR #458, opened 2023-02-21
refactor: simplify node npm workflow conditionals

torresxb1:
*Issue #, if available:*

*Description of changes:*
Currently there are a few nested conditions in the workflow. This PR refactors to remove some nesting and try to simplify the logic.

### DIFF
--- a/aws_lambda_builders/workflows/nodejs_npm/workflow.py
+++ b/aws_lambda_builders/workflows/nodejs_npm/workflow.py
@@ -60,35 +60,6 @@ class NodejsNpmWorkflow(BaseWorkflow):
 
         subprocess_npm = SubprocessNpm(osutils)
 
-        self.actions = self.actions_without_bundler(
-            source_dir, artifacts_dir, scratch_dir, manifest_path, osutils, subprocess_npm
-        )
-
-    def actions_without_bundler(self, source_dir, artifacts_dir, scratch_dir, manifest_path, osutils, subprocess_npm):
-        """
-        Generate a list of Nodejs build actions without a bundler
-
-        :type source_dir: str
-        :param source_dir: an existing (readable) directory containing source files
-
-        :type artifacts_dir: str
-        :param artifacts_dir: an existing (writable) directory where to store the output.
-
-        :type scratch_dir: str
-        :param scratch_dir: an existing (writable) directory for temporary files
-
-        :type manifest_path: str
-        :param manifest_path: path to package.json of an NPM project with the source to pack
-
-        :type osutils: aws_lambda_builders.workflows.nodejs_npm.utils.OSUtils
-        :param osutils: An instance of OS Utilities for file manipulation
-
-        :type subprocess_npm: aws_lambda_builders.workflows.nodejs_npm.npm.SubprocessNpm
-        :param subprocess_npm: An instance of the NPM process wrapper
-
-        :rtype: list
-        :return: List of build actions to execute
-        """
         tar_dest_dir = osutils.joinpath(scratch_dir, "unpacked")
         tar_package_dir = osutils.joinpath(tar_dest_dir, "package")
         npm_pack = NodejsNpmPackAction(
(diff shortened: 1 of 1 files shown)


aws_lambda_builders/workflows/nodejs_npm/workflow.py
```diff
@@ -60,35 +60,6 @@ def __init__(self, source_dir, artifacts_dir, scratch_dir, manifest_path, runtim
 
         subprocess_npm = SubprocessNpm(osutils)
 
-        self.actions = self.actions_without_bundler(
-            source_dir, artifacts_dir, scratch_dir, manifest_path, osutils, subprocess_npm
-        )
-
-    def actions_without_bundler(self, source_dir, artifacts_dir, scratch_dir, manifest_path, osutils, subprocess_npm):
```
torresxb1, line 67: this was originally meant for when the workflow would have `actions_with_bundler` (for esbuild) and `actions_without_bundler`. But the bundler actions were split into a separate workflow (esbuild), so there's not much use for having `actions_without_bundler` in a separate function

aws_lambda_builders/workflows/nodejs_npm/workflow.py
```diff
@@ -97,48 +68,46 @@ def actions_without_bundler(self, source_dir, artifacts_dir, scratch_dir, manife
...
-        else:
-            # if dependencies folder exists and not download dependencies, simply copy the dependencies from the
-            # dependencies folder to artifact folder
-            if self.dependencies_dir and self.combine_dependencies:
-                actions.append(CopySourceAction(self.dependencies_dir, artifacts_dir))
-            else:
-                LOG.info(
-                    "download_dependencies is False and dependencies_dir is None. Copying the source files into the "
-                    "artifacts directory. "
-                )
```
torresxb1, line 133: I thought this message was a bit weird and not totally accurate, so I removed it. The message could be printed if `dependencies_dir` is **not** None, but `combine_dependencies` is False. Let me know if you think I should bring this message back.

torresxb1: Note that the conditions that lead to this message being output is not a valid SAM CLI use case anyway (shouldn't be very common): 
https://github.com/aws/aws-lambda-builders/pull/272/files#r701280071

aws_lambda_builders/workflows/nodejs_npm/workflow.py
```diff
@@ -97,48 +68,46 @@ def actions_without_bundler(self, source_dir, artifacts_dir, scratch_dir, manife
...
+            NodejsNpmLockFileCleanUpAction(artifacts_dir, osutils=osutils),
+        ]
+
+        # if no dependencies dir, just cleanup artifacts and we're done
+        if not self.dependencies_dir:
+            self.actions += artifacts_cleanup_actions
+            return
+
+        # if we downloaded dependencies, update dependencies_dir
+        if self.download_dependencies:
```
sriram-mv, line 94: are we branching on the same condition twice?

sriram-mv: One other general comment is we have more than a few branches, we should go down a handler route where we mutate on a a given object all the way through.

eg: self.actions is the one we mutate in this workflow, something like below would be useful to grok through it.


```python
for action_mapper in action_mappers:
    action_mapper.apply(*args, **kwargs)
```

sriram-mv (changes requested):

aws_lambda_builders/workflows/nodejs_npm/workflow.py
torresxb1: > are we branching on the same condition twice?
The first time it doesn't depend on `self.dependencies_dir`. The second time we can assume we have `self.dependencies_dir` (since we checked it before).

sriram-mv: [L77](https://github.com/aws/aws-lambda-builders/blob/342357c66520a79154cccf5f856c8f763c30f906/aws_lambda_builders/workflows/nodejs_npm/workflow.py#L77) and [L99](https://github.com/aws/aws-lambda-builders/blob/342357c66520a79154cccf5f856c8f763c30f906/aws_lambda_builders/workflows/nodejs_npm/workflow.py#L94)

the conditions are the same, what am I missing?

torresxb1: this part in the middle: https://github.com/aws/aws-lambda-builders/blob/342357c66520a79154cccf5f856c8f763c30f906/aws_lambda_builders/workflows/nodejs_npm/workflow.py#L89-L91

torresxb1: The first time it doesn't care about, or depend on, `self.dependencies_dir`. The second time, since we have checked `dependencies_dir`, we can assume we have it and we don't have to check for it again. Otherwise it would require some conditional nesting or repetitive conditions, which is what I'm trying to refactor out here 😅

hnnasit (approved):

aws_lambda_builders/workflows/nodejs_npm/workflow.py
sriram-mv: to me this is not intuitive and requires a further refactor.

https://github.com/aws/aws-lambda-builders/blob/342357c66520a79154cccf5f856c8f763c30f906/aws_lambda_builders/workflows/nodejs_npm/workflow.py#L109 and https://github.com/aws/aws-lambda-builders/blob/342357c66520a79154cccf5f856c8f763c30f906/aws_lambda_builders/workflows/nodejs_npm/workflow.py#L90 is also repeated.

torresxb1: ^ the above is also due to something similar. One is performed when there's no dependencies dir and one is performed when there is.

To me the refactor is easier to read than before, but let me know if you disagree. Also open to suggestions on how to further refactor.

sriram-mv (approved):

torresxb1:
Created an internal ticket to suggest further, bigger, refactoring improvements. 
Merging this one.

PR #462, opened 2023-03-01
feat: update node_npm actions and workflow to support building in source directory

torresxb1:
*Issue #, if available:*

*Description of changes:*
Updates node actions and workflow to support building in source directory. Seems like a big PR but most is tests and changing comments to numpy docstring format. Tests will fail until this other PR is merged: https://github.com/aws/aws-lambda-builders/pull/461


aws_lambda_builders/workflows/nodejs_npm/actions.py
```diff
@@ -151,7 +159,11 @@ def execute(self):
         try:
             LOG.debug("NODEJS installing ci in: %s", self.install_dir)
 
-            self.subprocess_npm.run(["ci"], cwd=self.install_dir)
+            command = ["ci"]
+            if self.install_links:
+                command.append("--install-links")
+
+            self.subprocess_npm.run(command, cwd=self.install_dir)
```
torresxb1, line 166: I've realized the CI action maybe should've just been a subclass of the normal install action, and in general I'm not sure why we don't use the same flags. But it might be too late to change now.

aws_lambda_builders/workflows/nodejs_npm/workflow.py
```diff
@@ -75,39 +81,82 @@ def __init__(self, source_dir, artifacts_dir, scratch_dir, manifest_path, runtim
...
-            # install the dependencies into artifact folder
             self.actions.append(
-                NodejsNpmWorkflow.get_install_action(source_dir, artifacts_dir, subprocess_npm, osutils, self.options)
+                NodejsNpmWorkflow.get_install_action(
+                    source_dir=source_dir,
+                    install_dir=self.build_dir,
+                    subprocess_npm=subprocess_npm,
+                    osutils=osutils,
+                    build_options=self.options,
+                    install_links=self.build_dir == self.source_dir,
```
torresxb1, line 91: We want to use install-links so npm installs local dependencies without any symbolic links and without dev dependencies or temporary files, but instead just like any other dependency.

aws_lambda_builders/workflows/nodejs_npm/actions.py
```diff
@@ -151,7 +159,11 @@ def execute(self):
         try:
             LOG.debug("NODEJS installing ci in: %s", self.install_dir)
 
-            self.subprocess_npm.run(["ci"], cwd=self.install_dir)
+            command = ["ci"]
+            if self.install_links:
+                command.append("--install-links")
```
sriram-mv, line 164: something to keep in mind, is that we are appending in certain places and doing a + operation in other places, let's default to one.

torresxb1: I don't think there's a problem with using both 🤔 . Appending to add one element, using "+" to concatenate two lists.

mildaniel: I don't have any strong opinion here, but we can also use `.extend()` to concatenate two lists if we feel that's cleaner moving forward, though I think `+=` is usually descriptive enough.

sriram-mv (approved):

PR #467, opened 2023-03-13
fix: Enable --manifest flag for esbuild

mildaniel:
*Issue #, if available:*
https://github.com/aws/aws-sam-cli/issues/4832

*Description of changes:*
Pass through the manifest path explicitly if it differs from the one in the code URI root directory. This allows the `--manifest` flag passed down to be different from the one in the root.

### DIFF
--- a/aws_lambda_builders/workflows/nodejs_npm_esbuild/workflow.py
+++ b/aws_lambda_builders/workflows/nodejs_npm_esbuild/workflow.py
@@ -5,8 +5,10 @@ NodeJS NPM Workflow using the esbuild bundler
 import json
 import logging
 from pathlib import Path
+from typing import List
 
 from aws_lambda_builders.actions import (
+    BaseAction,
     CleanUpAction,
     CopySourceAction,
     LinkSourceAction,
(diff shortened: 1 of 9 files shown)


aws_lambda_builders/workflows/nodejs_npm_esbuild/workflow.py
```diff
@@ -78,13 +80,15 @@ def __init__(self, source_dir, artifacts_dir, scratch_dir, manifest_path, runtim
...
-        self.actions = (
+        self.actions: List[BaseAction] = (
             []
             if self.build_dir == self.source_dir
             else [CopySourceAction(source_dir=self.source_dir, dest_dir=self.build_dir, excludes=self.EXCLUDED_FILES)]
         )
 
         if self.download_dependencies:
+            # Manifest is different from the one in the project directory
+            specify_manifest = str(Path(manifest_path).parent) != source_dir
```
mndeveci, line 91: Is `manifest_path` specified all the time, even without `--manifest` option?

mildaniel: Yes

mildaniel:
Seems like this approach breaks with the newest version of NPM as the installation behaviour is different. Going to close this PR for now and revisit with a different approach.

PR #492, opened 2023-04-28
chore: remove deprecated runtime dotnetcore3.1

moelasmar:
Removing the support for the deprecated runtime dotnetcore3.1


hawflau (approved):

hnnasit (commented):
Should we also remove the `WithDefaultsFile3.1` folder? Seems like it was used in `TestDotnet31` class but that is removed now. And dotnet tests use `WithDefaultsFile6` testdata.

tests/unit/workflows/dotnet_clipackage/test_actions.py
```diff
@@ -176,7 +176,7 @@ def test_build_package_arguments(self):
                 "--msbuild-parameters",
                 "--runtime linux-x64",
                 "--framework",
-                "netcoreapp2.1",
+                "net6.0",
```
sriram-mv, line 179: This example and the actual option don't match: https://github.com/aws/aws-lambda-builders/pull/492/files#diff-a33133aa75d3cbb0bf53e9c46bc3bb116a3f94fbd343933949990fe3c4eef5f1R44

should we get them to be the same?

hnnasit (approved):

PR #498, opened 2023-05-08
feat: Make esbuild command builder accept any flag

mildaniel:
*Issue #, if available:*
https://github.com/aws/aws-sam-cli/issues/4827

*Description of changes:*
Instead of having to manually support esbuild flags, determines the type of the flag provided and creates the corresponding command line argument for that flag type.


aws_lambda_builders/workflows/nodejs_npm_esbuild/esbuild.py
```diff
@@ -99,28 +99,7 @@ def run(self, args, cwd=None):
...
-    "sources_content",
-]
-
-# Multi-value types (--external:axios --external:aws-sdk)
-SUPPORTED_ESBUILD_APIS_MULTI_VALUE = [
-    "external",
-    "loader",
-    "out_extension",
-]
+NON_CONFIGURABLE_VALUES = ["bundle", "platform", "outdir", "entry_points"]
```
mndeveci, line 102: Are these the ones that is set by another field or by another parameter?

aws_lambda_builders/workflows/nodejs_npm_esbuild/esbuild.py
```diff
@@ -154,15 +133,86 @@ def build_esbuild_args_from_config(self) -> "EsbuildCommandBuilder":
         """
         args = []
 
-        args.extend(self._get_boolean_args())
-        args.extend(self._get_single_value_args())
-        args.extend(self._get_multi_value_args())
+        for config_key, config_value in self._bundler_config.items():
+            if config_key in NON_CONFIGURABLE_VALUES:
+                continue
```
mndeveci, line 138: Worth logging a warning message so that user will be aware that their setting is not applied as they wanted to. We might want to provide detailed steps for them to fix the issue on their side.

aws_lambda_builders/workflows/nodejs_npm_esbuild/esbuild.py
```diff
@@ -154,15 +133,86 @@ def build_esbuild_args_from_config(self) -> "EsbuildCommandBuilder":
...
+            configuration_type_callback = self._get_config_type_callback(config_value)
+            args.extend(configuration_type_callback(config_key, config_value))
```
mndeveci, line 140: Same as above, we might want to log these params so that it will be easy to track any issue in the future.

mndeveci (commented):
Looks good in general, and very good improvement. Left some comments for some additional logging there.

aws_lambda_builders/workflows/nodejs_npm_esbuild/esbuild.py
mildaniel: These are ones the we set ourselves explicitly for it to work the way we went in Lambda.

mndeveci (approved):

lucashuy (approved):

PR #501, opened 2023-05-11
Add function architecture argument when building dotnet function 

mrkdeng:
*Issue #, if available:*

*Description of changes:*
This change made:
1. To complete support for building ARM64 dotnet7 functions https://github.com/aws/aws-sam-build-images/pull/81. Pass down directly the `Architecture` parameter defined in SAM template to Amazon Lambda Tools via `--function-architecture` argument. Previously when building dotnet7 functions, users  also need to make changes in `aws-lambda-tools-defaults.json` in their source code directories.

### DIFF
--- a/tests/integration/workflows/dotnet_clipackage/testdata/CustomRuntime6/aws-lambda-tools-defaults.json
+++ b/tests/integration/workflows/dotnet_clipackage/testdata/CustomRuntime6/aws-lambda-tools-defaults.json
@@ -12,5 +12,5 @@
   "function-memory-size": 256,
   "function-timeout": 30,
   "function-handler": "bootstrap",
-  "msbuild-parameters": "--self-contained true"
+  "msbuild-parameters": "--self-contained false"
 }
--- a/tests/integration/workflows/dotnet_clipackage/testdata/CustomRuntime6/aws-lambda-tools-defaults.json
+++ b/tests/integration/workflows/dotnet_clipackage/testdata/CustomRuntime6/aws-lambda-tools-defaults.json
@@ -12,5 +12,5 @@
   "function-memory-size": 256,
   "function-timeout": 30,
   "function-handler": "bootstrap",
-  "msbuild-parameters": "--self-contained false"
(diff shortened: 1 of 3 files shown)


tests/integration/workflows/dotnet_clipackage/testdata/CustomRuntime6/aws-lambda-tools-defaults.json
```diff
@@ -12,5 +12,5 @@
   "function-memory-size": 256,
   "function-timeout": 30,
   "function-handler": "bootstrap",
-  "msbuild-parameters": "--self-contained true"
+  "msbuild-parameters": "--self-contained false"
```
mndeveci, line 15: Does this mean customers might need to change this configuration on their end?

mndeveci (commented):
Thanks for raising this PR!

Left a comment about this one, but if this requires changes in the customer's application, we need to find an alternative approach as this change becomes backwards incompatible.

tests/integration/workflows/dotnet_clipackage/testdata/CustomRuntime6/aws-lambda-tools-defaults.json
Beau-Gosse-dev: Yes. Because we were always overriding msbuild-parameters at the command line, these properties were never actually used. So basically, users couldn't use msbuild-parameters in this file before, but now they can. I think about this more as fixing a bug or limitation.

But it could cause backward compatibility issues if someone had random msbuild parameters in here that they didn't actually want because now those parameters will actually get applied. 

So if we'd rather keep the somewhat limiting functionality but ensure backwards compatibility, we can probably keep passing both msbuild parameters and the new function-architecture arguments to the command line. 

Would you like us to revert it back to always passing msbuild-parameters down?

hawflau (approved):
LGTM! Thanks for the contribution!

mndeveci (approved):

PR #510, opened 2023-06-05
fix: Ignore flags not relevant to esbuild

mildaniel:
*Issue #, if available:*
https://github.com/aws/aws-sam-build-images/issues/90
https://github.com/aws/aws-sam-cli/issues/5236

*Description of changes:*
Since making the change to accept any flags for esbuild, there is an issue with esbuild trying to use flags that are not related to it. This changes forces the esbuild command builder to ignore those unrelated flags.

### DIFF
--- a/aws_lambda_builders/workflows/nodejs_npm_esbuild/esbuild.py
+++ b/aws_lambda_builders/workflows/nodejs_npm_esbuild/esbuild.py
@@ -101,6 +101,10 @@ class SubprocessEsbuild(object):
 
 NON_CONFIGURABLE_VALUES = {"bundle", "platform", "outdir"}
 
+# Ignore the values below. These are options that Lambda Builders accepts for
+# Node.js related workflows, but are not relevant to esbuild itself.
+ESBUILD_IGNORE_VALUES = {"use_npm_ci"}
+
 
 class EsbuildCommandBuilder:
     ENTRY_POINTS = "entry_points"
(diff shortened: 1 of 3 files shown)


aws_lambda_builders/workflows/nodejs_npm_esbuild/esbuild.py
```diff
@@ -101,6 +101,10 @@ def run(self, args, cwd=None):
 
 NON_CONFIGURABLE_VALUES = {"bundle", "platform", "outdir"}
 
+# Ignore the values below. These are options that Lambda Builders accepts for
+# Node.js related workflows, but are not relevant to esbuild itself.
+ESBUILD_IGNORE_VALUES = {"use_npm_ci", "entry_points"}
```
mildaniel, line 106: I explored other ways of ignoring these types of unrelated values too, however didn't find anything reliable (e.g. there's no API to use, and parsing the esbuild help text is very complicated. We can go with this solution for now as it fixes the immediate issue, and look for ways of making this more reliable moving forward.

jfuss (approved):
Change seems reasonable to me.

Are we sure these are the only two (`use_npm_ci`, `entry_points`)?

mildaniel:
> Change seems reasonable to me.
> 
> Are we sure these are the only two (`use_npm_ci`, `entry_points`)?

Yup, should be just these.

sriram-mv (approved):

PR #523, opened 2023-07-13
fix: Use the fallback PKG-INFO file if the one generated manually is incomplete

lucashuy:
*Issue #, if available:*
#526 

*Description of changes:*
Lambda Builders will invoke the `setup.py` file of source packages to determine the name and version of a package. The PKG-INFO file generated could be incomplete (`UNKNOWN` name or `0.0.0` version). If either of these are found, then we'll fall back to the file that is generated by `pip download`.


aws_lambda_builders/workflows/python_pip/packager.py
```diff
@@ -620,13 +639,79 @@ def _unpack_sdist_into_dir(self, sdist_path, unpack_dir):
...
+            The version of the package as a string
+
+        Returns
+        -------
+        bool
+            True if either name or version are the default values
+        """
+        # default values logic:
+        # https://github.com/pypa/setuptools/blob/6083e18f4afc40316c0112134c205c336afbcdfd/setuptools/_distutils/dist.py#L1185-L1189
+        return name == "UNKNOWN" or version == "0.0.0"
```
jfuss, line 677: Should this be `and`? Doc string indicates `and`.

lucashuy: Hmm, thats a good call out. This should be an `or`, so the doc string should be updated. Reason why the version check is also included here is that the assumption is being made that the packages wouldn't have version `0.0.0`. In any case, if it did it would just grab the fall back file (the one `pip download` generates with the build deps) and use what it has.

aws_lambda_builders/workflows/python_pip/packager.py
```diff
@@ -620,13 +639,79 @@ def _unpack_sdist_into_dir(self, sdist_path, unpack_dir):
...
+
+            if self._osutils.file_exists(fallback_pkg_info_fp):
+                # use the fallback values instead of the ones we generated
+                fallback_name, fallback_version = self._get_name_version(fallback_pkg_info_fp)
+
+                if not self._is_default_setuptools_values(fallback_name, fallback_version):
+                    name = fallback_name
+                    version = fallback_version
+
         return name, version
```
hawflau, line 715: in case they are not the fallback name and version, what are they?

lucashuy: If the fallback name and version turn out to also be the default values, then it will use the the initially determined values. Basically, no replacement logic will happen.

hawflau (approved):

aws_lambda_builders/workflows/python_pip/packager.py
jfuss: Got it. Probably fine. I did a quick search and seems like most package versions start with `0.1.0` or `0.0.1`. I am sure there are packages with `0.0.0` but we can always handle those in one offs. 

The "safer" thing is probably doing a version check for `0.0.0` only if `UNKNOWN` is given but I think what you have is fine.

jfuss (approved):

PR #535, opened 2023-08-29
fix: Fix missing encoding when logging from Makefile

lucashuy:
*Issue #, if available:*
N/A

*Description of changes:*
Some of the output that may come from a Makefile (eg. Terraform output) can contain some colour and encoding information (eg. `b'\x1b[0m\x1b[1mInitializing provider plugins...\x1b[0m\n' `). This does not get logged properly in the terminal.

Changes are to use `sys.stderr.buffer.write()` to print these messages instead.

### DIFF
--- a/aws_lambda_builders/workflows/custom_make/make.py
+++ b/aws_lambda_builders/workflows/custom_make/make.py
@@ -3,7 +3,6 @@ Wrapper around calling make through a subprocess.
 """
 import io
 import logging
-import os
 import shutil
 import sys
 import threading
@@ -100,7 +99,7 @@ class SubProcessMake(object):
             #
             # stderr is used since stdout appears to be reserved
             # for command responses
-            sys.stderr.buffer.write(line.strip() + os.linesep.encode())
+            sys.stderr.buffer.write(line)
             sys.stderr.flush()
 
             # Gather total stdout
--- a/aws_lambda_builders/workflows/custom_make/make.py
+++ b/aws_lambda_builders/workflows/custom_make/make.py
@@ -4,6 +4,7 @@ Wrapper around calling make through a subprocess.
 import io
 import logging
 import shutil
+import sys
 import threading
 
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


aws_lambda_builders/workflows/custom_make/make.py
```diff
@@ -92,9 +93,11 @@ def run(self, args, env=None, cwd=None):
 
         # Log every stdout line by iterating
         for line in p.stdout:
-            decoded_line = line.decode("utf-8").strip()
-            LOG.info(decoded_line)
+            sys.stderr.buffer.write(line)
```
mildaniel, line 96: AFAIK, `sys.stderr.buffer.write` won't include a newline at the end of the string which might make the next log statement messy. Should we add a newline after?

lucashuy:
Good call out, Terraform output includes this newline however other outputs from commands may not. Will look into it

From: Daniel Mil ***@***.***>
Reply-To: aws/aws-lambda-builders ***@***.***>
Date: Tuesday, August 29, 2023 at 2:31 PM
To: aws/aws-lambda-builders ***@***.***>
Cc: "Huynh, Lucas" ***@***.***>, Author ***@***.***>
Subject: Re: [aws/aws-lambda-builders] fix: Fix missing encoding when logging from Makefile (PR #535)


@mildaniel commented on this pull request.

________________________________

In aws_lambda_builders/workflows/custom_make/make.py<https://github.com/aws/aws-lambda-builders/pull/535#discussion_r1309374488>:

> @@ -92,9 +93,11 @@ def run(self, args, env=None, cwd=None):



         # Log every stdout line by iterating

         for line in p.stdout:

-            decoded_line = line.decode("utf-8").strip()

-            LOG.info(decoded_line)

+            sys.stderr.buffer.write(line)

AFAIK, sys.stderr.buffer.write won't include a newline at the end of the string which might make the next log statement messy. Should we add a newline after?

aws_lambda_builders/workflows/custom_make/make.py
lucashuy: ref: https://docs.python.org/3/library/io.html#io.IOBase.readline

For the line that gets returned by the iterator (`for line in stdout`), it calls `readline()` under the hood and there should always be a new line character at the end of each line. So here, we should be fine to not include a new line of our own.

mildaniel: Awesome, thanks for digging in!

moelasmar: just can you add a comment why we used sys.stderr to log the terraform output instead of log.info.

moelasmar (approved):

PR #546, opened 2023-09-18
feat: Validate minimum npm version required to use --install-links

lucashuy:
*Issue #, if available:*
None.

*Description of changes:*
`--install-links` is a relatively new addition to `npm`. We'll need to add a check against the current `npm` version if build in source is enabled since we need to use that option to properly install local dependencies.


aws_lambda_builders/workflows/nodejs_npm/exceptions.py
```diff
@@ -0,0 +1,30 @@
...
+from aws_lambda_builders.exceptions import LambdaBuilderError
+
+
+class NpmExecutionError(Exception):
+    """
+    Exception raised in case NPM execution fails.
+    It will pass on the standard error output from the NPM console.
+    """
+
+    MESSAGE = "NPM Failed: {message}"
```
mndeveci, line 15: We may move `MESSAGE` above so we don't need to define in multiple places

aws_lambda_builders/workflows/nodejs_npm/workflow.py
```diff
@@ -24,11 +24,15 @@
     NodejsNpmrcAndLockfileCopyAction,
     NodejsNpmrcCleanUpAction,
 )
+from .exceptions import OldNpmVersionError
```
mndeveci, line 27: Can we use absolute imports (and update existing ones as well)?

aws_lambda_builders/workflows/nodejs_npm/workflow.py
```diff
@@ -220,6 +223,11 @@ def get_install_action(
         BaseAction
             Install action to use
         """
+        if install_links and not NodejsNpmWorkflow.can_use_install_links(subprocess_npm):
+            raise OldNpmVersionError(
```
mndeveci, line 227: Is there a way that we can fallback to not build in source behavior if this happens?

aws_lambda_builders/workflows/nodejs_npm/exceptions.py
```diff
@@ -0,0 +1,25 @@
...
+
+    def __init__(self, **kwargs):
+        Exception.__init__(self, self.MESSAGE.format(**kwargs))
+
+
+class OldNpmVersionError(NpmExecutionError):
```
mildaniel, line 21: nit: I think it might be more clear if it was something like
```suggestion
class UnsupportedNpmVersionError(NpmExecutionError):
```

aws_lambda_builders/workflows/nodejs_npm/workflow.py
lucashuy: Updated, now instead of hard failing, we log a warning and revert the build directory back to the workflow's default (artifacts for regular Node and scratch for esbuild).

aws_lambda_builders/workflows/nodejs_npm/workflow.py
```diff
@@ -23,14 +23,20 @@
...
 LOG = logging.getLogger(__name__)
 
 # npm>=8.8.0 supports --install-links
 MINIMUM_NPM_VERSION_INSTALL_LINKS = (8, 8)
+UNSUPPORTED_NPM_VERSION_MESSAGE = (
+    "Building in source was enabled, however the "
+    "currently installed npm version does not support "
+    "--install-links. Please ensure that the npm "
+    "version is at least 8.8.0. Switching to build "
+    "in a scratch directory."
```
mndeveci, line 38: Link to NPM documentation might be useful

mndeveci (approved):

aws_lambda_builders/workflows/nodejs_npm/workflow.py
lucashuy: The closest thing to docs that would indicate the minimum version would be this: https://docs.npmjs.com/cli/v8/using-npm/changelog#v880-2022-04-27

I can add this to the message. Additionally I noticed that I've left the message as "switching to scratch directory", which isn't 100% true for Node.js. I can change this to "switching to building outside of the source directory".

mildaniel (approved):

PR #549, opened 2023-09-26
fix: Validate if pip is installed in the runtime path

lucashuy:
*Issue #, if available:*
N/A.

*Description of changes:*
Previously, the Python runtime validator would only validate if the runtime contained the correct Python for the workflow. It did not validate whether or not this runtime had `pip` installed inside of it or not.

This lead to a case where in a list of Python runtimes, it would select the first one that had Python without checking for `pip`. The correct runtime in this case would be the one that contains both Python and `pip`.

This change updates the build action to find and validate all the possible Python execution paths to find one that contains the correct version of Python while also containing `pip`.

### DIFF
--- a/aws_lambda_builders/workflows/python_pip/validator.py
+++ b/aws_lambda_builders/workflows/python_pip/validator.py
@@ -9,6 +9,7 @@ import subprocess
 from aws_lambda_builders.exceptions import MisMatchRuntimeError
 from aws_lambda_builders.validator import RuntimeValidator
 from aws_lambda_builders.workflows.python_pip.compat import pip_import_string
+from aws_lambda_builders.workflows.python_pip.exceptions import MissingPipError
 
 from .utils import OSUtils
 
@@ -38,7 +39,8 @@ class PythonRuntimeValidator(RuntimeValidator):
         Raises
         ------
         MisMatchRuntimeError
-            Raise runtime is not support or runtime does not support architecture.
+            Raise runtime is not support or runtime does not support architecture or
+            if the Python runtime does not contain `pip`.
         """
 
         runtime_path = super(PythonRuntimeValidator, self).validate(runtime_path)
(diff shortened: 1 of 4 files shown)


moelasmar (approved):

aws_lambda_builders/workflows/python_pip/validator.py
```diff
@@ -48,11 +50,24 @@ def validate(self, runtime_path):
...
             raise MisMatchRuntimeError(language=self.language, required_runtime=self.runtime, runtime_path=runtime_path)
-        else:
-            self._valid_runtime_path = runtime_path
-            return self._valid_runtime_path
+
+        try:
+            # call method to import `pip`
+            # ignoring method return values since we only want to check
+            # if `pip` is imported successfully
+            pip_import_string(runtime_path)
```
moelasmar, line 61: I think it is better if we extract the part where we execute the python command to import pip in a separate function, and to have a function to check if pip is installed or not, but any way it is not a blocking

mndeveci (commented):
Thanks for addressing this issue. I have one general question, how this change will behave when customer tries to build a python lambda which doesn't have `requirements.txt` file?

aws_lambda_builders/workflows/python_pip/validator.py
lucashuy: Based on the below feedback (the "what-if" case), this will be moved out of this validator and into the build action so that this only finds the pip runtime if things are being built.

lucashuy:
> build a python lambda which doesn't have requirements.txt file
Thats a good call out. I'll move that validation logic out of the general pip validator and into the build action, that way this validation+setting of runtime will only apply to projects that the workflow has identified to build.

mildaniel (approved):
Thanks for this fix!

mndeveci (approved):

PR #550, opened 2023-09-27
fix: Update esbuild binary fetching logic to look inside of `build_dir`

lucashuy:
*Issue #, if available:*
N/A

*Description of changes:*
The workflow currently looks at the scratch directory to determine if esbuild is present or not. This is not the case for build in source however, as the Node modules are installed in the source directory, not the scratch directory.

This change updates the path to be `build_dir`, which is already set to the correct folder from earlier in the workflow.

### DIFF
--- a/aws_lambda_builders/workflows/nodejs_npm_esbuild/workflow.py
+++ b/aws_lambda_builders/workflows/nodejs_npm_esbuild/workflow.py
@@ -178,7 +178,7 @@ class NodejsNpmEsbuildWorkflow(BaseWorkflow):
 
     def _get_esbuild_subprocess(self) -> SubprocessEsbuild:
         try:
-            npm_bin_path_root = self.subprocess_npm.run(["root"], cwd=self.scratch_dir)
+            npm_bin_path_root = self.subprocess_npm.run(["root"], cwd=self.build_dir)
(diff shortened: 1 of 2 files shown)


mndeveci:
I see how this will work with build in source improvement, but will it still work with existing workflow which builds in scratch directory? cc: @mildaniel

moelasmar (approved):

lucashuy:
I'm not Daniel :), but it should still work for non build in source flows.

The esbuild workflow defines a [default building directory](https://github.com/aws/aws-lambda-builders/blob/31eafba209c04639a52d6299fb2a506bf66660be/aws_lambda_builders/workflows/nodejs_npm_esbuild/workflow.py#L47). When the workflow is initialized, it [sets the build directory](https://github.com/aws/aws-lambda-builders/blob/cbe58c295d0dd20558f349fd6666d1080a0feab5/aws_lambda_builders/workflow.py#L299) based on if build in source is enabled, and if the workflow supports it.

The unit test should cover this case (build in source enabled vs disabled)

mildaniel:
What Lucas said 😆 This change shouldn't impact the existing workflow at all

mildaniel (approved):

Issue #548, opened 2023-09-26
German Umlaut in Maven stdout causes SAM build to fail


InsertPikachuFace:
### Description:

Because the log output from the Maven build is decoded with UTF-8 ([see line](https://github.com/aws/aws-lambda-builders/blob/a01d058be95a64b8ffdb80ee51c7edaf783686b9/aws_lambda_builders/workflows/java_maven/maven.py#L31)), an "ü" in our unit-test logs causes the SAM local Lambda build to fail.
If I add -DskipTests with a Makefile it completes as expected.

### Steps to reproduce the issue:

1. Set up Lambda project that logs out an Umlaut (äÄöÖüÜß) during `maven package`
2. run `sam build` with default template configuration

### Observed result:
```
 | JavaMavenWorkflow:CopySource succeeded
 | Running JavaMavenWorkflow:MavenBuild
 | JavaMavenWorkflow:MavenBuild raised unhandled exception
Traceback (most recent call last):
  File "C:\Program Files\Amazon\AWSSAMCLI\runtime\lib\site-packages\aws_lambda_builders\workflow.py", line 369, in run
    action.execute()
  File "C:\Program Files\Amazon\AWSSAMCLI\runtime\lib\site-packages\aws_lambda_builders\workflows\java_maven\actions.py", line 39, in execute
    self.subprocess_maven.build(self.scratch_dir)
  File "C:\Program Files\Amazon\AWSSAMCLI\runtime\lib\site-packages\aws_lambda_builders\workflows\java_maven\maven.py", line 31, in build
    LOG.debug("Maven logs: %s", stdout.decode("utf8").strip())
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xfc in position 43840: invalid start byte
```

### Expected result:
SAM build completes despite an encoding error in a debug-log or it logs out any Maven output correctly.

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: Windows 10
4. SAM CLI version: 1.96.0
5. AWS region: EU-CENTRAL-1

jfuss:
Thanks for reporting. Going to mark this as a bug.

@InsertPikachuFace Is this something you would want to contribute a fix for? We understand if not, just will be some lead time before we can patch. We do prioritize bugs highly regardless just might be faster if you submitted a patch is all.

jfuss:
@InsertPikachuFace No worries at all. We appreciate you taking the time to report.

[...]

PR #554, opened 2023-10-06
fix: Decoding non-unicode characters raises unhandled exception

mildaniel:
*Issue #, if available:*
#548 
https://github.com/aws/aws-sam-cli/issues/2317

*Description of changes:*
There are times when Lambda Builders will attempt to decode command output that contains unsupported characters. In these events, the program will crash with an unhandled exception. Instead, complete the build and log a message explaining that the output couldn't be decoded.

### DIFF
--- a/aws_lambda_builders/workflows/dotnet_clipackage/dotnetcli.py
+++ b/aws_lambda_builders/workflows/dotnet_clipackage/dotnetcli.py
@@ -4,8 +4,8 @@ Wrapper around calls to dotent CLI through a subprocess.
 
 import logging
 
-from ...utils import decode
-from .utils import OSUtils
+from aws_lambda_builders.utils import decode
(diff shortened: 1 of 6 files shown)


sriram-mv (approved):

aws_lambda_builders/workflows/dotnet_clipackage/dotnetcli.py
```diff
@@ -2,9 +2,9 @@
 Wrapper around calls to dotent CLI through a subprocess.
 """
 
-import locale
 import logging
 
+from ...utils import decode
```
sriram-mv, line 7: nit: absolute imports.

sriram-mv (approved):

sriram-mv:
LGTM, though there are windows failures that need to be addressed.

mndeveci (approved):

PR #558, opened 2023-10-14
fix: Reference variable from loop instead of the one from the validator

lucashuy:
*Issue #, if available:*
None

*Description of changes:*
The recent change to update the Python path resolution to find the correct one if pip is required could reference a variable that has not yet been assigned.

This fix updates the two logs to use the path string from the loop instead of the one from the validator output.


mildaniel (approved):

mildaniel:
Thanks Lucas, I think we need to get this merged before the next SAM CLI release so that we don't break anyone.

mndeveci (approved):

PR #578, opened 2023-11-20
fix: Symlinks to non-existent directories, symlinks re-creating

mildaniel:
*Issue #, if available:*
N/A

*Description of changes:*
There are two scenarios fixed by this PR:
1. The symbolic link action attempts to link to a non-existent directory. This results in an error when executing `sam package`.
2. The symbolic link utility function attempts to link to an already created directory resulting in an error when using auto-dependency layer.

### DIFF
--- a/aws_lambda_builders/actions.py
+++ b/aws_lambda_builders/actions.py
@@ -160,7 +160,7 @@ class LinkSinglePathAction(BaseAction):
         destination_path = Path(self._dest)
         if not source_path.exists():
             # Source path doesn't exist, nothing to symlink
-            LOG.debug("Source path %s does not exist, skipping generating symlink")
+            LOG.debug("Source path %s does not exist, skipping generating symlink", source_path)
             return
         if not destination_path.exists():
             os.makedirs(destination_path.parent, exist_ok=True)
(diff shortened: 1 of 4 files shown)


aws_lambda_builders/actions.py
```diff
@@ -156,7 +156,11 @@ def __init__(self, source: Union[str, os.PathLike], dest: Union[str, os.PathLike
         self._dest = dest
 
     def execute(self):
+        source_path = Path(self._source)
         destination_path = Path(self._dest)
+        if not source_path.exists():
+            # Source path doesn't exist, nothing to symlink
+            return
```
mndeveci, line 163: Shouldn't this be the case when this action tries to create a symlink? This will try to create a symlink from `/foo/bar` to `/foo/baz`. In this case, `/foo/bar` (`source_path`) shouldn't exist but `/foo/baz` (`destination_path`) should already exist.

mildaniel: In this case the symlink would create `/foo/baz` and link it to `/foo/bar` even though `/foo/bar` doesn't exist. The symlink itself will be created successfully even though it's not linking to anything.

mndeveci (approved):

lucashuy (approved):

PR #579, opened 2023-11-23
fix: use npm update when building in source

mndeveci:
*Issue #, if available:*
N/A

*Description of changes:*
When `--build-in-source` is used, `npm install` command won't remove any dependency that is removed between 2 invocations.

This PR addresses that by adding `NodejsNpmUpdateAction` and using it when build-in-source is enabled which runs `npm --no-audit --no-save --unsafe-perm --production --no-package-lock --install-links`.

`--install-links` flag is also removed from `NodejsNpmInstallAction` which was only used for build in source workflow.

### DIFF
--- a/tests/integration/workflows/nodejs_npm/test_nodejs_npm.py
+++ b/tests/integration/workflows/nodejs_npm/test_nodejs_npm.py
@@ -338,7 +338,7 @@ class TestNodejsNpmWorkflow(TestCase):
         # update package.json with empty one and re-run the build then confirm node_modules are cleared up
         shutil.copy2(
             os.path.join(self.temp_testdata_dir, "no-deps", "package.json"),
-            os.path.join(self.temp_testdata_dir, "npm-deps", "package.json")
+            os.path.join(self.temp_testdata_dir, "npm-deps", "package.json"),
         )
 
         self.builder.build(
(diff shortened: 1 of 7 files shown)


tests/integration/workflows/nodejs_npm/test_nodejs_npm.py
```diff
@@ -315,6 +315,46 @@ def test_build_in_source_with_download_dependencies(self, runtime):
         output_files = set(os.listdir(self.artifacts_dir))
         self.assertEqual(expected_files, output_files)
 
+    @parameterized.expand([("nodejs12.x",), ("nodejs14.x",), ("nodejs16.x",), ("nodejs18.x",), ("nodejs20.x",)])
```
mildaniel, line 318: Do we still need node12 here?

mildaniel (approved):
Thanks Mehmet, LGTM. One very minor question.

tests/integration/workflows/nodejs_npm/test_nodejs_npm.py
mndeveci: I think we should leave it until it is been deprecated https://github.com/aws/aws-sam-cli/issues/6298

lucashuy (approved):
Sorry about the delay in review, thanks for making the changes!

PR #585, opened 2023-12-04
chore: auto close issues with stage/waiting-for-release tag when a new release is completed

mndeveci:
Adds a new GHA which will run when a release completes to close all issues with `stage/waiting-for-release` label with following message;

> Patch is released in [{release_version}](release_url). If you are AWS SAM CLI user, please wait for next AWS SAM CLI release. Closing

### DIFF
--- a/.github/workflows/close-issue-on-release.yml
+++ b/.github/workflows/close-issue-on-release.yml
@@ -1,0 +1,21 @@
+name: Close issues on release cut
+
+on:
+  release:
+    types: [released]
+
+jobs:
+  run-workflow:
+    permissions:
+      issues: write
+    runs-on: ubuntu-latest
+    steps:
+    - name: Close issues marked
+      env:
+        REPO : ${{ github.repository }}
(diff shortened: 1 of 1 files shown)


.github/workflows/close-issue-on-release.yml
```diff
@@ -0,0 +1,21 @@
+name: Close issues on release cut
+
+on:
+  release:
+    types: [released]
```
hnnasit, line 5: Would this still work even if we don't publish releases on GitHub? I am looking at https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=released#release and it seems the event gets triggered on a release being published.

hnnasit: Discussed offline. GitHub releases are done along with PyPi releases https://github.com/aws/aws-lambda-builders/releases so this should work as expected.

hnnasit (approved):

lucashuy (approved):

PR #648, opened 2024-05-11
update _RUNTIME_GLIBC mapping to support newer python versions

hawflau:
*Issue #, if available:*
#624 

*Description of changes:*

Update python glibc mapping to support newer versions of python for the scenario that a package needs to be built from source (e.g. llama-cpp)

### DIFF
--- a/aws_lambda_builders/workflows/python_pip/packager.py
+++ b/aws_lambda_builders/workflows/python_pip/packager.py
@@ -203,6 +203,9 @@ class DependencyBuilder(object):
         "cp37m": (2, 17),
         "cp38": (2, 26),
         "cp39": (2, 26),
+        "cp310": (2, 26),
+        "cp311": (2, 26),
+        "cp312": (2, 26),
     }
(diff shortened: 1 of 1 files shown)


mildaniel (approved):

aws_lambda_builders/workflows/python_pip/packager.py
```diff
@@ -203,6 +203,9 @@ class DependencyBuilder(object):
         "cp37m": (2, 17),
         "cp38": (2, 26),
         "cp39": (2, 26),
+        "cp310": (2, 26),
```
hnnasit, line 206: Have we verified installing `llama-cpp-python==0.2.55` works on python 310/311/12 after this change? I recall seeing the same error even after making this change.

hawflau: There was an error if this particular library is not built in container, as the library provides only the source tar so it must be built from source. 

It built fine in container-build without any issue

hnnasit (approved):

PR #681, opened 2024-09-05
Feature: Add npm test command during npm build

saraiyakush:
*Issue #, if available:* https://github.com/aws/aws-sam-cli/issues/7348

*Description of changes:*
- Add a new NodejsNpmTestAction that runs the `npm run test` command, if a `test` script is available in `package.json`.
- The change is added to make sure tests are executed while running `sam build` command for Node projects.
- The change will not break any build if a `test` script isn't available.

> **CAUTION:** I want to make this change non breaking hence added `--if-present` in the command. However, if any existing Node project has the test script, this feature will start running the tests during its build. Now I do not know if the any version of sam cli will use this latest version of lambda builders once this PR is merged or clients will have to upgrade to a new sam cli. If clients will have to upgrade then we're good as any existing `sam build` won't break even if they have the test script in their package.json. But if any version of sam cli will use the latest version of lambda builder then clients who do not expect the test script to run will see it run. I need your confirmation for this behaviour. Depending on that, we may have to add some mechanism for the clients to specify if they want this action to take place for their project or not.


saraiyakush:
Hello, I can see some of the windows java builds have failed. Although I do not know why as the changes are limited to Node. Can it be possible these failures existed before the changes in this PR?

hnnasit:
Hi @saraiyakush, thanks for the contributions. We will investigate why the tests are failing, they don't look related to your changes.

hnnasit (approved):

lucashuy:
> But if any version of sam cli will use the latest version of lambda builder then clients who do not expect the test script to run will see it run.

Newer versions of SAM CLI will utilize the latest version of Lamda Builders that is uploaded on [PyPi](https://pypi.org/project/aws-lambda-builders/) at the time of creating that new SAM CLI version.

Since this is something we do automatically, clients will consume this change when the following happens:
1. release a new version of Lambda Builders
2. consume that new version in SAM CLI
3. release a new version of SAM CLI

Will this change any implementation details, or will raise up any discussions you had in mind?

saraiyakush:
> > But if any version of sam cli will use the latest version of lambda builder then clients who do not expect the test script to run will see it run.
> 
> Newer versions of SAM CLI will utilize the latest version of Lambda Builders that is uploaded on [PyPi](https://pypi.org/project/aws-lambda-builders/) at the time of creating that new SAM CLI version.
> 
> Since this is something we do automatically, clients will consume this change when the following happens:
> 
> 1. release a new version of Lambda Builders
> 2. consume that new version in SAM CLI
> 3. release a new version of SAM CLI
> 
> Will this change any implementation details, or will raise up any discussions you had in mind?

From what I understood from your response, I don't think we'd have a problem. But let me validate my understanding.

I am a client using SAM CLI let's say version v2.6.1, which points to Lambda builders v4.5. With this PR merged, we get a new lambda builders v4.6.
The SAM CLI v2.6.1 will continue using lambda builders v4.5 and will not see this change.
New SAM CLI v2.6.2 will use lambda builders v4.6 and see this change.

So until a client does not upgrade their local sam cli version (or in their pipelines), we don't have a problem. Most of the pipelines I have seen don't upgrade cli versions.

However, thinking out loud, if any client has some script that installs sam cli each time (e.g. Dockerfile) in their pipelines (highly unlikely), then I assume they will install the latest SAM CLI which will point to the latest lambda builder. So adding a feature flag in this PR may not be a bad idea. So clients can choose if they want to switch on this feature.

What do you think?

hawflau:
> So adding a feature flag in this PR may not be a bad idea. So clients can choose if they want to switch on this feature.

@saraiyakush I'd agree with this suggestion

saraiyakush:
Cool. I will work on it and add a commit to this PR.

mndeveci:
Hi @saraiyakush just checking if you had a chance to work on the changes discussed above. Please let us know when we can continue the review.

Thanks again for your contributions!

saraiyakush:
Summarising the need for a feature flag and where I am stuck to implement it.

If we don't add a feature flag, clients who have a script named `test` in their `package.json` will see the script being executed as part of `sam build`.

To allow clients to choose if they want to execute the script `test` as part of `sam build` command, clients should have the ability to specify that. I am thinking the following possibilities:
- Passing an argument to the `sam build` command (e.g. runTests=true), but this will involve a change to the sam cli project and since the change is specific to Node only, I don't think this approach is a good idea.
- Add a parameter in the `template.yaml` file. Again a change needed in the sam cli project.
- Allow clients to specify another script in `package.json` that will instruct the builder whether to run the `test` script or not. Exploring this approach at the moment.

mndeveci:
>     * Allow clients to specify another script in `package.json` that will instruct the builder whether to run the `test` script or not. Exploring this approach at the moment.

Hey @saraiyakush did you have a chance to explore the option above that you mentioned?

saraiyakush:
> > Allow clients to specify another script in `package.json` that will instruct the builder whether to run the `test` script or not. Exploring this approach at the moment.
> 
> Hey @saraiyakush did you have a chance to explore the option above that you mentioned?

Hi @mndeveci , yes. I was exploring many options.
- Existence of a predefined script `package.json` that would instruct the builder to run the tests
  - This did not work out due to a limitation of NPM that it does not have a way to check if a script exists
- Existence of a predefined environment variable that would instruct the builder to run the tests
  - This seems to be working. But it isn't the most elegant solution.
- Existence of a predefined field in `template.yaml`
  - This will be the cleanest but it increases the scope of the change beyong lambda builders.

I will update the PR with Option 2 in a few days. But I would love some advise.

saraiyakush:
Hello, a reminder to review the PR.

seshubaws:
Hi @saraiyakush, thank you for the updates! I see there are 13 failing checks, all with these 3 failing tests:
```
FAILED tests/integration/workflows/nodejs_npm/test_nodejs_npm.py::TestNodejsNpmWorkflow::test_runs_test_script_if_specified_0_nodejs16_x - AssertionError: Items in the first set but not the second:
'created.js'
FAILED tests/integration/workflows/nodejs_npm/test_nodejs_npm.py::TestNodejsNpmWorkflow::test_runs_test_script_if_specified_1_nodejs18_x - AssertionError: Items in the first set but not the second:
'created.js'
FAILED tests/integration/workflows/nodejs_npm/test_nodejs_npm.py::TestNodejsNpmWorkflow::test_runs_test_script_if_specified_2_nodejs20_x - AssertionError: Items in the first set but not the second:
'created.js'
```
Do you know if these failures are related to your changes?

saraiyakush:
> Hi @saraiyakush, thank you for the updates! I see there are 13 failing checks, all with these 3 failing tests:
> 
> ```
> FAILED tests/integration/workflows/nodejs_npm/test_nodejs_npm.py::TestNodejsNpmWorkflow::test_runs_test_script_if_specified_0_nodejs16_x - AssertionError: Items in the first set but not the second:
> 'created.js'
> FAILED tests/integration/workflows/nodejs_npm/test_nodejs_npm.py::TestNodejsNpmWorkflow::test_runs_test_script_if_specified_1_nodejs18_x - AssertionError: Items in the first set but not the second:
> 'created.js'
> FAILED tests/integration/workflows/nodejs_npm/test_nodejs_npm.py::TestNodejsNpmWorkflow::test_runs_test_script_if_specified_2_nodejs20_x - AssertionError: Items in the first set but not the second:
> 'created.js'
> ```
> 
> Do you know if these failures are related to your changes?

Hey @seshubaws , Yes they seem related. Duh! Sorry, I missed running the `make pr` before pushing the commit. I will fix these and add another commit.

saraiyakush:
Hi @seshubaws , I have added the fix for the failing tests. Do we know how long does it take for the PR Workflow to report the results?

vicheey:
Unfortunately, 13 test failed for various runtimes. It seems to be related to path resolver: 
```
             except ValueError as ex:
>               raise WorkflowFailedError(workflow_name=self.NAME, action_name="Resolver", reason=str(ex))
E               aws_lambda_builders.exceptions.WorkflowFailedError: NodejsNpmBuilder:Resolver - Path resolution for runtime: nodejs16.x of binary: npm was not successful
```

tests/integration/workflows/nodejs_npm/test_nodejs_npm.py
```diff
@@ -779,3 +779,52 @@ def test_builds_project_with_manifest_outside_root_and_local_dependencies_with_d
         # expected dependencies in source directory
         source_modules = set(os.listdir(os.path.join(source_dir, "node_modules")))
         self.assertTrue(all(expected_module in source_modules for expected_module in expected_modules))
+
+    @parameterized.expand([("nodejs16.x",), ("nodejs18.x",), ("nodejs20.x",)])
+    @mock.patch.dict("os.environ", {"SAM_NPM_RUN_TEST_WITH_BUILD": "true"}, clear=True)
+    def test_runs_test_script_if_specified(self, runtime):
```
Vandita2020, line 785: The checks are failing for this particular test case. I believe that setting the `clear=True` parameter might be clearing other important environment variables needed for npm resolution. If you can remove that parameter and try running the tests, it might work.

saraiyakush: Thanks. Did that and pushed. `make pr` did not report any issues before and after the change. Hoping this was it.

vicheey (approved):

PR #690, opened 2024-10-09
fix: Change the TEMP folder for Gradle tests to home for Windows

lucashuy:
*Issue #, if available:*
None.

*Description of changes:*
Something changed within the runner environment, causing all Java Gradle tests to fail with the following error message:
```
 * What went wrong:

Gradle could not start your build.
> Could not create service of type BuildLifecycleController using ServicesProvider.createBuildLifecycleController().
   > Could not create service of type BuildModelController using VintageBuildControllerProvider.createBuildModelController().
      > Could not create service of type OutputFilesRepository using ExecutionBuildServices.createOutputFilesRepository().
         > java.io.IOException: Cannot delete file: C:\Users\RUNNER~1\AppData\Local\Temp\tmpale85rhe\gradle-cache\buildOutputCleanup\buildOutputCleanup.lock
```

I still don't know why this is the case, but we can update the folder we point Gradle to, to a the home directory of the runner as a workaround.

### DIFF
--- a/tests/integration/workflows/java_gradle/test_java_gradle.py
+++ b/tests/integration/workflows/java_gradle/test_java_gradle.py
@@ -1,6 +1,7 @@
 import os
 import shutil
 import tempfile
+import platform
 
 from unittest import TestCase
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


tests/integration/workflows/java_gradle/test_java_gradle.py
```diff
@@ -36,7 +37,18 @@ class TestJavaGradle(TestCase):
...
-        self.scratch_dir = tempfile.mkdtemp()
+
+        scratch_folder_override = None
+        if platform.system().lower() == "windows" and os.getenv("GITHUB_ACTIONS"):
+            # lucashuy: there is some really odd behaviour where
+            # gradle will refuse to work it is run within
+            # the default TEMP folder location in Github Actions
+            #
+            # use the runner's home directory as a workaround
+            scratch_folder_override = os.getenv("userprofile")
```
mndeveci, line 48: Do we need to create another folder under `userprofile` so that test files won't clash with each other?

mndeveci (approved):

hawflau (approved):

tests/integration/workflows/java_gradle/test_java_gradle.py
lucashuy: Shouldn't need to, the `tempfile` library creates unique folder names each time that shouldn't clash with anything in the existing directory. We'll probably want to create a subfolder if this was a normal test, for a normal user just to keep things nice and tidy. But since this is the CI runner, I'm sure the runner won't mind

PR #756, opened 2025-06-20
feat: Python UV workflow

sriram-mv:
Core Features:
 * Intelligent handling of pyproject.toml with automatic uv.lock detection
 * Multi-format support: pyproject.toml, requirements.txt, and requirements-*.txt variants
 * Lock file optimization: Automatic uv.lock usage for reproducible builds when available
 * Platform targeting: Lambda-compatible builds for x86_64 and ARM64 architectures
 * Python version control: Precise Python version targeting via UV's --python flags

UV Functionality:
* `uv sync --python X.Y` for lock-based builds (pyproject.toml + uv.lock)
* `uv lock && uv export` workflow for pyproject.toml without lock files
* `uv pip install --python-version X.Y --python-platform` for requirements.txt
* Proper virtual environment handling and site-packages extraction

Documentation:
* Complete DESIGN.md with architecture overview and usage patterns
* Accurate API documentation reflecting constructor + method call pattern
* Smart dispatch logic clearly explained with examples
* Updated to match actual implementation (no outdated references)

Compatibility:
* Compatible with existing Lambda Builders interface
* Supports standard build parameters (source_dir, artifacts_dir, scratch_dir)
* Runtime parameter properly extracted and used for Python version targeting
* Architecture parameter mapped to UV platform specifications

*Issue #, if available:*

*Description of changes:*


aws_lambda_builders/workflows/python_uv/utils.py
```diff
@@ -0,0 +1,148 @@
...
+
+    Returns:
+        Path to the detected manifest file, or None if not found
+    """
+    # Check for pyproject.toml first (preferred manifest)
+    pyproject_path = os.path.join(source_dir, "pyproject.toml")
+    if os.path.isfile(pyproject_path):
+        return pyproject_path
+    
+    # Check for requirements.txt variants (in order of preference)
```
sriram-mv, line 54: Do we need these variants? could we go simpler?

aws_lambda_builders/workflows/python_uv/validator.py
```diff
@@ -0,0 +1,74 @@
...
+
+from aws_lambda_builders.exceptions import MisMatchRuntimeError
+from aws_lambda_builders.validator import RuntimeValidator
+
+from .utils import OSUtils
+
+LOG = logging.getLogger(__name__)
+
+
+class PythonUvRuntimeValidator(RuntimeValidator):
```
sriram-mv, line 17: I need to double check if we need the validator, uv allows you to specify the python version or even virtualenv afaik.

aws_lambda_builders/workflows/python_uv/utils.py
sriram-mv: Python_pip seems to support it. so let's keep it.

reedham-aws:
If there's not any movement on this PR in a couple weeks and I have some free time, I will try to update this PR with the feedback after taking a look.

sriram-mv:
@reedham-aws I'll pick it up maybe later next week. Is there any other feedback that I can address in one go apart from what's mentioned?

reedham-aws (commented):
I didn't look too hard at the tests, and overall the code looks good. I have a couple code quality suggestions and a question about the way the code chooses to do `uv pip install` versus `uv sync`.

aws_lambda_builders/workflows/python_uv/packager.py
```diff
@@ -0,0 +1,504 @@
...
+        self._osutils = osutils
+        self._uv_runner = uv_runner
+        self.runtime = runtime
+
+    def build_dependencies(
+        self,
+        artifacts_dir_path: str,
+        scratch_dir_path: str,
+        manifest_path: str,
+        architecture: str = X86_64,
```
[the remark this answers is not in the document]
reedham-aws: I think this is just the default value

roger-zhangg (dismissed):
LGTM! As a reminder we will need to add uv in https://github.com/aws/aws-sam-build-images and add this workflow in https://github.com/aws/aws-sam-cli to fully support UV build workflow

aws_lambda_builders/workflows/python_uv/exceptions.py
```diff
@@ -0,0 +1,32 @@
...
+"""
+
+from aws_lambda_builders.exceptions import LambdaBuilderError
+
+
+class MissingUvError(LambdaBuilderError):
+    """Exception raised when UV executable is not found."""
+
+    MESSAGE = (
+        "uv executable not found at {uv_path}. "
```
valerena: It probably doesn't make sense to have a `uv_path` if the error is precisely that we can't find uv. (if there was a "uv path", then that would mean that we have `uv`)

.github/workflows/build.yml
```diff
@@ -248,8 +250,37 @@ jobs:
...
+        os:
+          - ubuntu-latest
+          - windows-latest
+        python:
+          - "3.8"
+          - "3.9"
+          - "3.10"
+          - "3.11"
+          - "3.12"
+          - "3.13"
```
valerena: This is missing 3.14, that was launched in October.

valerena (dismissed):

reedham-aws:
I have made one change to the PR, which is to add a cwd to when trying to install dependencies from an existing `uv.lock`. This is because I was seeing inconsistencies with UV being able to find the lockfile when it was in the source code directory (the most logical place for it to be).

sriram-mv:
in SAM CLI can you actually set which workflow you would want to pick? is that something we can add in the metadata for that lambda function. I vaguely recall that something like that being present for building layers.

reedham-aws:
I think that would be possible, right now we're using the experimental feature tag in SAM CLI so that would be in the future. What I referenced above would still be a problem if a user chose UV and had that setup.

ncgl-git:
@reedham-aws ?

Issue #700, opened 2024-11-04
No support for Python packages that rely on manylinux wheels newer than `manylinux2014`


svet-b:
### Description:
Using `sam build` (without `--use-container`) and trying to install a recent version of Polars, such as 1.12.0, fails with "Could not find a version that satisfies the requirement ...". Only Polars versions up to 0.19.12 are listed as available.

This is because Polars stopped shipping wheels based on `manylinux2014` ([link](https://github.com/pola-rs/polars/issues/13799)), on the basis that it's outdated ([and essentially EOL](https://github.com/mayeut/pep600_compliance?tab=readme-ov-file#acceptable-distros-to-build-wheels)).

Meanwhile, at present `sam build` supports _only_ `manylinux` up to `manylinux2014` (aka `manylinux_2_17`) wheels, and no versions more recent than that ([source](https://github.com/aws/aws-lambda-builders/blob/main/aws_lambda_builders/workflows/python_pip/packager.py#L173)). I'm experiencing this on ARM64, but it appears to be true for x86_64 also.

I haven't done a thorough search on how widespread of an issue this is, but there's at least one other similar report here: https://github.com/aws/aws-lambda-builders/issues/674

Given that Python package maintainers are moving away from an outdated version of `manylinux`, it would be great for this toolchain to also support more recent versions.

I should note that using `sam build --use-container` does build successfully. But I'd like to avoid that since (a) it adds some weight to the build process, and (b) I'm running the build inside CodePipeline, which is not happy with me trying to spin up a container inside it.

Hopefully it's ok to not add further detail for now as I think the issue I'm describing is self-evident - but happy to attach some detailed logs/reproducers if actually helpful.

mndeveci:
Thanks for reporting this issue.

If you build it with `sam build --use-container`, does running `sam local invoke` and deploying the function works as expected? If so, we need to attach this issue to the one above to make sure AWS Lambda Builders can support the newer `manylinux` versions that is supported by Lambda.

svet-b:
Thank you for looking into this. Indeed, after building with `--use-container`, running `sam local invoke` works as expected.

vicheey:
Thank you for confirming. We are keeping this issue open for now to see if we have any reaction from community who face similar issue and will prioritize accordingly.

stschulte:
[...]

PR #762, opened 2025-07-14
feat: allow manylinux 2.28 and 2.34 on python 3.12+ when compiled on a different architecture

alesanfra:
*Issue:* #700

*Description of changes:*
As described by @valerena [here](https://github.com/aws/aws-lambda-builders/issues/700#issuecomment-3049644352), when trying to build cross-architecture, the platform tags added to the pip command are not correct. With this PR, I've fixed the issue while preserving most of the original logic. The approach is to add all compatible platform tags for each python version and architecture pair to the pip command, as per the [official pip documentation](https://pip.pypa.io/en/stable/cli/pip_download/#cmdoption-platform).

The code may look a bit hacky, but I tried to change as few lines as possible. Feel free to propose improvements - I'm ready to accept them.

### DIFF
--- a/aws_lambda_builders/workflows/python_pip/packager.py
+++ b/aws_lambda_builders/workflows/python_pip/packager.py
@@ -2,11 +2,12 @@
 Installs packages using PIP
 """
 
+import itertools
 import logging
 import re
 import subprocess
 from email.parser import FeedParser
-from typing import Tuple
+from typing import Tuple, List
 
 from aws_lambda_builders.architecture import ARM64, X86_64
 from aws_lambda_builders.utils import extract_tarfile
@@ -171,19 +172,25 @@ class DependencyBuilder(object):
     packager.
     """
 
-    _COMPATIBLE_PLATFORM_ARM64 = {
+    _COMPATIBLE_PLATFORM_ARM64 = [
         "any",
         "linux_aarch64",
         "manylinux2014_aarch64",
-    }
+        "manylinux_2_17_aarch64",
+        "manylinux_2_28_aarch64",
+        "manylinux_2_34_aarch64",
+    ]
 
-    _COMPATIBLE_PLATFORM_X86_64 = {
+    _COMPATIBLE_PLATFORM_X86_64 = [
         "any",
         "linux_x86_64",
         "manylinux1_x86_64",
         "manylinux2010_x86_64",
         "manylinux2014_x86_64",
-    }
+        "manylinux_2_17_x86_64",
+        "manylinux_2_28_x86_64",
+        "manylinux_2_34_x86_64",
+    ]
 
     _COMPATIBLE_PLATFORMS = {
         ARM64: _COMPATIBLE_PLATFORM_ARM64,
@@ -213,6 +220,14 @@ class DependencyBuilder(object):
     # not in _RUNTIME_GLIBC.
     # Unlikely to hit this case.
     _DEFAULT_GLIBC = (2, 17)
+
+    # Mapping of glibc version to the most recent manylinux version compatible.
+    # The offically supported manylinux versions are 2_17, 2_28 and 2_34 as per https://github.com/pypa/manylinux
+    _GLIBC_TO_LATEST_MANYLINUX = {
+        (2, 17): "manylinux_2_17",
+        (2, 26): "manylinux_2_17",
+        (2, 34): "manylinux_2_34",
+    }
 
     def __init__(self, osutils, runtime, python_exe, pip_runner=None, architecture=X86_64):
         """Initialize a DependencyBuilder.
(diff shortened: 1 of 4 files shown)


aws_lambda_builders/workflows/python_pip/packager.py
```diff
@@ -171,19 +172,25 @@ class DependencyBuilder(object):
...
-    _COMPATIBLE_PLATFORM_ARM64 = {
+    _COMPATIBLE_PLATFORM_ARM64 = [
         "any",
         "linux_aarch64",
         "manylinux2014_aarch64",
-    }
+        "manylinux_2_17_aarch64",
+        "manylinux_2_28_aarch64",
+        "manylinux_2_34_aarch64",
+    ]
```
JonZeolla, line 182: I'm not a codeowner here, just an interested party, but is there a reason to make this a list instead of a set? afaict order doesn't matter and set operations are very useful here

alesanfra: TL; DR: I need the platforms to be sorted by release time.

Good question! The problem with the `--platform` option in `pip download` is that it tries to match exactly the platform you pass as an argument, excluding potentially compatible wheels compiled with an earlier version of glibc.

According to pip's documentation, it is up to the caller to pass the full list of compatible platforms to the target, because pip makes no a priori assumptions.

Let's take an example: suppose we want to install numpy on Amazon Linux 2023, which runs glibc version 2. 34, but have a different operating system or architecture (e.g., Mac), so let's try downloading numpy by adding the platform tag:
```shell
$ pip download numpy --platform manylinux_2_34_x86_64 --only-binary=:all:
ERROR: Could not find a version that satisfies the requirement numpy (from versions: none)
ERROR: No matching distribution found for numpy
```

This is strange but expected, because there is no version of numpy on pypi.org with that platform tag. You need to **pass all possible compatible tags** to the command, so a working command would be:
```shell
$ pip download numpy --platform manylinux_2_17_x86_64 --platform manylinux_2_28_x86_64 --platform manylinux_2_34_x86_64 --only-binary=:all:
```

With this command you will get numpy version 2.3.0, which comes with the manylinux_2_28_x86_64 platform tag.

This means that for each pair of glibc versions and architectures there is a different set of compatible platform tags. Unfortunately, I could not find a clever way to implement this in python, but I used this simple logic:
1. For each architecture, list all possible platform tags **ordered** from oldest to newest;
2. At runtime, determine the most recent platform tag compatible with our target (in the example it is `manylinux_2_34_x86_64`);
3. Pass the platform determined in step 2 and all previously released platforms to the command.

This is why I turned the set into a list: I need the platforms to be sorted by release time.

If you have any ideas to improve the code feel free to share them here, this is the simplest approach I could find.

bnusunny:
@alesanfra Thanks for this PR. It looks great! I would suggest a few additional improvements to make it more maintainable. 

1. Add ManylinuxSpec dataclass

```python
from dataclasses import dataclass

@dataclass(frozen=True)
class ManylinuxSpec:
    platform_template: str
    min_glibc_major: Optional[int] = None
    min_glibc_minor: Optional[int] = None
    introduced_year: Optional[int] = None
```

2. Replace static platform lists with ordered evolution 

```python
class DependencyBuilder(object):
    _MANYLINUX_EVOLUTION = [
        ManylinuxSpec("any"),
        ManylinuxSpec("linux_{arch}"),
        ManylinuxSpec("manylinux1_{arch}", 2, 5, 2016),
        ManylinuxSpec("manylinux2010_{arch}", 2, 12, 2019),
        ManylinuxSpec("manylinux2014_{arch}", 2, 17, 2019),
        ManylinuxSpec("manylinux_2_17_{arch}", 2, 17, 2019),
        ManylinuxSpec("manylinux_2_28_{arch}", 2, 28, 2022),
        ManylinuxSpec("manylinux_2_34_{arch}", 2, 34, 2023),
    ]
```
 
3. Implement dynamic platform selection 

```python
class DependencyBuilder(object):
    @property
    def compatible_platforms(self) -> List[str]:
        lambda_abi = get_lambda_abi(self.runtime)
        runtime_glibc = self._RUNTIME_GLIBC.get(lambda_abi, self._DEFAULT_GLIBC)
        arch = "aarch64" if self.architecture == ARM64 else "x86_64"
        
        platforms = []
        for spec in self._MANYLINUX_EVOLUTION:
            if spec.is_compatible_with_glibc(runtime_glibc):
                platforms.append(spec.format_for_arch(arch))
            else:
                break  # Stop at first incompatible version
                
        return platforms
```

aws_lambda_builders/workflows/python_pip/packager.py
```diff
@@ -171,19 +172,25 @@ class DependencyBuilder(object):
...
 
-    _COMPATIBLE_PLATFORM_X86_64 = {
+    _COMPATIBLE_PLATFORM_X86_64 = [
         "any",
         "linux_x86_64",
         "manylinux1_x86_64",
         "manylinux2010_x86_64",
         "manylinux2014_x86_64",
-    }
+        "manylinux_2_17_x86_64",
```
roger-zhangg, line 190: I don't know enough, but seems manylinux_2_17 is just a alias of manylinux2014_x86_64, do we need this here?

roger-zhangg (approved):
Just FYI if anyone else reviewing, manylinux_2_17 is an alias for manylinux2014. So although this change seems complex the actual behavior change is added manylinux_2_28 and manylinux_2_34 when python runtime is 3.12+. 
Thanks for the contribution!

plondino:
I am curious why this is restricted to only supporting python runtime is 3.12+ - I am trying to install confluent Kafka on python 3.11 and it is failing because it's manylinux_2_28: https://pypi.org/project/confluent-kafka/#confluent_kafka-2.11.1-cp311-cp311-manylinux_2_28_x86_64.whl

PR #768, opened 2025-08-13
Add Gradle 9.0 support

mbfreder:
Add Gradle 9.0 support by removing deprecated --build-file argument and distsDirName property

- Remove --build-file argument from gradle.py as it was deprecated in Gradle 8 and removed in Gradle 9
- Update lambda-build-init.gradle to use 'distributions' instead of deprecated distsDirName property
- Update all related unit tests to reflect the command line changes
- All unit and integration tests now pass with Gradle 9.0

Fixes: Unknown command-line option '--build-file' error when using Gradle 9.0

*Issue #, if available:*

*Description of changes:*


aws_lambda_builders/workflows/java_gradle/gradle.py
```diff
@@ -33,7 +33,7 @@ def build(self, source_dir, build_file, cache_dir=None, init_script_path=None, p
         if not self.os_utils.exists(build_file):
             raise BuildFileNotFoundError(build_file)
 
-        args = ["build", "--build-file", build_file]
```
valerena, line 36: Random thoughts here.

I see that now the `build_file` parameter is not used anywhere in the method anymore, so maybe we could remove it from the function signature: `build(self, source_dir, cache_dir=None, init_script_path=None, properties=None)`

If we do that, we would probably want to update the rest of the places where this is used, like the [actions file](https://github.com/aws/aws-lambda-builders/blob/develop/aws_lambda_builders/workflows/java_gradle/actions.py#L50-L52) and remove it from the [JavaGradleBuildAction fields](https://github.com/aws/aws-lambda-builders/blob/develop/aws_lambda_builders/workflows/java_gradle/actions.py#L24), and then remove it from the [workflow file](https://github.com/aws/aws-lambda-builders/blob/1c1f98e1dbff45a517647b5a3528ca0b5a881d32/aws_lambda_builders/workflows/java_gradle/workflow.py#L49) where this is called.

This would potentially break users if they're using these classes directly instead of using the `LambdaBuilder` object that [SAM CLI uses](https://github.com/aws/aws-sam-cli/blob/develop/samcli/lib/build/app_builder.py#L891C19-L916), but I guess if that's the case, they can "not upgrade to the new lambda-builders version until they remove this extra parameter from their code". 

Nah, but I think at the end of the day, it's okay to leave it like this. Maybe it could be better to add a comment here just mentioning that we're not using the `build_file` parameter because Gradle doesn't support it anymore, but outside of that, I think we're okay.

valerena (approved):

seshubaws (approved):

tests/unit/workflows/java_gradle/test_gradle.py
```diff
@@ -60,7 +60,7 @@ def test_build_no_init_script(self):
         gradle = SubprocessGradle(gradle_binary=self.gradle_binary, os_utils=self.os_utils)
         gradle.build(self.source_dir, self.manifest_path)
         self.os_utils.popen.assert_called_with(
-            [self.gradle_path, "build", "--build-file", self.manifest_path],
+            [self.gradle_path, "build"],
```
roger-zhangg, line 63: do we have test that actually runs gradle, rather than the mock?

roger-zhangg: https://github.com/aws/aws-lambda-builders/blob/3e531b3b118ee17003aef104d94ed225c55d2289/tests/integration/workflows/java_gradle/testdata/single-build/java21/with-deps-gradlew/gradle/wrapper/gradle-wrapper.properties#L3

Seems the highest gradle version we are testing is gradle 8.4. Do we need to test 9.0?

Vandita2020: Addressed this in a new PR: https://github.com/aws/aws-lambda-builders/pull/769

PR #793, opened 2025-10-28
feat: add Python 3.14 and Java 25 runtimes

bnusunny:
*Issue #, if available:*

### Description of changes

Add Python 3.14 and Java 25 Runtimes

### Checklist

### DIFF
--- a/aws_lambda_builders/workflows/python_pip/packager.py
+++ b/aws_lambda_builders/workflows/python_pip/packager.py
@@ -5,6 +5,7 @@ Installs packages using PIP
 import itertools
 import logging
 import re
+import subprocess
 from email.parser import FeedParser
 from typing import List, Tuple
 
@@ -670,8 +671,6 @@ class SDistMetadataFetcher(object):
         # In Python 3.12+, setuptools might not be available by default
         try:
             # Check if setuptools is available in the current environment
-            import subprocess
-
             check_cmd = [self.python_exe, "-c", "import setuptools"]
             result = subprocess.run(check_cmd, capture_output=True, timeout=10, check=False)
             if result.returncode != 0:


aws_lambda_builders/workflows/python_pip/packager.py
```diff
@@ -5,7 +5,6 @@
 import itertools
 import logging
 import re
-import subprocess
```
vicheey: I see that we are removing `import subprocess` and replace with import statement on line 673. However, wouldn't this break line 672 (689 new)? 
```
        p = subprocess.Popen(
            cmd, cwd=package_dir, stdout=subprocess.PIPE, stderr=subprocess.PIPE, env=self._osutils.original_environ()
        )
```

tests/unit/test_validator.py
```diff
@@ -17,13 +18,293 @@ def test_validate_runtime(self):
         self.assertEqual(self.validator._runtime_path, "/usr/bin/python3.8")
 
     def test_validate_with_unsupported_runtime(self):
-        validator = RuntimeValidator(runtime="unknown_runtime", architecture="x86_64")
-        with self.assertRaises(UnsupportedRuntimeError):
-            validator.validate("/usr/bin/unknown_runtime")
+        """Test validation fails for completely unsupported runtimes"""
+        unsupported_runtimes = [
```
seshubaws: i don't know if we need to test with all these unsupported runtimes, for the sake of test latency can we keep the list to 5? but very thorough though!

tests/unit/test_validator.py
```diff
@@ -17,13 +18,293 @@ def test_validate_runtime(self):
...
 
     def test_validate_with_runtime_and_incompatible_architecture(self):
-        runtime_list = ["python3.12"]
-        for runtime in runtime_list:
-            validator = RuntimeValidator(runtime=runtime, architecture="invalid_arch")
-            with self.assertRaises(UnsupportedArchitectureError):
-                validator.validate("/usr/bin/{}".format(runtime))
+        """Test validation fails for unsupported architecture combinations"""
+        # Test with various invalid architectures
+        invalid_architectures = [
```
seshubaws: same comment as above, and i see the same large lists below too that we can cut down on

seshubaws (dismissed):

tests/unit/test_validator.py
bnusunny: Yes, it's not necessary. I removed them.

aws_lambda_builders/workflows/python_pip/packager.py
bnusunny: Updated, thanks.

seshubaws (approved):

PR #799, opened 2025-11-11
feat: add nodejs24.x support

bnusunny:
feat: add nodejs24.x support

### Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -80,7 +80,7 @@ jobs:
         run: make func-test
 
   node-integration:
-    name: ${{ matrix.os }} / ${{ matrix.python }} / node / npm ${{ matrix.npm }}.x
+    name: ${{ matrix.os }} / ${{ matrix.python }} / node ${{ matrix.nodejs }} / npm ${{ matrix.npm }}.x
     if: github.repository_owner == 'aws'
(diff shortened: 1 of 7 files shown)


aws_lambda_builders/__init__.py
```diff
@@ -5,5 +5,5 @@
 # Changing version will trigger a new release!
 # Please make the version change as the last step of your development.
 
-__version__ = "1.59.0"
+__version__ = "1.58.0"
```
valerena: I imagine we don't want to undo the version update

.github/workflows/build.yml
```diff
@@ -90,33 +90,32 @@ jobs:
           - ubuntu-latest
           - windows-latest
         python:
-          - "3.9"
-          - "3.10"
-          - "3.11"
-          - "3.12"
           - "3.13"
-          - "3.14"
```
valerena: These versions of Python helps us confirm that `aws-lambda-builders` **as a library** works in different versions of python (Meaning that if you have python3.9, you can do `pip install aws-lambda-builders` and you'll be able to use it correctly. 

Because this usage is limited (mostly used through SAM CLI anyway), then this is not the most crucial thing to check. But if it's not super slow or causing particular issues, we could probably add it back in the future.

valerena (approved):

PR #847, opened 2026-03-17
chore: drop Python 3.9 support for running aws-lambda-builders

roger-zhangg:
## Summary

Drop Python 3.9 as a supported version for **running** aws-lambda-builders. The minimum Python version is now 3.10. CI test matrices now cover Python 3.10 through 3.14.

## What changed

- `setup.py`: `python_requires` bumped to `>=3.10`, removed 3.8/3.9 classifiers
- `.github/workflows/build.yml`: removed `3.9` from unit-functional, python-pip-integration, and python-uv-integration matrices
- `pyproject.toml`: updated black `target_version` to `py310`–`py314`
- `requirements/dev.txt`: simplified version-conditional pins that only existed for <3.10 compat

## What did NOT change

Building Lambda functions targeting the `python3.9` runtime is **still fully supported**. The `python3.9` entry in `supported_runtimes.py` and all related Lambda workflow code remains untouched. This change only affects the minimum Python version required to run aws-lambda-builders itself.

### DIFF
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -1,8 +1,6 @@
-coverage==7.13.5; python_version>="3.10"
-coverage==7.10.7; python_version<"3.10"
+coverage==7.13.5
 flake8==3.8.4
-pytest-cov==6.3.0; python_version>="3.9"
-pytest-cov==5.0.0; python_version<"3.9"
+pytest-cov==6.3.0
 
 # Test requirements
 pytest>=6.1.1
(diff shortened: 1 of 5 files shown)


requirements/dev.txt
```diff
@@ -1,15 +1,12 @@
-coverage==7.13.5; python_version>="3.10"
```
seshubaws, line 1: why get rid of python_version>="3.10"? and in line 13?

roger-zhangg: We are dropping python3.9 support here. So the minimum version is python3.10. Making this check unnecessary

valerena (approved):

seshubaws (approved):

PR #855, opened 2026-04-08
feat: add support for ruby 4.0

reedham-aws:
### Description of changes
Adds basic support and testing for Ruby 4.0. This is mostly just adding it to our supported runtime list and testing that it is successful.

`make pr` passed, seeing if the CI passes.

### Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -312,7 +312,7 @@ jobs:
       - run: pytest -vv tests/integration/workflows/python_uv
 
   ruby-integration:
-    name: ${{ matrix.os }} / ${{ matrix.python }} / ruby
+    name: ${{ matrix.os }} / ${{ matrix.python }} / ruby ${{ matrix.ruby }}
     if: github.repository_owner == 'aws'
     runs-on: ${{ matrix.os }}
     strategy:
@@ -323,6 +323,9 @@ jobs:
           - windows-latest
         python:
           - "3.13"
+        ruby:
+          - "3.2"
+          - "4.0"
     steps:
       - uses: actions/checkout@v6
       - uses: actions/setup-python@v6
@@ -330,7 +333,7 @@ jobs:
           python-version: ${{ matrix.python }}
       - uses: ruby/setup-ruby@4dc28cf14d77b0afa6832d9765ac422dbf0dfedd # v1
         with:
-          ruby-version: "3.2"
+          ruby-version: ${{ matrix.ruby }}
       - run: make init
       - run: pytest -vv tests/integration/workflows/ruby_bundler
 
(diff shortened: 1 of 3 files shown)


reedham-aws:
The issues with the CI are unrelated to any builders changes, and instead seem to be an issue with `npm`: https://github.com/npm/cli/issues/9151. Seems like there might be some workarounds mentioned in the thread but for this PR I think it's fine to ignore.

.github/workflows/build.yml
```diff
@@ -323,14 +323,17 @@ jobs:
           - windows-latest
         python:
           - "3.13"
+        ruby:
+          - "3.2"
```
seshubaws, line 337: create for ruby3.2 is actually supposed to be deprecated April 30, can we remove this?

reedham-aws: Based on what I see [here](https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html) function create is blocked 2026-08-31 and function update is blocked on 2026-09-30. My understanding is we usually don't remove runtimes until their update is blocked.

.github/workflows/build.yml
```diff
@@ -323,14 +323,17 @@ jobs:
...
       - uses: actions/checkout@v6
       - uses: actions/setup-python@v6
         with:
           python-version: ${{ matrix.python }}
       - uses: ruby/setup-ruby@4dc28cf14d77b0afa6832d9765ac422dbf0dfedd # v1
         with:
-          ruby-version: "3.2"
+          ruby-version: ${{ matrix.ruby }}
       - run: make init
       - run: pytest -vv tests/integration/workflows/ruby_bundler
```
valerena, line 348: It's interesting that we use every these two versions of ruby (3.2, 4.0) to check that lambda-builders can build every version of ruby (from the `test_ruby.py` file).

One would think that we just need to build each version with the same version, but this is a more complete test (assuming that the test is real and it's actually building for each of the different ruby versions instead of just using the ruby from the environment and ignoring the runtime passed as a parameter.. in that case, we would be running the same test 4 times... so hopefully this all works how it's supposed to, and we have a very good test)

valerena (approved):

seshubaws (approved):

.github/workflows/build.yml
reedham-aws: I don't remember too well, but I think I had previously considered this issue when working on https://github.com/aws/aws-lambda-builders/pull/808. I think you might be right (unfortunately) that we're running tests for functions with runtimes that are supposedly 3.4 or 3.6, but it's just in an environment that doesn't even have that version. I think that's non-blocking for this PR, but probably requires some more testing.

PR #870, opened 2026-05-13
fix(python-uv): cached build copies dependencies to the build dir

### LINKED ISSUE #864: Bug: PythonUvBuilder:CopyDependencies fails when sam build --cached deletes .aws-sam/build/<FunctionName>
## Description

`PythonUvBuilder:CopyDependencies` fails with `[Errno 2] No such file or directory: '.aws-sam/build/<FunctionName>'` when invoked from `sam build --cached --beta-features`. The cached-build path removes the build output directory but `CopyDependencies` expects it to already exist.

**Destructive failure mode:** the failed invocation also wipes `.aws-sam/build/` for all functions, so users must `sam build --beta-features` from scratch to recover.

Filed here per @vicheey's diagnosis on aws/aws-sam-cli#8838 that the fix likely belongs in `aws-lambda-builders`.

## Steps to reproduce

```bash
# Any SAM project with 2+ Python functions using python-uv builder
sam build --beta-features          # succeeds
sam build --cached --beta-features # fails — also wipes prior .aws-sam/build/
```

## Observed result

```
Starting Build use cache
Building codeuri: /path/to/src/my_function runtime: python3.14 architecture: arm64 functions: MyFunction
Auto-detected manifest file: /path/to/src/my_function/pyproject.toml
 Running PythonUvBuilder:CleanUp
 Running PythonUvBuilder:ResolveDependencies
Building Python dependencies using UV
Found uv.lock alongside pyproject.toml - using lock-based build for precise dependencies
Building from UV lock file
 Running PythonUvBuilder:CopyDependencies

Build Failed
Error: PythonUvBuilder:CopyDependencies - [Errno 2] No such file or directory: '/path/to/.aws-sam/build/MyFunction'
```

`.aws-sam/build.toml`: every function entry has `manifest_hash = ""`, so cache-decision logic routes through `CopyDependencies` against a non-existent target dir.

## Expected result

`sam build --cached --beta-features` reuses cached deps when manifest unchanged; only re-copies source. Build dir recreated before `CopyDependencies` runs. No destructive state-loss on failure.

## Notes

- `sam build --beta-features` (no `--cached`): works
- `sam build --parallel --beta-features` (no `--cached`): works
- `sam build --cached --parallel --beta-features`: same failure

## Environment

Reproduced 2026-04-26:

```
SAM CLI: 1.158.0
aws-lambda-builders: shipped with sam-cli 1.158.0
Python: 3.14
uv: 0.11.7
OS: macOS 26.3 (aarch64-apple-darwin)
```

Related: aws/aws-sam-cli#8838

tomaDev:
### Issue

Fixes #864.

\`sam build --cached --beta-features\` against a project using the
python-uv workflow fails with:

\`\`\`
Build Failed
Error: PythonUvBuilder:CopyDependencies - [Errno 2] No such file or directory: '.aws-sam/build/<FunctionName>'
\`\`\`

The failure is also destructive — it wipes \`.aws-sam/build/\` for every
function, forcing a full \`sam build --beta-features\` to recover.

### Root cause

\`sam build --cached\` runs the workflow with a \`dependencies_dir\` set
(\`.aws-sam/deps/<uuid>\`) so deps can be reused across builds. Two bugs
on that path:

1. **\`PythonUvWorkflow._setup_build_actions\` constructs \`CopyDependenciesAction\` with the source/destination swapped.** Dependencies are installed into \`dependencies_dir\`, so that is the source of the copy and \`artifacts_dir\` the destination. As written, the action listed the not-yet-created \`artifacts_dir\` as source and never populated it — hence the \`[Errno 2]\`.

2. **\`UvRunner.install_requirements\` passes \`--target\` to UV verbatim.** UV runs with \`cwd\` set to the project directory, so a relative \`.aws-sam/deps/<uuid>\` target is resolved under the source dir, not the build root. \`uv pip install\` succeeds but writes to the wrong place — the cached path then can't find the deps.

### Fix

- Swap the \`CopyDependenciesAction\` arguments in \`workflow.py\` so it copies from \`dependencies_dir\` → \`artifacts_dir\`.
- Resolve \`--target\` to an absolute path before invoking UV in \`packager.py\`.

### Tests

- Unit test in \`test_workflow.py\` asserts \`CopyDependenciesAction\` is constructed with the correct source/artifact/destination wiring.
- Unit test in \`test_packager.py\` asserts a relative \`--target\` is made absolute.
- Existing \`dependencies_dir\` integration test now asserts deps end up in \`artifacts_dir\` (previously only checked that the build didn't crash).

Verified locally: \`sam build --cached --parallel --beta-features\` now
succeeds end-to-end on a 9-function python-uv project.

### DIFF
--- a/aws_lambda_builders/workflows/python_uv/packager.py
+++ b/aws_lambda_builders/workflows/python_uv/packager.py
@@ -136,8 +136,10 @@ class UvRunner:
         # Add requirements file
         args.extend(["-r", requirements_path])
 
-        # Add target directory
-        args.extend(["--target", target_dir])
+        # Resolve --target to an absolute path: UV runs with cwd set to the project directory, so a
+        # relative target (e.g. the incremental-build dependencies dir) would otherwise be created
+        # under the source directory instead of the build root.
+        args.extend(["--target", os.path.abspath(target_dir)])
 
         # Add configuration arguments
         args.extend(config.to_uv_args())
--- a/aws_lambda_builders/workflows/python_uv/workflow.py
+++ b/aws_lambda_builders/workflows/python_uv/workflow.py
@@ -139,13 +139,15 @@ class PythonUvWorkflow(BaseWorkflow):
                 )
             )
 
-        # Advanced case: Copy dependencies from dependencies_dir to artifacts_dir if configured
+        # Advanced case: copy dependencies from dependencies_dir to artifacts_dir if configured.
+        # Dependencies were installed into dependencies_dir above, so that is the copy source
+        # (artifact_dir) and artifacts_dir is the destination.
         if self.dependencies_dir and self.combine_dependencies:
             self.actions.append(
                 CopyDependenciesAction(
                     source_dir=source_dir,
-                    artifact_dir=artifacts_dir,
-                    destination_dir=self.dependencies_dir,
+                    artifact_dir=self.dependencies_dir,
+                    destination_dir=artifacts_dir,
                     maintain_symlinks=False,
                 )
             )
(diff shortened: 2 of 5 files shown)


reedham-aws (approved):

reedham-aws:
Thank you for this fix, apologies it took so long to merge

licjun (approved):

PR #887, opened 2026-07-09
fix(python-uv): take into account dependencies on other workspace members

### LINKED ISSUE #892: Bug: PythonUvBuilder fails to build app with dependencies on editable installs in the workspace
### Description:

An application created as a member of a uv workspace would fail to build if they depended on other workspace
members.

The following shows the problematic workspace structure.

```
workspace root
├── lib
│   ├── pyproject.toml
│   └── src
├── pyproject.toml
├── sam-app
│   ├── __init__.py
│   ├── app.py
│   ├── pyproject.toml
│   ├── samconfig.toml
│   └── template.yaml
└── uv.lock
```

### Steps to reproduce:

```bash
uv init --bare
uv init --lib lib
sam init --name sam-app --runtime python3.14 --architecture arm64 \
    --dependency-manager pip --package-type Zip \
    --app-template hello-world
cd sam-app
uv init
uv add lib@../lib
# remove requirements.txt and edit the app
# edit template.yaml to configure `BuildMethod: python-uv` and `CodeUri: .`
sam build --beta-features
```

### Observed result:

The following shows the problematic workspace structure and the error message.

```
2026-07-15 13:49:33,345 | Building from UV lock file
2026-07-15 13:49:33,346 | Executing UV command: /Users/me/.cargo/bin/uv export --format requirements.txt --no-emit-project --no-hashes --no-default-groups --output-file
/var/folders/8s/g1znx0c90vx3tz81t_rqqgyr0000gn/T/tmptz94ffbg/lock_requirements.txt --python 3.14
2026-07-15 13:49:33,499 | UV command return code: 0
2026-07-15 13:49:33,500 | UV stdout: # This file was autogenerated by uv via the following command:
#    uv export --format requirements.txt --no-emit-project --no-hashes --no-default-groups --output-file /var/folders/8s/g1znx0c90vx3tz81t_rqqgyr0000gn/T/tmptz94ffbg/lock_requirements.txt --python 3.14
-e ./lib
    # via sam-app

2026-07-15 13:49:33,500 | UV stderr: Using CPython 3.14.6 interpreter at: /opt/homebrew/opt/python@3.14/bin/python3.14
Resolved 3 packages in 8ms

2026-07-15 13:49:33,501 | Executing UV command: /Users/me/.cargo/bin/uv pip install -r /var/folders/8s/g1znx0c90vx3tz81t_rqqgyr0000gn/T/tmptz94ffbg/lock_requirements.txt --target
/Users/me/workspace/sam-app/.aws-sam/deps/351538c7-a05e-4aad-b5c5-928f241070b6 --cache-dir /var/folders/8s/g1znx0c90vx3tz81t_rqqgyr0000gn/T/tmptz94ffbg/uv-cache
--python-version 3.14 --python-platform aarch64-unknown-linux-gnu
2026-07-15 13:49:33,653 | UV command return code: 2
2026-07-15 13:49:33,655 | UV stdout:
2026-07-15 13:49:33,656 | UV stderr: Using CPython 3.13.0 interpreter at: /Users/me/workspace/.venv/bin/python3
error: Distribution not found at: file:///Users/me/workspace/sam-app/lib

2026-07-15 13:49:33,657 | Failed to build dependencies: UV package build failed: Failed to build from pyproject.toml: Lock file operation failed: Failed to install dependencies using uv: UV pip install
failed: Using CPython 3.13.0 interpreter at: /Users/me/workspace/.venv/bin/python3
error: Distribution not found at: file:///Users/me/workspace/sam-app/lib

2026-07-15 13:49:33,662 | PythonUvBuilder:ResolveDependencies failed
  File "/opt/homebrew/Cellar/aws-sam-cli/1.162.1/libexec/lib/python3.14/site-packages/aws_lambda_builders/workflows/python_uv/packager.py", line 348, in _build_from_lock_file
    self._uv_runner.install_requirements(
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        requirements_path=temp_requirements,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<6 lines>...
        architecture=architecture,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/opt/homebrew/Cellar/aws-sam-cli/1.162.1/libexec/lib/python3.14/site-packages/aws_lambda_builders/workflows/python_uv/packager.py", line 167, in install_requirements
    raise UvInstallationError(reason=f"UV pip install failed: {stderr}")
aws_lambda_builders.workflows.python_uv.exceptions.UvInstallationError: Failed to install dependencies using uv: UV pip install failed: Using CPython 3.13.0 interpreter at:
/Users/me/workspace/.venv/bin/python3
error: Distribution not found at: file:///Users/me/workspace/sam-app/lib


During handling of the above exception, another exception occurred:

  File "/opt/homebrew/Cellar/aws-sam-cli/1.162.1/libexec/lib/python3.14/site-packages/aws_lambda_builders/workflows/python_uv/packager.py", line 389, in _build_from_pyproject
    self._build_from_lock_file(lock_path, target_dir, scratch_dir, python_version, architecture, config)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/aws-sam-cli/1.162.1/libexec/lib/python3.14/site-packages/aws_lambda_builders/workflows/python_uv/packager.py", line 361, in _build_from_lock_file
    raise LockFileError(reason=str(e))
aws_lambda_builders.workflows.python_uv.exceptions.LockFileError: Lock file operation failed: Failed to install dependencies using uv: UV pip install failed: Using CPython 3.13.0 interpreter at:
/Users/me/workspace/.venv/bin/python3
error: Distribution not found at: file:///Users/me/workspace/sam-app/lib


During handling of the above exception, another exception occurred:

  File "/opt/homebrew/Cellar/aws-sam-cli/1.162.1/libexec/lib/python3.14/site-packages/aws_lambda_builders/workflows/python_uv/actions.py", line 68, in execute
    package_builder.build_dependencies(
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        artifacts_dir_path=target_artifact_dir,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ...<3 lines>...
        config=self.config,
        ^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/opt/homebrew/Cellar/aws-sam-cli/1.162.1/libexec/lib/python3.14/site-packages/aws_lambda_builders/workflows/python_uv/packager.py", line 229, in build_dependencies
    handler(manifest_path, artifacts_dir_path, scratch_dir_path, python_version, architecture, config)
    ~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/aws-sam-cli/1.162.1/libexec/lib/python3.14/site-packages/aws_lambda_builders/workflows/python_uv/packager.py", line 282, in _handle_pyproject_build
    self._build_from_pyproject(manifest_path, target_dir, scratch_dir, python_version, architecture, config)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/aws-sam-cli/1.162.1/libexec/lib/python3.14/site-packages/aws_lambda_builders/workflows/python_uv/packager.py", line 392, in _build_from_pyproject
    raise UvBuildError(reason=f"Failed to build from pyproject.toml: {str(e)}")
aws_lambda_builders.workflows.python_uv.exceptions.UvBuildError: UV package build failed: Failed to build from pyproject.toml: Lock file operation failed: Failed to install dependencies using uv: UV pip
install failed: Using CPython 3.13.0 interpreter at: /Users/me/workspace/.venv/bin/python3
error: Distribution not found at: file:///Users/me/workspace/sam-app/lib


During handling of the above exception, another exception occurred:

  File "/opt/homebrew/Cellar/aws-sam-cli/1.162.1/libexec/lib/python3.14/site-packages/aws_lambda_builders/workflow.py", line 374, in run
    action.execute()
    ~~~~~~~~~~~~~~^^
  File "/opt/homebrew/Cellar/aws-sam-cli/1.162.1/libexec/lib/python3.14/site-packages/aws_lambda_builders/workflows/python_uv/actions.py", line 76, in execute
    raise ActionFailedError(str(ex))
aws_lambda_builders.actions.ActionFailedError: UV package build failed: Failed to build from pyproject.toml: Lock file operation failed: Failed to install dependencies using uv: UV pip install failed:
Using CPython 3.13.0 interpreter at: /Users/me/workspace/.venv/bin/python3
error: Distribution not found at: file:///Users/me/workspace/sam-app/lib

2026-07-15 13:49:33,711 | Exception raised during the execution

Build Failed
```

Even though `lib` and `sam-app` are in the same directory level in the workspace, the workflow attempts to install `lib` under `sam-app`.

### Expected result:

Successful build is expected.

### Notes

I've submitted PR #887 to fix for this issue.

### Additional environment details (Ex: Windows, Mac, Amazon Linux etc)

1. OS: Mac
2. If using SAM CLI, `sam --version`: version 1.162.1
3. AWS region: ap-northeast-1

`Add --debug flag to any SAM CLI commands you are running`

noritada:
This PR fixes an issue where an application created as a member of a
uv workspace would fail to build if they depended on other workspace
members.

The following shows the problematic workspace structure and the error
message.

```
workspace root
├── lib
│   ├── pyproject.toml
│   └── src
├── pyproject.toml
├── sam-app
│   ├── __init__.py
│   ├── app.py
│   ├── pyproject.toml
│   ├── samconfig.toml
│   └── template.yaml
└── uv.lock
```

```
Build Failed
Error: PythonUvBuilder:ResolveDependencies - UV package build failed: Failed to build from pyproject.toml: Lock file operation failed: Failed to install dependencies using uv: UV pip install failed: Using CPython 3.13.0 interpreter at: /path/to/workspace/.venv/bin/python3
error: Distribution not found at: file:///path/to/workspace/sam-app/lib
```

Even though `lib` and `sam-app` are in the same directory level in the
workspace, the workflow attempts to install `lib` under `sam-app`.

In the workflow, `uv export` outputs a list of dependency packages,
which is then passed to `uv pip install` for installation. When doing
so, [`uv export` outputs relative paths from the workspace root][1].
Therefore, `uv pip install` must be run from the workspace root, not
from the application directory.

Additionally, dependencies on other packages within the workspace are
exported as editable installations (e.g., `-e ./lib`) by default.
When this is passed to `uv pip install`, only the `.pth` (path
configuration) file for the package will be installed without the
package body. To prevent this, the `--no-editable` option needs to be
used.

[1]: https://github.com/astral-sh/uv/issues/20238

Closes #892.

### Commands to reproduce the build failure

```bash
uv init --bare
uv init --lib lib
sam init --name sam-app --runtime python3.14 --architecture arm64 \
    --dependency-manager pip --package-type Zip \
    --app-template hello-world
cd sam-app
uv init
uv add lib@../lib
# remove requirements.txt and edit the app
# edit template.yaml to configure `BuildMethod: python-uv` and `CodeUri: .`
sam build --beta-features
```

### DIFF
--- a/aws_lambda_builders/workflows/python_uv/packager.py
+++ b/aws_lambda_builders/workflows/python_uv/packager.py
@@ -332,6 +332,7 @@ class PythonUvDependencyBuilder:
                 "--no-emit-project",  # Don't include the project itself, only dependencies
                 "--no-hashes",  # Skip hashes for cleaner output (optional)
                 "--no-default-groups",  # Exclude PEP 735 default groups (e.g. dev/test) from Lambda zips
+                "--no-editable",  # Export editable dependencies as non-editable
                 "--output-file",
                 temp_requirements,
                 # We want to specify the version because `uv export` might default to using a different one
@@ -344,6 +345,15 @@ class PythonUvDependencyBuilder:
             if rc != 0:
                 raise LockFileError(reason=f"Failed to export lock file: {stderr}")
 
+            # Get the workspace root (or project directory if no workspace is used)
+            # For packages in the workspace, exported paths are relative to the workspace root,
+            # regardless of where in the workspace uv export is called
+            workspace_args = ["workspace", "dir"]
+            rc, stdout, stderr = self._uv_runner._uv.run_uv_command(workspace_args, cwd=project_dir)
+            if rc != 0:
+                raise LockFileError(reason=f"Failed to get workspace root: {stderr}")
+            workspace_dir = stdout.strip()
+
             # Install with platform targeting
             self._uv_runner.install_requirements(
                 requirements_path=temp_requirements,
@@ -352,7 +362,7 @@ class PythonUvDependencyBuilder:
                 config=config,
                 python_version=python_version,
                 platform="linux",
-                cwd=project_dir,
+                cwd=workspace_dir,
                 architecture=architecture,
             )
         except LockFileError:
--- a/aws_lambda_builders/workflows/python_uv/packager.py
+++ b/aws_lambda_builders/workflows/python_uv/packager.py
@@ -332,7 +332,8 @@ class PythonUvDependencyBuilder:
                 "--no-emit-project",  # Don't include the project itself, only dependencies
                 "--no-hashes",  # Skip hashes for cleaner output (optional)
                 "--no-default-groups",  # Exclude PEP 735 default groups (e.g. dev/test) from Lambda zips
-                "--no-editable",  # Export editable dependencies as non-editable
+                # Install package bodies instead of editable .pth links, which break in Lambda zips.
+                "--no-editable",
                 "--output-file",
                 temp_requirements,
                 # We want to specify the version because `uv export` might default to using a different one
@@ -350,9 +351,13 @@ class PythonUvDependencyBuilder:
             # regardless of where in the workspace uv export is called
             workspace_args = ["workspace", "dir"]
             rc, stdout, stderr = self._uv_runner._uv.run_uv_command(workspace_args, cwd=project_dir)
-            if rc != 0:
-                raise LockFileError(reason=f"Failed to get workspace root: {stderr}")
-            workspace_dir = stdout.strip()
+            if rc == 0:
+                workspace_dir = stdout.strip()
+            else:
+                # `uv workspace dir` requires uv >= 0.9.9. Fall back to the project directory,
+                # which is what that command returns for any non-workspace project anyway.
+                LOG.debug("Could not determine workspace root, assuming no workspace: %s", stderr)
+                workspace_dir = project_dir
 
             # Install with platform targeting
             self._uv_runner.install_requirements(
--- a/aws_lambda_builders/workflows/python_uv/packager.py
+++ b/aws_lambda_builders/workflows/python_uv/packager.py
@@ -136,9 +136,9 @@ class UvRunner:
         # Add requirements file
         args.extend(["-r", requirements_path])
 
-        # Resolve --target to an absolute path: UV runs with cwd set to the project directory, so a
-        # relative target (e.g. the incremental-build dependencies dir) would otherwise be created
-        # under the source directory instead of the build root.
+        # Resolve --target to an absolute path: UV runs from the project or workspace directory,
+        # so a relative target (e.g. the incremental-build dependencies dir) would otherwise be
+        # created under the UV's cwd instead of the build root.
         args.extend(["--target", os.path.abspath(target_dir)])
 
         # Add configuration arguments
--- a/tests/integration/workflows/python_uv/test_python_uv.py
+++ b/tests/integration/workflows/python_uv/test_python_uv.py
@@ -297,3 +297,27 @@ class TestPythonUvWorkflow(TestCase):
 
         finally:
             shutil.rmtree(temp_source_dir)
+
+    @skipIf(which("uv") is None, "uv not available")
+    def test_workflow_builds_with_dependencies_within_workspace(self):
+        with tempfile.TemporaryDirectory() as workspace_dir:
+            shutil.copytree(os.path.join(self.TEST_DATA_FOLDER, "workspace"), workspace_dir, dirs_exist_ok=True)
+            source_dir = os.path.join(workspace_dir, "app")
+            builder = LambdaBuilder(language="python", dependency_manager="uv", application_framework=None)
+            builder.build(
+                source_dir,
+                self.artifacts_dir,
+                self.scratch_dir,
+                os.path.join(source_dir, "pyproject.toml"),
+                runtime=f"python{sys.version_info.major}.{sys.version_info.minor}",
+                experimental_flags=self.experimental_flags,
+            )
+
+            self.assertTrue(os.path.isfile(os.path.join(workspace_dir, "uv.lock")))
+            self.assertFalse(os.path.exists(os.path.join(source_dir, "uv.lock")))
+            for filename in ("__init__.py", "py.typed"):
+                installed = pathlib.Path(self.artifacts_dir, "workspace_lib", filename)
+                original = pathlib.Path(workspace_dir, "lib", "src", "workspace_lib", filename)
+                self.assertEqual(installed.read_bytes(), original.read_bytes())
+            self.assertTrue(os.path.isdir(os.path.join(self.artifacts_dir, "workspace_lib-0.1.0.dist-info")))
+            self.assertEqual(list(pathlib.Path(self.artifacts_dir).rglob("*.pth")), [])
--- a/tests/integration/workflows/python_uv/testdata/workspace/app/main.py
+++ b/tests/integration/workflows/python_uv/testdata/workspace/app/main.py
@@ -1,0 +1,5 @@
+from workspace_lib import message
+
+
+def handler(event, context):
+    return message()
--- a/tests/integration/workflows/python_uv/testdata/workspace/app/pyproject.toml
+++ b/tests/integration/workflows/python_uv/testdata/workspace/app/pyproject.toml
@@ -1,0 +1,8 @@
+[project]
+name = "workspace-app"
+version = "0.1.0"
+requires-python = ">=3.9"
+dependencies = ["workspace-lib"]
+
+[tool.uv.sources]
+workspace-lib = { workspace = true }
--- a/tests/integration/workflows/python_uv/testdata/workspace/lib/pyproject.toml
+++ b/tests/integration/workflows/python_uv/testdata/workspace/lib/pyproject.toml
@@ -1,0 +1,8 @@
+[project]
+name = "workspace-lib"
+version = "0.1.0"
+requires-python = ">=3.9"
+
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
--- a/tests/integration/workflows/python_uv/testdata/workspace/lib/src/workspace_lib/__init__.py
+++ b/tests/integration/workflows/python_uv/testdata/workspace/lib/src/workspace_lib/__init__.py
@@ -1,0 +1,2 @@
+def message():
+    return "Hello from the workspace dependency"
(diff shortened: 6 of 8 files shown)


roger-zhangg:
Thanks for the fix and the unusually clear write-up, @noritada — and apologies for the long silence on this one.

I reproduced this end to end and your diagnosis is correct. Details below, including one thing I'd like changed before merge.

## Yes, this fixes #892

I built the exact workspace from your repro (bare root + `lib` + an app member depending on `lib@../lib`) and drove the real `PythonUvDependencyBuilder.build_dependencies()` against it:

- **base `37cd1d1` + uv 0.12.9** → `UvBuildError: ... error: Distribution not found at: file:///private/tmp/wstest/app/lib` — the same failure as #892.
- **this PR + uv 0.12.9** → success, and `--target` contains `lib/` with the actual package body (`__init__.py`, `py.typed`) plus `lib-0.1.0.dist-info`, and no `.pth` files.

So #892 is closable when this merges.

I also confirmed that **both halves of the change are load-bearing**, which wasn't obvious to me from the diff alone. With only the `cwd` change (`packager.py:365`) and without `--no-editable` (`packager.py:335`), `uv pip install` *succeeds* but writes only `lib.pth` + `lib-0.1.0.dist-info` into the target — a Lambda zip that would `ModuleNotFoundError` at runtime. Worth stating that in the comment on line 335, since a future reader might otherwise take `--no-editable` for a cosmetic tidy-up and drop it.

And your claim about export paths (astral-sh/uv#20238) checks out: exporting from the member dir emits `-e ./lib`, i.e. relative to the workspace root, not `../lib` relative to the member.

Your approach also looks like the only reliable one here. Because uv writes `uv.lock` at the *workspace root*, a workspace member never takes the `os.path.exists(uv_lock_path)` branch at `packager.py:274-276`; it goes through `_build_from_pyproject`, and the `lock_path` assembled at `packager.py:398` does not actually exist on disk. That path works today only because `_build_from_lock_file` uses it purely for `os.path.dirname` (`packager.py:324`). So there is no local `uv.lock` to walk up from, and asking uv directly is the right call.

## One change requested: fall back when `uv workspace dir` is unavailable

`uv workspace dir` (`packager.py:351-355`) is recent — added in astral-sh/uv#16678, first released in **uv 0.9.9** (2025-11-12), stabilized in 0.10.0. (I tested 0.9.9 specifically: it works there with no `--preview` flag and no stderr warning, despite the `workspace-dir` preview gate mentioned in that PR.)

The issue is that `rc != 0` is fatal, and **every** `pyproject.toml` build funnels through `_build_from_lock_file` — workspace or not. On uv 0.8.17 against a plain single-project `pyproject.toml`, with no workspace involved at all:

- base `37cd1d1`: succeeds
- this PR: `LockFileError: Lock file operation failed: Failed to get workspace root: error: unrecognized subcommand 'workspace'`

aws-lambda-builders neither pins nor checks a minimum uv version (the only reference is `DESIGN.md:298`, "Minimum UV version: 0.1.0 (to be determined)"), and SAM CLI doesn't vendor uv — users bring their own, frequently pinned in CI images. As written, this breaks builds that work today for users who have no workspace and no need for the fix.

Since `uv workspace dir` returns the project directory itself for a non-workspace project (I verified this), degrading to the previous behaviour is exact rather than approximate:

```python
rc, stdout, stderr = self._uv_runner._uv.run_uv_command(workspace_args, cwd=project_dir)
if rc == 0:
    workspace_dir = stdout.strip()
else:
    # `uv workspace dir` requires uv >= 0.9.9. Fall back to the project directory,
    # which is what that command returns for any non-workspace project anyway.
    LOG.debug("Could not determine workspace root, assuming no workspace: %s", stderr)
    workspace_dir = project_dir
```

Workspace users on uv < 0.9.9 then get today's error instead of a confusing new one, and everyone on 0.9.9+ gets the fix. A version check would also work, but a fallback seems lower-risk than parsing version strings.

## Test results

- `pytest tests/unit/workflows/python_uv/` on this PR: **65 passed**. Base `37cd1d1` is also 65, so nothing was dropped.
- The PR **merges cleanly** into current `develop` (`587257c`); on the merge result: **68 passed** (develop has added 3 tests in the meantime).
- Full `pytest tests/unit`: **831 passed, 6 subtests passed**.
- `black --check` clean on both changed files.

## Minor, non-blocking

1. The unit tests mock `run_uv_command`, so they cannot catch the class of bug this PR fixes — the new assertions would pass just as happily against a wrong-but-consistent cwd string. Since `tests/integration/workflows/python_uv/testdata/` already exists, a small workspace fixture there is where this behaviour could actually be pinned down. Noting it more as follow-up work for us than as a request to you.
2. `test_packager.py:277` — `call_args_list[-2][0][0]` will silently point at the wrong call if a third uv invocation is ever added ahead of the install. Matching on the call whose first arg starts with `"export"` would survive that.
3. The comment at `packager.py:139-141` states "UV runs with cwd set to the project directory", which is no longer strictly true. The `os.path.abspath(target_dir)` on line 142 remains correct (it resolves against the builder's cwd, not uv's), but the stated rationale is now slightly off and could mislead later.
4. One thing I checked and did *not* find a problem with: moving cwd up to the workspace root does not appear to change uv's config discovery in a way that matters. A member-level `[tool.uv.pip] index-url` was not applied from either directory, so I observed no behavioural difference.

Overall the approach and the analysis behind it are right, and the end-to-end result is verified working. With the fallback above added, this looks good to me.

Issue #900, opened 2026-07-28
Rust workflow recompiles the full dependency tree once per function in a Cargo workspace (build time scales linearly with function count)


bnusunny:
### Description

This is a **performance** follow-up to #634. That issue fixed *correctness* for Cargo-workspace projects (binaries are now found), but the underlying build strategy still compiles each function in isolation, so shared dependencies are recompiled once per function.

For a workspace with N Rust functions that share heavy dependencies (e.g. the AWS SDK), `sam build` compiles the shared dependency tree **N times**. In a real project with 16 functions sharing the AWS SDK, this makes `sam build` take **over an hour**.

### Root cause

In [`cargo_lambda.py`](https://github.com/aws/aws-lambda-builders/blob/develop/aws_lambda_builders/workflows/rust_cargo/cargo_lambda.py#L104-L107) (v1.65.0):

```python
if not os.getenv("CARGO_TARGET_DIR"):
    # This results in the "target" dir being created under the member dir of a cargo workspace
    os.environ["CARGO_TARGET_DIR"] = "target"
```

`CARGO_TARGET_DIR` is set to the **relative** path `target`, and the build runs with `cwd=self._source_dir` (each function's own member directory — [`actions.py#L97`](https://github.com/aws/aws-lambda-builders/blob/develop/aws_lambda_builders/workflows/rust_cargo/actions.py#L97)). So each function gets an isolated `<member>/target/`, and cargo cannot reuse any compiled dependency across functions. The dependency graph is rebuilt from scratch for every function.

### Expected behavior

For a workspace, compile once and package many. A single `cargo lambda build --release [--arm64]` at the workspace root builds every member binary in one pass, compiling shared deps only once.

Local measurement on the same 16-function workspace:
- **Current (per-function):** > 1 hour
- **Single workspace build:** ~10m cold, ~2m19s warm

### Possible fix

The builder already supports selecting a single binary — `build_command` appends `--bin <handler>` ([`actions.py#L90`](https://github.com/aws/aws-lambda-builders/blob/develop/aws_lambda_builders/workflows/rust_cargo/actions.py#L90)). A workspace-aware path could:

1. Detect the Cargo workspace root and run one `cargo lambda build` there (shared `target/`, deps compiled once).
2. Resolve each function's binary from the shared `target/lambda/<name>/` output by package/bin name, and copy it into that function's `ARTIFACTS_DIR` — instead of shelling out per `CodeUri`.

[...]

PR #901, opened 2026-07-28
fix(rust): build Cargo workspace members into a shared target dir

bnusunny:
*Issue #, if available:* #900

### Description of changes

For a Cargo workspace, `sam build` invokes the Rust workflow once per function, and each invocation set `CARGO_TARGET_DIR` to the relative `"target"` under that member's own directory. Cargo therefore recompiled the entire shared dependency tree once per function, so build time scaled linearly with the number of functions (a real project with 16 functions sharing the AWS SDK took over an hour).

This resolves the workspace's shared `target` directory **and** the member's binary name from a single `cargo metadata` call, and builds every member into the shared directory so cargo compiles common dependencies once and reuses them across the per-function builds.

Because the shared `target/lambda` directory now holds *every* member's binary, the copy step can no longer assume it contains a single binary. When no explicit handler (`artifact_executable_name`) is given, it selects the binary of the package whose manifest lives in the function's source directory.

**Colliding bin names.** If two or more members define a `bin` target with the same name (e.g. each package declaring a `bootstrap` bin), those binaries build to the same path and overwrite each other. `sam build` still produces correct artifacts because it copies each function's binary out immediately after building it, but this relies on build ordering — so the workflow logs a warning recommending unique bin names.

Backward compatibility:
- **Standalone (non-workspace) project**: the shared target directory is the project's own `target` — unchanged.
- **Explicit `CARGO_TARGET_DIR`** in the environment still takes precedence.
- If `cargo metadata` can't be resolved, it falls back to the previous target directory and the single-binary directory-listing heuristic.

### Performance Improvement:

Local measurement on the same 16-function workspace:

**Current (per-function):** > 1 hour
**Single workspace build:** ~10m cold, ~2m19s warm

- New unit tests for workspace-layout resolution (shared target dir, member-binary selection, ambiguous/absent packages, colliding-bin warning, and the failure fallbacks) and the copy action's binary selection.
- New integration test that builds two workspace members **with no explicit handler** into a single shared `target/lambda`, asserting each function's own binary is copied with no per-member `target/`.
- Full rust unit + integration suites pass; `ruff` and `black` clean.
- Validated end-to-end with real `sam build -c`: cold build succeeds and copies each function's binary correctly; unchanged rebuild is served from SAM's cache; changing one function rebuilds only that function while reusing shared compiled dependencies. Also verified a workspace where every member names its bin `bootstrap` still builds correct, distinct artifacts (sequential and `--parallel`) and emits the collision warning.

### Checklist

### DIFF
--- a/aws_lambda_builders/workflows/rust_cargo/DESIGN.md
+++ b/aws_lambda_builders/workflows/rust_cargo/DESIGN.md
@@ -16,9 +16,15 @@ The general algorithm for preparing a rust executable for use on AWS Lambda is a
 
 It builds a binary in the standard cargo target directory. The binary's name is always `bootstrap`, and it's always located under `target/lambda/HANDLER_NAME/bootstrap`.
 
+For a Cargo workspace, the build targets the workspace's shared `target` directory rather than a `target` directory under each member. Because `sam build` invokes this workflow once per function, sharing a single target directory lets cargo compile common dependencies once instead of recompiling the whole dependency tree for every function. The shared directory and the member's binary name are both read from a single `cargo metadata` call. For a standalone (non-workspace) project the shared `target` directory is the project's own — unchanged from prior behavior. An explicit `CARGO_TARGET_DIR` in the environment always takes precedence.
+
 ### Copy and Rename executable
 
 It then copies the executable to the target directory honoring the provided runtime's [expectation on executable names](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-custom.html).
+
+Because every workspace member now builds into the same `target/lambda` directory, that directory holds all of the workspace's binaries. When no handler (`artifact_executable_name`) is given, the copy step selects the binary named for the package whose manifest lives in the function's source directory (from the same `cargo metadata` call). It falls back to the previous single-directory heuristic when the binary name cannot be resolved.
+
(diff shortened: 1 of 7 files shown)


bnusunny:
Good catch, and confirmed. `os.path.normpath` does not absolutize a relative path, so a relative `source_dir` never matched cargo's absolute `manifest_path`, leaving `binary_name = None` and (for a workspace) falling through to the multi-entry `os.listdir` heuristic that raises `unable to find function binary`.

Fixed by resolving both operands with `os.path.realpath` (absolute + symlink-resolved, matching cargo metadata's output), and added a regression test `test_resolves_binary_when_source_dir_is_relative` that passes a relative `source_dir` — it fails on the previous `normpath` code and passes with the fix. Pushed in the latest revision.

bnusunny:
Confirmed, and fixed. Verified empirically: for a workspace member, cargo's *default* target dir (no `CARGO_TARGET_DIR`) is the workspace **root** `target/`, whereas the copy step's fallback looks in `<source_dir>/target/lambda` — so on any metadata failure the build landed in the root while the copy looked in the member dir, and failed where the pre-PR code (which set relative `CARGO_TARGET_DIR=target`) succeeded.

Fixed by setting `cargo_env["CARGO_TARGET_DIR"] = target_directory or "target"`, restoring the exact legacy fallback that matches the copy step. Added `TestRunTargetDir` covering the resolved-dir, metadata-failure fallback, and explicit-env cases; the fallback test fails on the previous code (CARGO_TARGET_DIR left unset) and passes now. Pushed in the latest revision.

roger-zhangg (approved):
LGTM

licjun (approved):

Issue #903, opened 2026-07-29
Bug: python_uv workflow forces UV cache into per-build scratch dir, so every build starts cold


bnusunny:
### Description:

The `python_uv` workflow forces UV's cache into the per-build scratch directory, which is a fresh `mkdir_temp()` that SAM CLI deletes when the build finishes. UV therefore starts from a completely cold cache on **every function, on every build**, and can never reuse anything it downloaded before.

`UvRunner._ensure_cache_dir()` ([packager.py#L98-L103](https://github.com/aws/aws-lambda-builders/blob/develop/aws_lambda_builders/workflows/python_uv/packager.py#L98-L103)):

```python
def _ensure_cache_dir(self, config: UvConfig, scratch_dir: str) -> None:
    """Ensure UV cache directory is configured."""
    if not config.cache_dir:
        config.cache_dir = os.path.join(scratch_dir, "uv-cache")
        if not os.path.exists(config.cache_dir):
            self._osutils.makedirs(config.cache_dir)
```

That value is emitted as an explicit `--cache-dir` argument ([utils.py#L131-L132](https://github.com/aws/aws-lambda-builders/blob/develop/aws_lambda_builders/workflows/python_uv/utils.py#L131-L132)), which also overrides any `UV_CACHE_DIR` the user has set. `scratch_dir` originates from `osutils.mkdir_temp()` inside SAM CLI's `ApplicationBuilder._build_function()`, so it is created and destroyed once per build definition.

This makes the effect twofold:

- **Between functions in one `sam build`** — each function gets its own empty `uv-cache`, so shared dependencies (boto3, pydantic, …) are re-downloaded and re-unpacked once per function. Build time scales linearly with function count.
- **Between successive builds** — the scratch directory is deleted on exit, so the next `sam build` starts cold again.

Both manifest dispatch paths are affected, since `_handle_pyproject_build` and `_build_from_requirements` both call `UvRunner.install_requirements()`.

For contrast, `python_pip` never sets `PIP_CACHE_DIR`, so pip transparently uses the user-global `~/.cache/pip` and stays warm across functions and builds. UV — whose main advantage is aggressive caching — is the only Python workflow deprived of it.

This is structurally the same class of problem as #900 (Rust workspace members each getting an isolated `target/`, defeating cargo's cache), with the additional penalty that the cache is discarded between builds and not just between functions.

### Steps to reproduce:

[...]

PR #904, opened 2026-07-29
fix(python-uv): let UV use its own persistent cache

bnusunny:
*Issue #, if available:* #903

### Description of changes

`UvRunner._ensure_cache_dir()` pointed UV's `--cache-dir` at a `uv-cache` directory inside the build's scratch directory. Scratch comes from `osutils.mkdir_temp()` in SAM CLI's `ApplicationBuilder._build_function()` and is deleted when the build finishes, so UV started from a cold cache on every function of every build. Within one `sam build` each function re-downloaded shared dependencies (build time scaled linearly with function count), and the next build started cold again. The explicit flag also overrode any `UV_CACHE_DIR` the user had set, so there was no workaround.

This removes `_ensure_cache_dir()`. With no `--cache-dir`, UV uses its own default cache, which persists across functions and builds — matching `python_pip`, which never sets `PIP_CACHE_DIR` and inherits pip's user-level cache. A caller-supplied `config.cache_dir` is still forwarded, and `UV_CACHE_DIR` is honoured again.

`scratch_dir` is dropped from `install_requirements()`, where it had no remaining use. It stays on the manifest handlers: `_build_from_lock_file` still writes exported requirements there, and `_build_from_requirements` keeps it to match the signature shared by handlers dispatched from `build_dependencies()`.

Existing suites: 831 unit, 13 functional, 20 integration — all pass. `ruff` and `black` clean.

New unit tests assert no `--cache-dir` is passed and no cache directory is created by default, that an explicit `cache_dir` is still honoured, and that a build never derives `cache_dir` from scratch. I verified these fail against the old implementation on the assertion (not merely on the signature change) by re-introducing the old cache-defaulting logic under the new signature.

Measured with boto3 + pydantic + requests, medians of 5 iterations, all paths on one filesystem, UV 0.11.28:

| | before | after (warm) |
|---|---|---|
| requirements.txt | 2622 ms | 245 ms |
| pyproject.toml | 2963 ms | 481 ms |
| 4 functions | 10436 ms | 3236 ms |
| 8 functions | 20658 ms | 4051 ms |

Cold first build is unchanged (2664 ms vs 2622 ms, within noise), so nothing gets slower.

Two risks I checked before proposing a shared cache:

- **Concurrency** (`sam build --parallel`): six concurrent installs against one cold cache completed with no errors and identical output trees. UV locks its cache.
- **Hardlinks**: files installed via `--target` report `nlink=1`, so UV copies rather than hardlinks for `--target` installs and a persistent cache can't be corrupted by later mutation of build artifacts.

### Checklist

### DIFF
--- a/aws_lambda_builders/workflows/python_uv/DESIGN.md
+++ b/aws_lambda_builders/workflows/python_uv/DESIGN.md
@@ -201,7 +201,7 @@ The workflow supports various configuration options through the config parameter
 config = {
     "index_url": "https://pypi.org/simple/",  # Custom package index
     "extra_index_urls": [],                   # Additional package indexes
-    "cache_dir": "/tmp/uv-cache",            # Custom cache directory
+    "cache_dir": None,                       # Custom cache directory; UV's own default when unset
     "no_cache": False,                       # Disable caching
     "prerelease": "disallow",                # Handle pre-release versions
     "resolution": "highest",                 # Resolution strategy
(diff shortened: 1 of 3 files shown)


vicheey (approved):
Approved!
One small comment, not blocking: consider documenting the full cache precedence chain (`config.cache_dir` > `UV_CACHE_DIR` > UV default) in the DESIGN.md caching section.

roger-zhangg (approved):

